PR #2, opened 2021-10-14
Clean and streamline code in Benchmark.swift

kaandedeoglu:
## Summary

Various quality of life improvements in Benchmark.swift

- Remove abundant closure for `static let main` instance initialization.
- ~Make the `metrics` array read-only to the outside world.~
- Avoid calculating `id` & `displayName` in case of an early-exit.
- Unify checks prior to running benchmarks, remove duplication.

## Testing

Unit tests still pass, no functionality change intended.

## Checklist

### DIFF
--- a/Sources/SwiftDocC/Semantics/MarkupReferenceResolver.swift
+++ b/Sources/SwiftDocC/Semantics/MarkupReferenceResolver.swift
@@ -101,15 +101,6 @@ struct MarkupReferenceResolver: MarkupRewriter {
             return link
         }
         guard url.components.scheme == ResolvedTopicReference.urlScheme else {
-            if url.components.scheme == "topic", let linkRange = link.range {
-                var components = url.components
-                components.scheme = ResolvedTopicReference.urlScheme
-                let diagnostic = Diagnostic(source: source, severity: .warning, range: link.range, identifier: "org.swift.docc.deprecatedSchemaReference", summary: "'topic' reference scheme is deprecated in favor of the 'doc' sheme")
-                let solution = Solution(summary: "Replace 'topic' reference scheme with 'doc' scheme.", replacements: [
-                    Replacement(range: linkRange, replacement: "<\(components.url!.absoluteString)>")
-                ])
-                problems.append(Problem(diagnostic: diagnostic, possibleSolutions: [solution]))
-            }
             return link // Create a non-topic link
         }
         let unresolved = TopicReference.unresolved(.init(topicURL: url))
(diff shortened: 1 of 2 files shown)


Sources/SwiftDocC/Benchmark/Benchmark.swift
```diff
@@ -51,7 +49,7 @@ public class Benchmark: Encodable {
     #endif
 
     /// The list of metrics included in this benchmark.
-    public var metrics: [BenchmarkMetric] = []
+    public fileprivate(set) var metrics: [BenchmarkMetric] = []
```
franklinsch, line 52: I think keeping the setter `public` would help with flexibility here, for example for clients that want to manually process this array before encoding.

Sources/SwiftDocC/Benchmark/Benchmark.swift
```diff
@@ -130,3 +122,9 @@ public func benchmark<E, Result>(wrap event: @autoclosure () -> E, benchmarkLog
         return try body()
     }
 }
+
+private extension BenchmarkMetric {
+    static func canBenchmark(log: Benchmark) -> Bool {
```
franklinsch, line 127: This reads a bit odd to me, because it could be inferred as benchmarking the log. I’m wondering if flipping the extension here would read nicer:

```swift
private extension Benchmark {
    func shouldLogMetric(_ metric: BenchmarkMetric) -> Bool {
        …
    }
}
```

What do you think?

icanzilb: Other than the name, I think that's a very neat optimization to extract the logic into a separate method 👍🏼

franklinsch: Makes sense, good idea!

Sources/SwiftDocC/Benchmark/Benchmark.swift
```diff
@@ -71,55 +69,55 @@ public class Benchmark: Encodable {
...
-        log.metrics.append(event())
-    }
+    guard log.shouldLogMetricType(E.self) else { return }
+
+    log.metrics.append(event())
 }
 
 /// Starts the given metric.
 /// - Parameter event: The metric to start.
 public func benchmark<E>(begin event: @autoclosure () -> E, benchmarkLog log: Benchmark = .main) -> E? where E: BenchmarkBlockMetric {
```
bontoJR, line 99: Can we also update the doc adding the missing `log` parameter?

franklinsch:
@swift-ci test

PR #3, opened 2021-10-15
Fix SR-15315

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

## Dependencies

## Testing

applicable._

Steps:

## Checklist

- [ ] Added tests


Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1232,7 +1237,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
         }
         
         // Check if media is a supported image.
-        if DocumentationContext.isFileExtension(fileExtension, supported: .image),
+        if fileExtension.isEmpty || DocumentationContext.isFileExtension(fileExtension, supported: .image),
```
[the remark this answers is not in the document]
Kyle-Ye: I agree with it, and update it as follow. Any new suggestion?

https://github.com/Kyle-Ye/swift-docc/blob/a91994717f5bc527160bbafb63d0788cfc968fc7/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2326-L2358

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2328,9 +2328,20 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
     /// - Parameters:
     ///   - name: The name of the asset.
     ///   - parent: The topic where the asset is referenced.
+    ///   - type: If not nil, check the asset's extension type
```
franklinsch, line 2331: I’d suggest describing what the parameter represents, then what it means for this function if it’s not nil:

>The type of the asset to resolve. If an asset with such type could not be found in the context, this function returns nil.

franklinsch: Yes, that sounds right, thank you!

Kyle-Ye:
Rebase to the updated main branch.
Update the doc comment and add the related test.
Any suggestion on the test function? @franklinsch

franklinsch:
Thanks for adding the test! It tests the behavior of the API in `DocumentationContext` , but it doesn’t cover that `RenderNodeTranslator` actually calls it with the correct parameters, i.e., the test would still pass if we weren’t specifying the type argument to this function in RenderNodeTranslator. Does that make sense?

I’d recommend implementing another test in `RenderNodeTranslatorTests` to verify similar assertions of your existing test. Thanks so much for doing this work and iterating on it, I really appreciate it!

Kyle-Ye:
> Thanks for adding the test! It tests the behavior of the API in `DocumentationContext` , but it doesn’t cover that `RenderNodeTranslator` actually calls it with the correct parameters, i.e., the test would still pass if we weren’t specifying the `type` argument to this function in RenderNodeTranslator. Does that make sense?
> 
> I’d recommend implementing another test in `RenderNodeTranslatorTests` to verify similar assertions of your existing test. Thanks so much for doing this work and iterating on it, I really appreciate it!

Did not find a file named `RenderNodeTranslatorTests.swift` and add the test into `SemaToRenderNodeTests.swift`

PR #5, opened 2021-10-15
Port to Android

buttaface:
## Summary

This pull gets this repo to build natively on Android and pass all its tests.

## Dependencies

None

## Testing

Install [the Termux app on an Android phone or tablet](https://termux.com/), install the Swift 5.5 and git packages with `pkg install git swift`, clone this repo, and run `swift test -j 9` as you would elsewhere.

## Checklist

- [ ] Added tests - no need as the goal is to just get the existing tests running
- [ ] Updated documentation if necessary - nothing extra to be noted

Since this pull mostly consists of adding conditional compilation checks for Android where needed, let me point out where I _didn't_ add them:

- [Switching to JSON here seemed to make no difference](https://github.com/apple/swift-docc/blob/ff972f1/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift#L15).
- I don't know [what this is about](https://github.com/apple/swift-docc/blob/ff972f1b6f6ce0a73336e05e21a3406facb7571c/Sources/SwiftDocC/Infrastructure/Workspace/LocalFileSystemDataProvider.swift#L34), but no difference again.
- [Concurrently rendering worked fine](https://github.com/apple/swift-docc/blob/ff972f1b6f6ce0a73336e05e21a3406facb7571c/Sources/SwiftDocC/Model/Rendering/RenderContext.swift#L64).
- [These constants appear to be unused](https://github.com/apple/swift-docc/blob/ff972f1b6f6ce0a73336e05e21a3406facb7571c/Sources/SwiftDocC/Utility/LinuxBundleShims.swift#L11).

In addition, I ran all the tests on Android first and only disabled the tests that failed, so three tests that are disabled for Linux in `ConvertServiceTests.swift` and eight tests in `NavigatorIndexTests.swift` are enabled for Android. There's a strange issue where if I run the tests in parallel, two `LogHandle` tests fail only because they're grabbing extra lines from `stdout`, which may be an XCTest bug.

When I run the tests serially, it reports all 976 tests passing.


Tests/SwiftDocCUtilitiesTests/PreviewServer/PreviewServerTests.swift
```diff
@@ -26,7 +26,7 @@ class PreviewServerTests: XCTestCase {
         ])
         try tempFolder.write(to: URL(fileURLWithPath: NSTemporaryDirectory().appending(UUID().uuidString)))
 
-        let socketURL = URL(fileURLWithPath: "/var/tmp").appendingPathComponent(UUID().uuidString).appendingPathExtension("sock")
+        let socketURL = URL(fileURLWithPath: FileManager.default.temporaryDirectory.path).appendingPathComponent(UUID().uuidString).appendingPathExtension("sock")
```
buttaface, line 29: No `/var/tmp` on Android.

ethan-kusters:
@swift-ci please test

buttaface:
> We actually have a related bug tracking supporting Android toolchain builds (https://bugs.swift.org/browse/SR-15321). Is that something we should consider addressing in this PR as well?

Yes, I was surprised to see that. 😄 I have only built this repo with the Termux Swift toolchain as detailed above, not with your Python script. I will try it with your script later and submit a separate pull for that.

> DocC’s bin/test script that we encourage contributors to use runs tests in parallel so I’d prefer if we continue to disable these tests for now and file a bug to track the issue with LogHandle.

I simply modified `bin/test` so that step passed, and all subsequent steps passed too. No need to disable LogHandle for Android, as it passes when run serially. I cannot reproduce that issue on linux, so this may be a bug that crept into XCTest on Android. I will investigate it further, but it shouldn't affect getting this pull in.

No problem on waiting till this gets into the trunk toolchain, looks like the linux CI is broken right now.

ethan-kusters:
@swift-ci please test

ethan-kusters:
Thank you again @buttaface! I think we are good to merge.

PR #6, opened 2021-10-15
Remove linux-specific code and enable tests that pass on linux

buttaface:
## Summary

These linuxisms were not needed on Fedora Core 33. 

## Dependencies

None

## Testing

Steps:
1. Apply this patch.
2. Run `./bin/test` on linux with Swift 5.5 installed.

## Checklist

- [x] Added tests - sort of, since I'm enabling more tests
- [ ] Updated documentation if necessary - nothing to mention

I noted that all these weren't needed on Android in #5, so I checked on Fedora Core 33 x86_64 and they weren't needed on linux either.

Note that I have not looked into why any of these conditional compilation checks for linux were originally added, I just know that they make no difference in getting the tests passing. If some of these should stay the same, just let me know.


ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

PR #7, opened 2021-10-16
[Test Fix] Fix test fail on non-English locale (SR-15345)

Kyle-Ye:
Bug
https://bugs.swift.org/browse/SR-15345

## Summary

Fix test fail on non-English locale machine

## Testing

1. Change the machine locale to a non-English locale
2. Run the ./bin/test

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Model/SemaToRenderNodeTests.swift
```diff
@@ -639,13 +639,13 @@ class SemaToRenderNodeTests: XCTestCase {
         }
         XCTAssertEqual(firstTutorialReference.url, "/tutorials/test-bundle/testtutorial")
         XCTAssertFalse(firstTutorialReference.abstract.isEmpty)
-        XCTAssertEqual(firstTutorialReference.estimatedTime, "20min")
+        XCTAssertEqual(firstTutorialReference.estimatedTime, translator.contentRenderer.formatEstimatedDuration(minutes: 20))
```
Kyle-Ye, line 642: What's the purpose of the test expression?

I just afraid of that if the purpose is to test `DocumentationContentRenderer'formatEstimatedDuration(minutes:)`, it seemed  the test now is useless, since we call the function again on the expect string.

And what is the best solution for this?

franklinsch: Yes, part of the purpose here is to test this behavior. I’m assuming you’re seeing a test failure on macOS because DateComponentsFormatter (correctly) localizes the date. Since general support for localization is tracked by [SR-15352](https://bugs.swift.org/browse/SR-15352), I’d suggest we instead use the Linux behavior on all platforms and re-introduce using DateComponentsFormatter as part of that work.

PR #21, opened 2021-10-29
Navigator index multi-language render JSON support

franklinsch:
Bug/issue #, if applicable: rdar://83398083

## Summary

Adds support for the navigator index to process render JSON that contains variant overrides by applying patches for the languages in which the page is available.

Note: the first commit adds the support and second commit fixes a typo

## Dependencies

None.

## Testing

Verify that when building a navigator index that contains multi-language render JSON, the content gets curated under each language tree.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -707,6 +707,22 @@ extension NavigatorIndex {
                 multiCuratedUnvisited.remove(normalizedIdentifier)
             }
             
+            // Check if the render node has a variant for Objective-C
+            let objCVariantTrait = renderNode.variantOverrides?.values.flatMap({ $0.traits }).first { trait in
+                switch trait {
+                case .interfaceLanguage(let language):
+                    return InterfaceLanguage.from(string: language) == .objc
+                }
+            }
+            
+            // In case we have a variant for Objective-C, apply the variant and re-index the render node.
+            if let variantToApply = objCVariantTrait {
+                let encodedRenderNode = try renderNode.encodeToJSON()
+                let transformedData = try RenderNodeVariantOverridesApplier().applyVariantOverrides(in: encodedRenderNode, for: [variantToApply])
+                let variantRenderNode = try RenderNode.decode(fromJSON: transformedData)
+                try index(renderNode: variantRenderNode)
+            }
+            
             // Bump the nodes counter.
             counter += 1
         }
--- a/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
+++ b/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
@@ -14,7 +14,7 @@ import XCTest
 typealias Node = NavigatorTree.Node
 typealias PageType = NavigatorIndex.PageType
 
-let testBundleIndetifier = "org.swift.docc.example"
+let testBundleIdentifier = "org.swift.docc.example"
 
 class NavigatorIndexingTests: XCTestCase {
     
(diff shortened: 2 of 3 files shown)


ethan-kusters (approved):

ethan-kusters:
@swift-ci please test this

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -707,6 +707,22 @@ extension NavigatorIndex {
                 multiCuratedUnvisited.remove(normalizedIdentifier)
             }
             
+            // Check if the render node has a variant for Objective-C
+            let objCVariantTrait = renderNode.variantOverrides?.values.flatMap({ $0.traits }).first { trait in
```
jackhl, line 711: This code assumes that the primary variant is Swift, FWIW. It'd be more flexible if it just checked for other variants besides the primary variant.

Also, it assumes there are only two variants, swift and objective-c, but I understand that's a larger problem to resolve.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -707,6 +707,22 @@ extension NavigatorIndex {
...
+                switch trait {
+                case .interfaceLanguage(let language):
+                    return InterfaceLanguage.from(string: language) == .objc
```
jackhl, line 714: Is the interface language for a pure C library `objc`?

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
jackhl: Couldn't this just loop over all the `variantOverrides` instead of reaching into it for a particular one?

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -707,6 +707,22 @@ extension NavigatorIndex {
...
+                }
+            }
+            
+            // In case we have a variant for Objective-C, apply the variant and re-index the render node.
+            if let variantToApply = objCVariantTrait {
+                let encodedRenderNode = try renderNode.encodeToJSON()
+                let transformedData = try RenderNodeVariantOverridesApplier().applyVariantOverrides(in: encodedRenderNode, for: [variantToApply])
```
jackhl, line 721: Is it worth just keeping around the JSON for every render node so you don't have to re-encode? Or is the perf cost not high enough to warrant the increase in peak memory usage?

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
franklinsch: For now, this is just indexing the variant overrides for Objective-C. Supporting an arbitrary set of languages would be a larger task.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
franklinsch: This will be decided in the implementation of the compiler that will emit symbol graph files, we'll want to extend the support for pure C libraries then.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
franklinsch: The indexing during conversion happens on in-memory `RenderNode` models, so we're not "re-encoding". The `RenderNodeVariantOverridesApplier` works on the JSON data rather than a Codable model, so there isn't really a better way to write this at the moment.

franklinsch:
@swift-ci please test

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
jackhl: Ahh I see, ok :thumbsup:

jackhl (approved):

franklinsch:
@swift-ci please test

PR #23, opened 2021-11-01
Refactor DocumentationBundle.Info

ethan-kusters:
## Summary

This work refactors the existing logic for handling DocumentationBundle Info to make it easier to add new configuration options moving forward.

## Details

1. Use DocumentationBundle.Info to pass around bundle info. 

    Currently we pass around DocumentationBundle Info as function parameters in a multitude of places. This means that when we add a new property to the `DocumentationBundle.Info` model, we have to add new function parameters in several places across the codebase. It's easy to miss some of these and I think likely to introduce bugs in the future.

   We already have a model for handling documentation bundle info, so this just updates a number of calling sites to rely on passing around the DocumentationBundle.Info directly instead of its properties.

   This change is isolated in this commit c357b35 with another commit here (1c850a6) that addresses deprecation warnings.

2. Use Codable to parse the Info.plist

    I believe when this code was originally written, `PropertyListEncoder` was macOS only so we relied on using `PropertyListSerialization` directly instead.

    `PropertyListEncoder` is now supported on Linux so I refactored the DocumentationBundle.Info models to be Codable which simplifies a lot of our logic here.

    That change is isolated here 2e5658d.

3. Use a strongly-typed API for creating bundle discovery options

    The bundle discovery options are intentionally stringly-typed so that we can easily add new info plist fields and have clients adopt them immediately. However, there's no reason for us to be using the stringly-typed API within DocC itself. This led to some unnecessary duplication of the Info.plist keys.

    This change is isolated to this commit 22c6c80 and just adds some convenience initializers to the bundle discovery options to reduce duplication of the Info.plist keys in DocC itself.

## Dependencies

None.

## Testing

Perform basic regression testing. No functionality change is expected.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/DocumentationService/Convert/ConvertService+DataProvider.swift
+++ b/Sources/SwiftDocC/DocumentationService/Convert/ConvertService+DataProvider.swift
@@ -21,26 +21,20 @@ extension ConvertService {
         var files: [URL: Data] = [:]
         
         mutating func registerBundle(
-            displayName: String,
-            identifier: String,
-            version: String,
+            info: DocumentationBundle.Info,
             symbolGraphs: [Data],
             markupFiles: [Data],
-            miscResourceURLs: [URL],
-            defaultCodeListingLanguage: String?
+            miscResourceURLs: [URL]
         ) {
             let symbolGraphURLs = symbolGraphs.map { registerFile(contents: $0, isMarkupFile: false) }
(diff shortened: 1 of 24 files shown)


ethan-kusters:
@swift-ci please test

franklinsch (approved):
This is awesome, thanks for doing this! And thanks for splitting up the work into different commits so that it's easier to follow

PR #24, opened 2021-11-01
DocC convert should produce deterministic output

ethan-kusters:
Bug/issue #, if applicable: rdar://80114489

## Summary

Enables deterministic sorting of RenderJSON output in the few places where it's missing, such as "Default Implementation" sections and media variant sections.

## Dependencies

None.

## Testing

Perform multiple conversions of the same documentation content and confirm that the RenderJSON files in the produced `/data` directory are the same across conversions.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Converter/TopicRenderReferenceEncoder.swift
```diff
@@ -38,10 +38,8 @@ enum TopicRenderReferenceEncoder {
         // The keys of the node render references
         var referenceIndexes = Array(references.keys)
         
-        // Sort the keys if `.sortedKeys` is set on the encoder.
-        if encoder.outputFormatting.contains(.sortedKeys) {
-            referenceIndexes.sort()
-        }
+        // Sort the render references so that RenderJSON output is stable and deterministic
+        referenceIndexes.sort()
```
ethan-kusters, line 42: I don't think it makes sense to gate this sorting on the encoder settings since we sort the rest of the RenderJSON output regardless of that setting.

However, if this causes a performance impact in CI testing, I'll revert this change and we can explore this in a future PR.

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci Please test compiler performance

ethan-kusters:
@swift-ci please test

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1664,6 +1664,89 @@ class ConvertActionTests: XCTestCase {
         }
     }
     
+    func testConvertActionProducesDeterministicOutput() throws {
```
franklinsch, line 1667: Good idea!

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1664,6 +1664,89 @@ class ConvertActionTests: XCTestCase {
...
+        // Pretty printing the output JSON also enables sorting keys during encoding
+        // which is required for testing if the conversion output is deterministic.
+        let priorPrettyPrintValue = shouldPrettyPrintOutputJSON
+        shouldPrettyPrintOutputJSON = true
+        defer {
+            shouldPrettyPrintOutputJSON = priorPrettyPrintValue
+        }
```
franklinsch, line 1674: When running tests in parallel, I think this will affect other documentation builds (i.e., they might get `shouldPrettyPrintOutputJSON = yes`). One way to work around this I think would be to decode the JSON on disk below, then re-encode them using an encoder that has pretty formatting enabled.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1664,6 +1664,89 @@ class ConvertActionTests: XCTestCase {
...
+                outputURL: URL(fileURLWithPath: "/2", isDirectory: true),
+                testFileSystem: testFileSystem
+            )
+            
+            // Extract and sort the RenderJSON output of each conversion
+            
+            let firstConversionFiles = testFileSystem.files.lazy.filter { key, _ in
+                key.hasPrefix("/1/data/")
+            }.map { (key, value) in
+                return (String(key.dropFirst(2)), value)
```
franklinsch, line 1724: Nit, so that it's a bit clearer what we're dropping

```suggestion
                return (String(key.dropFirst("/1".count)), value)
```

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1664,6 +1664,89 @@ class ConvertActionTests: XCTestCase {
...
+            let secondConversionFiles = testFileSystem.files.lazy.filter { key, _ in
+                key.hasPrefix("/2/data/")
+            }.map { (key, value) in
+                return (String(key.dropFirst(2)), value)
+            }.sorted(by: \.0)
+            
+            // Zip the two sets of sorted files and loop through them, ensuring that
+            // each conversion produced the same RenderJSON output.
+            
+            for (first, second) in zip(firstConversionFiles, secondConversionFiles) {
```
franklinsch, line 1736: We should check that `firstConversionFiles.count` == `secondConversionFiles.count` because `zip` will silently drop unmatched pairs if the sequences have different lengths.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
ethan-kusters: Thanks!

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
ethan-kusters: I added an in-source comment to clarify this but, even when run in parallel, only one test is being executed per test-process. So because we’re modifying this value in-process (instead of in the environment) there isn’t concern of this affecting other tests when run in parallel.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
ethan-kusters: Good idea! Done.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
ethan-kusters: Ah good catch. I went ahead and added an `XCTAssertEqual` above that verifies that each array of file paths are entirely equal. This should address that issue and allows us to remove the individual assert in this loop.

ethan-kusters:
@swift-ci please test

franklinsch (approved):

ethan-kusters:
@swift-ci please test

PR #25, opened 2021-11-02
Add variants APIs for multi-language symbols

franklinsch:
Bug/issue #, if applicable: rdar://83665056

## Summary

Update the Symbol API to support variants. Each property has a `variants` counterpart property and the original properties have been updated to forward to the variants counterpart. 

The first commit just hardcodes accesses to use the Swift variant so that there are no user-facing changes.

Then, the second commit adds support for emitting multi-language symbol data:

1. The `VariantCollection+Symbol.swift` file contains initializers that create render node variant models 
(`VariantCollection`) from symbol data (`DocumentationDataVariants`).
2. The `RenderNodeTranslator` has been updated to consume all the symbols' variants and populate 
render node `VariantCollection` properties.
3. The logic to create render node sections for the `RenderNode.primaryContentSections` array has been
refactored to subtypes of the new `RenderSectionTranslator` protocol, for clarity.
4. The `JSONPatch` infrastructure now supports the `add` operation, which was needed for values
that are not present in the default render node's fields, but for which we have overrides for.
5. In the second commit, `SymbolDataVariants` was renamed to `DocumentationDataVariants` since the type
is now used beyond symbol data, e.g., as the return value of `LinkTitleResolver.title(for:)`.

## Dependencies

None.

## Testing

There should be no user-facing changes yet, since this doesn't bring support for ingesting Objective-C symbol graph files yet.

## Checklist

- [x] Added tests


franklinsch:
@swift-ci please test

franklinsch:
@swift-ci test

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -13,174 +13,284 @@ import Markdown
 import SymbolKit
 
 /// A programming symbol semantic type.
+///
+/// This class's properties are represented using ``SymbolDataVariants`` values which encode the different variants for each piece of information
+/// about the symbol. For example, if a symbol is available in multiple programming languages, the ``titleVariants`` property represents the title of the symbol
+/// for each programming language it's available in. Use a ``SymbolDataVariantsTrait`` to access a specific variant.
+///
+/// ## Topics
```
QuietMisdreavus, line 21: Really thankful that you added documentation links for all the new fields!

QuietMisdreavus (approved):
Looks good, thanks!

PR #26, opened 2021-11-03
Disabling noisy readonly fs test

etcwilde:
This patch disables the readonly fs indexing test due to the CI noise
and potential to leak resources on failures.

rdar://84986427


etcwilde:
@swift-ci please test

d-ronnqvist (approved):

PR #28, opened 2021-11-05
Disabling PreviewServerTests due to networking

etcwilde:
We shouldn't be making network calls or accessing the temp directories
in CI tests.

rdar://85046362


etcwilde:
@swift-ci please test

PR #40, opened 2021-11-17
Fix crasher on Python 3.9

ProfFan:
Bug/issue #, if applicable: 

## Summary

Toolchain build crashes with

```
Traceback (most recent call last):
  File "/home/fan/AUR/swift-language-git/src/swift-docc/build-script-helper.py", line 295, in <module>
    main()
  File "/home/fan/AUR/swift-language-git/src/swift-docc/build-script-helper.py", line 28, in main
    run(args)
  File "/home/fan/AUR/swift-language-git/src/swift-docc/build-script-helper.py", line 132, in run
    install(args, env)
  File "/home/fan/AUR/swift-language-git/src/swift-docc/build-script-helper.py", line 220, in install
    docc_path = docc_bin_path(
  File "/home/fan/AUR/swift-language-git/src/swift-docc/build-script-helper.py", line 262, in docc_bin_path
    return os.path.join(
  File "/usr/lib/python3.9/posixpath.py", line 90, in join
    genericpath._check_arg_types('join', a, *p)
  File "/usr/lib/python3.9/genericpath.py", line 155, in _check_arg_types
    raise TypeError("Can't mix strings and bytes in path components") from None
TypeError: Can't mix strings and bytes in path components
ERROR: command terminated with a non-zero exit status 1, aborting
```

## Dependencies

None

## Testing

Steps:
1. Build
2. `main` branch should crash at aforementioned line

## Checklist

- [x] Added tests (N/A)


build-script-helper.py
```diff
@@ -260,7 +260,7 @@ def docc_bin_path(swift_exec, args, env, verbose):
   if verbose:
     print(' '.join([escape_cmd_arg(arg) for arg in cmd]))
   return os.path.join(
-    subprocess.check_output(cmd, env=env).strip(), 'docc')
+    subprocess.check_output(cmd, env=env).strip().decode(), 'docc')
```
ethan-kusters, line 263: Should be decoding the bytes to a string before calling `strip()`?

```suggestion
    subprocess.check_output(cmd, env=env).decode().strip(), 'docc')
```

ethan-kusters (commented):
Thank you for posting this @ProfFan!

I think this must be a change in behavior between Python 3.8.9 and Python 3.9. I've verified this is still backwards compatible with Python 3.8.9 which is used in SwiftCI.

Are there any other instances of `os.path.join` that we should be updating with a call to `decode()` here?

ethan-kusters:
I believe this has been resolved with #73. Closing this out for now.

PR #50, opened 2021-12-05
Remove deprecated code

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

- Remove unnessary available check
- Clean up deprecated methods and declaration

Should SwiftDocC target be depended by other clients? It seems we have too much deprecated mark everywhere in that target.

If it should, I think maybe we can have some kind of version release, so that we can one day remove those deprecated code. And I'll adjust the PR to only change the "doc" update and "available check" removal.

If it should only be intended for the repo's internal use, I can we could safely delete them as this PR does.

## Checklist

- [x] Added tests


Kyle-Ye:
@swift-ci test

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
@swift-ci Please test

franklinsch:
Hi @Kyle-Ye, thanks for opening this!

> Should SwiftDocC target be depended by other clients?

Yes, this allows for integrations with other tools and IDEs. For example, using these APIs: https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/Communication.md

Regarding the deprecated public APIs, I personally think it's still too early to remove them as some clients might be relying on them. I would say that we should wait at least a release before removing them.

PS: I see that this branch was pushed to this repository—I don't think we have this currently documented, but I think we should prefer keeping the branches in the `apple/swift-docc` repository to track releases and push development branches to _forks_, to align with the workflow of contributors that don't have write access.

Kyle-Ye:
> Hi @Kyle-Ye, thanks for opening this!
> 
> > Should SwiftDocC target be depended by other clients?
> 
> Yes, this allows for integrations with other tools and IDEs. For example, using these APIs: [`main`/Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/Communication.md](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/Communication.md?rgh-link-date=2021-12-06T11%3A27%3A41Z)
> 
> Regarding the deprecated public APIs, I personally think it's still too early to remove them as some clients might be relying on them. I would say that we should wait at least a release before removing them.
> 
> PS: I see that this branch was pushed to this repository—I don't think we have this currently documented, but I think we should prefer keeping the branches in the `apple/swift-docc` repository to track releases and push development branches to _forks_, to align with the workflow of contributors that don't have write access.

Got it, I'll adjust the PR to only change the "doc" update and "available check" removal.

And sorry for interrupting the branches on the repo, I'll remove it and create a new PR from my fork branch.

PR #73, opened 2022-01-19
Added decode() to fix issue joining strings and bytes

tachoknight:
## Summary 

Building Swift with Fedora Linux which uses Python 3.10, installing swift-docc fails because joining a byte array with a string requires an explicit decode()

## Dependencies
None

## Testing
The command being run by the Fedora build system is generally the same as Swift is built regardless of environment:
`swift/utils/build-script --preset=buildbot_linux,no_test install_destdir=%{_builddir} installable_package=%{_builddir}/swift-%{version}-fedora.tar.gz`

## Checklist

- [ ] Added tests _No additional test due to no additional functionality_
- [ ] Updated documentation if necessary _No changes to functionality_


build-script-helper.py
```diff
@@ -261,7 +261,7 @@ def docc_bin_path(swift_exec, args, env, verbose):
   if verbose:
     print(' '.join([escape_cmd_arg(arg) for arg in cmd]))
   return os.path.join(
-    subprocess.check_output(cmd, env=env).strip(), 'docc')
+    subprocess.check_output(cmd, env=env).strip().decode(), 'docc')
```
ethan-kusters, line 264: Should we be decoding the the bytes to a string before calling `strip()`?
```suggestion
    subprocess.check_output(cmd, env=env).decode().strip(), 'docc')
```

ethan-kusters (commented):
Thanks for opening this @tachoknight!

I've verified this is still backwards compatible with Python 3.8.9 which is used in SwiftCI.

This looks ready-to-go to me; I just had a couple of questions.

Are there any other instances of os.path.join that we should be updating with a call to decode() as part of this PR?

build-script-helper.py
tachoknight: I don't believe so; that was the only spot I found where it was trying to join a byte array and a string.

ethan-kusters:
@swift-ci please test

ethan-kusters (approved):

ethan-kusters:
@swift-ci please test

ethan-kusters:
Thank you @tachoknight!

PR #82, opened 2022-02-08
Add external media support in documentation

Kyle-Ye:
Bug/issue #, if applicable: [SR-15839](https://bugs.swift.org/browse/SR-15839)
Forum discussion: https://forums.swift.org/t/can-we-link-and-render-external-images-that-is-from-urls

## Summary

Add a new ExternalMediaResolver to support external media reference.

## Dependencies

None

## Testing

Test external image and video support in docs and articles.

## Checklist

- [x] Added tests

## TODO
- [x] Currently this PR does not support external media in tutorials. Will solve it in a follow-up PR.

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -2622,14 +2622,21 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
             return localAsset
         }
         
-        if let externallyResolvedAsset = fallbackAssetResolvers[bundleIdentifier].flatMap({
-            $0.resolve(assetNamed: name, bundleIdentifier: bundleIdentifier)
-        }) {
-            assetManagers[bundleIdentifier, default: DataAssetManager()].register(
-                dataAsset: externallyResolvedAsset, forName: name)
+        if let fallbackAssetResolver = fallbackAssetResolvers[bundleIdentifier],
+           let externallyResolvedAsset = fallbackAssetResolver.resolve(assetNamed: name, bundleIdentifier: bundleIdentifier) {
+            assetManagers[bundleIdentifier, default: DataAssetManager()]
+                .register(dataAsset: externallyResolvedAsset, forName: name)
             return externallyResolvedAsset
         }
         
+        // If no fallbackAssetResolver is set, try to treat it as external media link
+        if let externalMediaLink = URL(string: name),
+           externalMediaLink.isAbsoluteWebURL {
+            var asset = DataAsset()
+            asset.context = .display
+            asset.register(externalMediaLink, with: DataTraitCollection(userInterfaceStyle: .light, displayScale: .standard))
+            return asset
+        }
         return nil
     }
     
--- a/Sources/SwiftDocC/Model/Rendering/References/ImageReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ImageReference.swift
@@ -74,9 +74,7 @@ public struct ImageReference: MediaReference, URLReference {
         var result = [VariantProxy]()
         // sort assets by URL path for deterministic sorting of images
         asset.variants.sorted(by: \.value.path).forEach { (key, value) in
-            // If a scheme is present it means it's an absolute URL
-            let isAbsoluteWebURL = !value.isFileURL && value.scheme?.isEmpty == false
-            let url = isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
+            let url = value.isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
             result.append(VariantProxy(url: url, traits: key))
         }
         try container.encode(result, forKey: .variants)
--- a/Sources/SwiftDocC/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
+++ b/Sources/SwiftDocC/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
@@ -1,0 +1,26 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+extension URL {
+    /// Returns true if the scheme is not `nil` , not `file:`, not `ResolvedTopicReference.urlScheme`.
+    ///
+    /// - Returns: A Boolean value indicating whether the url is an absolute web URL.
+    var isAbsoluteWebURL: Bool {
+        guard !isFileURL,
+              let scheme = scheme ,
+              !scheme.isEmpty,
+              scheme != ResolvedTopicReference.urlScheme else {
+            return false
+        }
+        return true
+    }
+}
--- a/Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertFileWritingConsumer.swift
+++ b/Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertFileWritingConsumer.swift
@@ -55,7 +55,7 @@ struct ConvertFileWritingConsumer: ConvertOutputConsumer {
     
     func consume(assetsInBundle bundle: DocumentationBundle) throws {
         func copyAsset(_ asset: DataAsset, to destinationFolder: URL) throws {
-            for sourceURL in asset.variants.values {
+            for sourceURL in asset.variants.values where !sourceURL.isAbsoluteWebURL  {
                 let assetName = sourceURL.lastPathComponent
                 try fileManager.copyItem(
                     at: sourceURL,
--- a/Tests/SwiftDocCTests/Utility/URL+IsAbsoluteWebURLTests.swift
+++ b/Tests/SwiftDocCTests/Utility/URL+IsAbsoluteWebURLTests.swift
@@ -1,0 +1,26 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+@testable import SwiftDocC
+import XCTest
+
+class URL_IsAbsoluteWebURLTests: XCTestCase {
+    func testisAbsoluteWebURL() throws {
+        let localURL = URL(fileURLWithPath: "/Users/username/Documents/Some Folder/Some Document.txt")
+        let topReferenceURL = try XCTUnwrap(URL(string: "doc://swift-doc"))
+        let webURL = try XCTUnwrap(URL(string: "swift.org"))
+        let absoluteWebURL = try XCTUnwrap(URL(string: "https://swift.org"))
+
+        XCTAssertEqual(localURL.isAbsoluteWebURL, false)
+        XCTAssertEqual(topReferenceURL.isAbsoluteWebURL, false)
+        XCTAssertEqual(webURL.isAbsoluteWebURL, false)
+        XCTAssertEqual(absoluteWebURL.isAbsoluteWebURL, true)
+    }
+}
--- a/Sources/SwiftDocC/Model/Rendering/References/VideoReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/VideoReference.swift
@@ -81,9 +81,7 @@ public struct VideoReference: MediaReference, URLReference {
         // convert the data asset to a serializable object
         var result = [VariantProxy]()
         asset.variants.sorted(by: \.value.path).forEach { (key, value) in
-            // If a scheme is present it means it's an absolute URL
-            let isAbsoluteWebURL = !value.isFileURL && value.scheme?.isEmpty == false
-            let url = isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
+            let url = value.isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
             result.append(VariantProxy(url: url, traits: key))
         }
         try container.encode(result, forKey: .variants)
--- a/Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
@@ -62,7 +62,6 @@ public struct DownloadReference: RenderReference, URLReference {
 
 extension DownloadReference {
     private func renderURL(for url: URL) -> URL {
-        let isAbsoluteWebURL = !url.isFileURL && url.scheme?.isEmpty == false && url.scheme != ResolvedTopicReference.urlScheme // if scheme is present it means it's an absolute URL
-        return isAbsoluteWebURL ? url : destinationURL(for: url.lastPathComponent)
+        url.isAbsoluteWebURL ? url : destinationURL(for: url.lastPathComponent)
     }
 }
--- a/Sources/SwiftDocCUtilities/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
+++ b/Sources/SwiftDocCUtilities/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
@@ -1,0 +1,25 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+extension URL {
+    /// Returns true if the scheme is not `nil` , not `file:`.
+    ///
+    /// - Returns: A Boolean value indicating whether the url is an absolute web URL.
+    var isAbsoluteWebURL: Bool {
+        guard !isFileURL,
+              let scheme = scheme ,
+              !scheme.isEmpty else {
+            return false
+        }
+        return true
+    }
+}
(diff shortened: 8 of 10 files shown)


Kyle-Ye:
Implementation is almost done. Will update test and documentation later this week

Sources/SwiftDocC/Model/Rendering/References/ExternalImageReference.swift
```diff
@@ -0,0 +1,102 @@
...
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+/// A reference to an extern image.
+public struct ExternalImageReference: RenderReference {
```
d-ronnqvist, line 14: Would it be possible to use `ImageReference` for both local and external images?

At a glance this looks like an `ImageReference` without a `DataAsset` value. 

I'm wondering if making the `asset` optional (which would have to be made on all MediaReference types) or if passing an empty `DataAsset` (or another value signifying an external asset) could work.

Sources/SwiftDocC/Model/Rendering/RenderContentCompiler.swift
```diff
@@ -77,6 +78,8 @@ struct RenderContentCompiler: MarkupVisitor {
         let imageIdentifier: RenderReferenceIdentifier = .init(unescapedSource)
         if let resolvedImages = context.resolveAsset(named: unescapedSource, in: identifier) {
             imageReferences[unescapedSource] = ImageReference(identifier: imageIdentifier, altText: image.altText, imageAsset: resolvedImages)
+        } else if unescapedSource.contains("://"){ // FIXME: - Maybe there is a better way to check scheme
```
d-ronnqvist, line 81: Is there some way that the context could be responsible for resolving the external reference? 

For example, maybe the context could return an empty asset value for the external asset or the context could have another method that creates and returns the reference.

Would this work if the check is the reference is an external image were moved into `context.resolveAsset(named: unescapedSource, in: identifier)` instead?

d-ronnqvist (changes requested):
I'm wondering if there's a way to implement this without new render reference types. Ideally the solution could be easily shared to also add support for external videos and external downloads (if that's something that we want to support)

Sources/SwiftDocC/Model/Rendering/References/ExternalImageReference.swift
Kyle-Ye: Yes, I think the same at first. But later I thought hacking the current ImageReference may require changing more code. I'll update the implementation to use ImageReference if this is the original design.

franklinsch:
Hi @Kyle-Ye, are you still working on this? Let @d-ronnqvist or I know we can help!

Kyle-Ye:
> Hi @Kyle-Ye, are you still working on this? Let @d-ronnqvist or I know we can help!

Yes. Sorry for the late update (Busy with my full-time iOS RD internship recently)

I have a few question here in order to better finish this PR and your review.

After performing "docc preview" action to the xx.docc file, it will produce a `xx.docc/.docc-build/data/documentation/swiftdoccrender.json` file

For example if I have this markdown file in my xx.docc folder
```
# Swift-DocC-Render

@Metadata {
  @TechnologyRoot
}

A web renderer for DocC Archives produced by DocC.

## Overview

Local image

![swift icon](swift.svg)

External image

![swift icon](https://www.swift.org/assets/images/swift.svg)

DocC-Render displays documentation produced by [DocC](https://www.swift.org/documentation/docc) on the web and is included in DocC Archives.
```
It will produce something like this in reference part
```json
{
    "https://www.swift.org/documentation/docc": {
        "title": "DocC",
        "titleInlineContent": [
            {
                "type": "text",
                "text": "DocC"
            }
        ],
        "type": "link",
        "identifier": "https://www.swift.org/documentation/docc",
        "url": "https://www.swift.org/documentation/docc"
    },
    // MARK: - This is the current behavior. Without the PR, this item will not exist.
    "https://www.swift.org/assets/images/swift.svg": {
        "alt": "swift icon",
        "type": "externalImage",
        "identifier": "https://www.swift.org/assets/images/swift.svg",
        "variants": [
            {
                "url": "https://www.swift.org/assets/images/swift.svg",
                "traits": [
                    "1x",
                    "light"
                ]
            }
        ]
    },
    "swift.svg": {
        "alt": "swift icon",
        "type": "image",
        "identifier": "swift.svg",
        "variants": [
            {
                "url": "/images/swift.svg",
                "traits": [
                    "1x",
                    "light"
                ]
            }
        ]
    }
}
```
I think "d-ronnqvist" is suggesting me to generate an item like this without introducing a new externalImage type, right?
```json
    "https://www.swift.org/assets/images/swift.svg": {
        "alt": "swift icon",
        "type": "image",
        "identifier": "https://www.swift.org/assets/images/swift.svg",
        "variants": [
            {
                // MARK: Problem here, since in current design, for image type it'll do some magical logic here, how can we identity this is actually an external image? Do we need introduce a new field or just use host == nil?
                "url": "https://www.swift.org/assets/images/swift.svg",
                "traits": [
                    "1x",
                    "light"
                ]
            }
        ]
    },
```

Sources/SwiftDocC/Model/Rendering/References/ExternalImageReference.swift
Kyle-Ye: Is `fallbackAssetResolvers` the right direction for solving this problem?

Currently fallbackAssetResolvers is an empty array. Following your suggestion, I think I should create a class which conforms to `FallbackAssetResolver` and register it at the proper time.

https://github.com/apple/swift-docc/blob/0d61dc5550c9bdb04fd1579144679e7d3287f640/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2541-L2562

Kyle-Ye:
Currently, I was stuck by the "display" and "download" context in DataAsset. Should externalImage a "display" context or "download" context? And what is "download" context mean in docc?

> The current implementation will throw an error due to "copyAsset" API calling `FileManager.copyItem(at:to:)` where "at" is a https schema file. Will fix this once I figure what "download" context really means.
> `error: The file “swift.svg” couldn’t be opened because URL type https isn’t supported.`

https://github.com/apple/swift-docc/blob/0d61dc5550c9bdb04fd1579144679e7d3287f640/Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertFileWritingConsumer.swift#L70-L101

Appreciate for your help and clarification @franklinsch

d-ronnqvist:
I’ll look at this tomorrow

Sources/SwiftDocC/Model/Rendering/References/ExternalImageReference.swift
d-ronnqvist: > Is `fallbackAssetResolvers` the right direction for solving this problem?

A fallback resolver will behave correctly but it's a public property and it's possible to customize. So it could happen that a developer would unintentionally disable this behavior if they _assigned_ the `fallbackAssetResolvers` property instead of _appending_ to it. 

We haven't documented the default values for any of the public resolver related properties nor have we documented how developers should interact with them. Personally I feel that external image support should be considered a built in feature that can't be disabled but I don't hold that opinion strongly.

d-ronnqvist:
> Currently, I was stuck by the "display" and "download" context in DataAsset. Should externalImage a "display" context or "download" context? And what is "download" context mean in docc?

From what I recall, this is an indication to Swift-DocC Render that has some impact on how the request for that access is made. @mportiz08 May be able to provide more information about how this information is used.

A "download" context is used for zip archives and similar files that are downloaded but not displayed on the page and a "display" context is used for images and video that are displayed on the page. So in this case I would expect a "display" context.

mportiz08:
> > Currently, I was stuck by the "display" and "download" context in DataAsset. Should externalImage a "display" context or "download" context? And what is "download" context mean in docc?
> 
> From what I recall, this is an indication to Swift-DocC Render that has some impact on how the request for that access is made. @mportiz08 May be able to provide more information about how this information is used.
> 
> A "download" context is used for zip archives and similar files that are downloaded but not displayed on the page and a "display" context is used for images and video that are displayed on the page. So in this case I would expect a "display" context.

`"display"` sounds right to me. It seems that [this enum](https://github.com/apple/swift-docc/blob/0661e969a2cd683ea67b0612dc3075a60fbcf6f8/Sources/SwiftDocC/Infrastructure/Bundle%20Assets/DataAssetManager.swift#L159) is used to determine whether a particular asset should be turned into a media or download reference in the Render JSON: https://github.com/apple/swift-docc/blob/0661e969a2cd683ea67b0612dc3075a60fbcf6f8/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift#L1288

This will better indicate the type of reference and how it should be displayed or linked to for any renderers, including DocC-Render. The `"download"` value might be used for something like a zip file to download sample code, etc. Note that DocC-Render might support rendering content that DocC itself is not yet generating at the moment.

Kyle-Ye:
This PR adds external media support in documentation(docs and articles) but not in tutorials.

Should we add support external media support for tutorials? And if we should maybe we can do it in a follow-up PR? cc @franklinsch

Sources/SwiftDocC/Model/Rendering/References/ExternalImageReference.swift
Kyle-Ye: Yes, I found that after I set a fallbackAssetResolvers manually with my `ExternalMediaResolver`, the `testConvertPageWithLinkResolving` test will fail.
```
if let linkResolvingServer = linkResolvingServer {
    let resolver = try OutOfProcessReferenceResolver(
        bundleIdentifier: request.bundleInfo.identifier,
        server: linkResolvingServer,
        convertRequestIdentifier: messageIdentifier
    )
    
    context.fallbackReferenceResolvers[request.bundleInfo.identifier] = resolver
    context.fallbackAssetResolvers[request.bundleInfo.identifier] = resolver
    context.externalSymbolResolver = resolver
}
```
So should we add a new `externalAssetResolvers` property or just hardcode it at someplace to just ignore the https and http external image(Which is really a bad idea). How can we determine if a given schema is external accessible or something that a reserved  like `docc://image.png` and `docs-media:///path/to/image.png`

franklinsch:
Supporting external images in tutorials in a separate PR sounds good to me. Please make sure to file a ticket for it on bugs.swift.org to make sure we don't lose track of it :)

Sources/SwiftDocC/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
```diff
@@ -0,0 +1,20 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021 Apple Inc. and the Swift project authors
```
d-ronnqvist, line 4:
```suggestion
 Copyright (c) 2022 Apple Inc. and the Swift project authors
```
(since these are new files)

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -1564,6 +1564,7 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
         let miscResources = Set(bundle.miscResourceURLs)
         try assetManagers[bundle.identifier, default: DataAssetManager()]
             .register(data: miscResources, dataProvider: dataProvider, bundle: bundle)
+        fallbackAssetResolvers[bundle.identifier] = ExternalMediaResolver()
```
d-ronnqvist, line 1567: Thank you. This resolves my previous feedback about potentially surprising behaviors with default values for fallback resolvers.

d-ronnqvist (commented):
Thank you. This looks good to me.

I left a few comments and suggestions about some of the details.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertFileWritingConsumer.swift
```diff
@@ -55,7 +55,7 @@ struct ConvertFileWritingConsumer: ConvertOutputConsumer {
     
     func consume(assetsInBundle bundle: DocumentationBundle) throws {
         func copyAsset(_ asset: DataAsset, to destinationFolder: URL) throws {
-            for sourceURL in asset.variants.values {
+            for sourceURL in asset.variants.values.filter({ !$0.isAbsoluteWebURL() }) {
```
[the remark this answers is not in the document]
Kyle-Ye: Sound great.

Tests/SwiftDocCTests/Utility/URL+IsAbsoluteWebURLTests.swift
```diff
@@ -0,0 +1,28 @@
...
+*/
+
+import XCTest
+@testable import SwiftDocC
+
+class URL_IsAbsoluteWebURLTests: XCTestCase {
+
+    func testIsAbsoluteWebURL() throws{
+        let localURL = URL(fileURLWithPath: "/Users/username/Documents/Some Folder/Some Document.txt")
+        let topReferenceURL = try XCTUnwrap(URL(string: "docc://swift-doc"))
```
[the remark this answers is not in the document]
Kyle-Ye: Yes. My mistake here

Sources/SwiftDocC/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
Kyle-Ye: Got it

Sources/SwiftDocC/Utility/FoundationExtensions/URL+IsAbsoluteWebURL.swift
```diff
@@ -0,0 +1,20 @@
...
+*/
+
+import Foundation
+
+extension URL {
+    /// If a scheme is present it means it's an absolute URL
+    ///
+    /// - Returns: A Boolean value indicating whether the url is an absolute web URL.
+    func isAbsoluteWebURL() -> Bool {
+        !isFileURL && scheme?.isEmpty == false
```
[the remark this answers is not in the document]
Kyle-Ye: I agree.

Also I think maybe `isAbsoluteWebURL` should be a computed var instead of function to align with `isFileURL` API.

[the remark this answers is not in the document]
Kyle-Ye: > * Would it make sense to check that this isn't a topic reference `url.scheme != ResolvedTopicReference.urlScheme`?

We can't do it in SwiftDocCUtilities unless we hardcode the "doc" string.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
Kyle-Ye: Currently this will break one test case(testConvertPageWithLinkResolving) since we will set fallbackAssetResolvers when linkResolvingServer is not nil.

https://github.com/apple/swift-docc/blob/783cc969a0185373c92eb864c04c51b689cc978a/Sources/SwiftDocC/DocumentationService/Convert/ConvertService.swift#L148-L158

Should we move the external image functionality into a default behavior instead of injecting it into fallbackAssetResolvers? 👇
Which means we move `ExternalMediaResolver.swift`'s resolve logic into `resolveAsset(named name: String, in parent: ResolvedTopicReference)` directly.

https://github.com/apple/swift-docc/blob/783cc969a0185373c92eb864c04c51b689cc978a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2625-L2640

(Add the following logic into L2630 instead of `ExternalMediaResolver.swift` and registering it as a FallbackAssetResolver)
```swift
    func resolve(assetNamed assetName: String, bundleIdentifier: String) -> DataAsset? {
        guard let externalMediaLink = URL(string: assetName),
              externalMediaLink.isAbsoluteWebURL else {
            return nil
        }
        var asset = DataAsset()
        asset.context = .display
        asset.register(externalMediaLink, with: DataTraitCollection(userInterfaceStyle: .light, displayScale: .standard))
        return asset
    }
```

d-ronnqvist: > Should we move the external image functionality into a default behavior instead of injecting it into fallbackAssetResolvers? 👇
> Which means we move ExternalMediaResolver.swift's resolve logic into resolveAsset(named name: String, in parent: ResolvedTopicReference) directly.

That sounds good to me 👍

Kyle-Ye:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/References/ImageReference.swift
```diff
@@ -74,9 +74,7 @@ public struct ImageReference: MediaReference, URLReference {
         var result = [VariantProxy]()
         // sort assets by URL path for deterministic sorting of images
         asset.variants.sorted(by: \.value.path).forEach { (key, value) in
-            // If a scheme is present it means it's an absolute URL
-            let isAbsoluteWebURL = !value.isFileURL && value.scheme?.isEmpty == false
-            let url = isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
+            let url = value.isAbsoluteWebURL ? value : destinationURL(for: value.lastPathComponent)
```
Kyle-Ye, line 77: URL.isAbsoluteWebURL delete the host check, otherwise it will fail on some test due to the case like 'docc-media:///path/a.png'. We can fix it by not using the URL.isAbsoluteWebURL API here. And I think it's better to use one uniformed API, so I delete the host check in URL.isAbsoluteWebURL. Any suggestion is welcomed.

Kyle-Ye:
Finally fix all the test cases. Ready for the final review. @d-ronnqvist

Kyle-Ye:
> Supporting external images in tutorials in a separate PR sounds good to me. Please make sure to file a ticket for it on bugs.swift.org to make sure we don't lose track of it :)

Update the original SR's title to "Add external media support for DocC documentation" and file up a new ticket to track the issue. https://bugs.swift.org/browse/SR-16026

Sources/SwiftDocC/Model/Rendering/References/ImageReference.swift
d-ronnqvist: I agree that it's better to use the same API everywhere. 

We can revisit the other tests' test data in the future to see if anything should be updated but I don't think that needs to happen now.

d-ronnqvist (approved):
Thank you! This looks great to me.

Kyle-Ye:
@swift-ci please test and merge

Kyle-Ye:
@swift-ci please test

PR #83, opened 2022-02-08
[NFC] Optimus some unchangeable member to be "let" instead of "var"

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

Optimus some unchangeable member to be "let" instead of "var"

## Dependencies

None

## Testing

Non-Functional Change

## Checklist

### DIFF
--- a/Sources/SwiftDocC/Model/AnchorSection.swift
+++ b/Sources/SwiftDocC/Model/AnchorSection.swift
@@ -14,8 +14,8 @@ import Markdown
 /// Section of a document that can be linked to.
 public struct AnchorSection {
     /// The unique reference to the section.
-    public var reference: ResolvedTopicReference
+    public let reference: ResolvedTopicReference
     
     /// The title of the section.
-    public var title: String
+    public let title: String
 }
--- a/Sources/SwiftDocC/Model/BuildMetadata.swift
+++ b/Sources/SwiftDocC/Model/BuildMetadata.swift
@@ -21,10 +21,10 @@ public struct BuildMetadata: Codable {
     )
     
     /// The display name of the documentation bundle that DocC built.
-    public var bundleDisplayName: String
+    public let bundleDisplayName: String
     
     /// The bundle identifier of the documentation bundle that DocC built.
-    public var bundleIdentifier: String
+    public let bundleIdentifier: String
     
     /// Creates a build metadata value for a documentation bundle built by DocC.
     ///
--- a/Sources/SwiftDocC/Model/DocumentationNode.swift
+++ b/Sources/SwiftDocC/Model/DocumentationNode.swift
@@ -21,10 +21,10 @@ public struct DocumentationNode {
     public var reference: ResolvedTopicReference
     
     /// The type of node.
-    public var kind: Kind
+    public let kind: Kind
     
     /// The programming language in which the node is relevant.
-    public var sourceLanguage: SourceLanguage
+    public let sourceLanguage: SourceLanguage
     
     /// All the languages in which the node is available.
     public var availableSourceLanguages: Set<SourceLanguage>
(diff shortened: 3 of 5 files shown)


Kyle-Ye:
@swift-ci please test

Kyle-Ye:
> My take on this is that since these types are structs, mutability is already controlled by the client, and making their properties `var`s helps clients manipulate/re-assign these values (especially useful when writing unit tests, for example).

Yes, I agree on this. But I think the default behavior should be minimal permission which is  "private" and "let" unless we have a solid reason to do it. That is said, if we someday later find some member really needs to be modified, we can then change it to "var"(Just like what we do to remove "private" sometimes).

Moreover there is something like "@testable import and spi" in swift to better constraint the access control model (Making it accessible in test or other module without making it totally public or open).

I was wondering if there is something similar in get/set model. Currently we only have "private(set)" which seems not flexible enough in some cases. Is there a grammar which allows to "set" in test module or on specific clients in swift?(Like testableSet or something)

Kyle-Ye:
Although currently there is no code in swift-docc modify those members. But other package which depends on swift-docc may use them and potentially modify them in their existing codebase since those members are public.

So this should be a source breaking change 😂

Did not know if you are worried about this?

franklinsch:
To maintain compatibility with existing clients, we'd deprecate the setter of the property and remove it down the line. Since we do expect clients to import the `SwiftDocC` library directly (e.g., IDE projects), I do think the added flexibility of keeping these as `var`s helps. 

Making the property `internal(set)` would still allow Swift-DocC's tests to mutate those properties (via `@testable import SwiftDocC`), but I believe there are valid cases beyond testing where flexibility would be beneficial for clients. Another idea is to use the `@_spi` annotation to prevent access to clients by default, but I think that would be an over-complication and not valuable in this case.

Once reason for making these `let`s would be if it would allow clients to set a value that violates a precondition enforced by the implementation of the type. For example, if we had:
```swift
public struct SourceLanguage: Hashable, Coddle {
  public var name: String

  public init(name: String) {
    precondition(!name.isEmpty)
    self.name = name
  }

  public var firstCharacterOfName: Character {
    name.first!
  } 
}
```
in that case we wouldn't make to make it a `var` because the client would be able to set `name` to an empty string and cause a crash in the computed property.

Kyle-Ye:
> To maintain compatibility with existing clients, we'd deprecate the setter of the property and remove it down the line. Since we do expect clients to import the `SwiftDocC` library directly (e.g., IDE projects), I do think the added flexibility of keeping these as `var`s helps.
> 
> Making the property `internal(set)` would still allow Swift-DocC's tests to mutate those properties (via `@testable import SwiftDocC`), but I believe there are valid cases beyond testing where flexibility would be beneficial for clients. Another idea is to use the `@_spi` annotation to prevent access to clients by default, but I think that would be an over-complication and not valuable in this case.
> 
> Once reason for making these `let`s would be if it would allow clients to set a value that violates a precondition enforced by the implementation of the type. For example, if we had:
> 
> ```swift
> public struct SourceLanguage: Hashable, Coddle {
>   public var name: String
> 
>   public init(name: String) {
>     precondition(!name.isEmpty)
>     self.name = name
>   }
> 
>   public var firstCharacterOfName: Character {
>     name.first!
>   } 
> }
> ```
> 
> in that case we wouldn't make to make it a `var` because the client would be able to set `name` to an empty string and cause a crash in the computed property.

I'm very sorry if I get you wrong. Still did not know your conclusion.

So your opinion is not change the code or suggest that we should change from `public var xx` to `public internal(set) var xx`?

franklinsch:
Sorry about that, I should have made it clearer. I think we should keep the code as-is (`public var xxx`) for added flexibility.

Kyle-Ye:
Got it.

PR #84, opened 2022-02-09
Emit a JSON representation of the navigator index

ethan-kusters:
Bug/issue #, if applicable: [rdar://88032854](rdar://88032854)

## Summary

Adds support for emitting a JSON representation of a DocC archive's navigator index.

This JSON representation can be consumed by a wider variety of clients than the current LMDB representation that DocC emits. However, the JSON representation currently contains less data than the LMDB database (for example, it's missing availability information) so it's (at least initially) meant to be complementary to the LMDB version.

## Details

This PR is broken into four commits:

- [ca50354]: Adds a `--enable-experimental-json-index` feature flag to gate this feature.
- [7ea313a]: Only create an on-disk LMDB navigator environment when needed

    Moves the creation of the LMDB environment to the `finalize()` function where we write out the navigator data to disk.

    This means we only write to disk when is actually needed and if `finalize()` is never called, we won't unnecessarily create an an empty LMDB database on disk.

    This also prepares for a world where we have multiple output formats of navigator index data.

- [f88a681]: Include a string id of the interface language

    Updates the `InterfaceLanguage` model used by the NavigatorIndex to include a string identifier in addition to the existing, presentation-friendly, name.

- [db443512]: This is the core of the PR and adds a support for emitting the JSON representation of the Index.

    The index will be emitted to the index subdirectory of a docc archive
as `index.json`. For example:

    ```
    MyFramework.doccarchive/index/index.json
    ```

## Testing

```
docc convert <path-to-catalog> --enable-experimental-json-index
```

Inspect the output at `/index/index.json`.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Indexing/IndexJSON/Index.swift
```diff
@@ -0,0 +1,188 @@
...
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import SymbolKit
+
+public struct Index: Codable {
```
franklinsch, line 13: The `Index` vs `NavigatorIndex` naming isn't totally clear to me. Can we use the same model for LMDB and JSON outputs (and abstract the serialisation logic), and if not, why? If `Index` is an implementation detail to serialise things to JSON, maybe we should keep it `internal`.

Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
```diff
@@ -395,12 +395,12 @@ Root
         assertUniqueIDs(node: navigatorIndex.navigatorTree.root)
         assertEqualDumps(navigatorIndex.navigatorTree.root.dumpTree(), """
         [Root]
-        ┣╸Objective-C
+        ┣╸occ
```
franklinsch, line 398: Why this change and can we avoid it? I think this will break clients that use that node's name in UI.

Sources/SwiftDocC/SwiftDocC.docc/Resources/Index.spec.json
```diff
@@ -0,0 +1,119 @@
...
+                            "propertyListKeyReference",
+                            "protocol",
+                            "resources",
+                            "root",
+                            "sampleCode",
+                            "section",
+                            "struct",
+                            "subscript",
+                            "symbol",
+                            "type.method",
```
dobromir-hristov, line 76: I have a `tutorial` listed in my code as well. Should I remove it?

ethan-kusters: Currently we use the `"project"` to describe tutorials in both RenderJSON and this new Index spec. It can't hurt to map both `"project"` and `"tutorial"` to the same thing- but I wouldn't expect it to be emitted so you shouldn't need to support it.

Sources/SwiftDocC/Indexing/IndexJSON/Index.swift
ethan-kusters: I agree this is confusing. Ideally- I'd like to break things out here so we'd have something like:

1. NavigatorIndex
2. JSONNavigatorIndexReaderWriter
3. LMDBNavigatorIndexReaderWriter

And the JSON representation would be totally inter-changeable with the LMDB representation.

I looked into doing that now but I don't have bandwidth for that full refactor with the way the code is currently architected.

> If Index is an implementation detail to serialise things to JSON, maybe we should keep it internal.

I'm not opposed to that but there is some value in letting clients adopt the API for parsing.

Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
ethan-kusters: Good call- I reworked things to avoid this change.

ethan-kusters:
This is now ready for review. I've updated the existing PR description to reflect the more-final implementation here.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -647,7 +648,6 @@ extension NavigatorIndex {
                     availabilityToID[entryIDs] = newID
                     availabilityIDs[newID] = entryIDs
                     availabilityID = newID
-                    try navigatorIndex.availability.put(key: newID, value: entryIDs)
```
ethan-kusters, line 650: Actually putting the availability information into the LMDB database was moved to the `finalize()` function where the rest of the database writing logic is.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -881,6 +901,134 @@ extension NavigatorIndex {
...
+                    )
+                }
+                
+            }
+            
+            guard emitLMDBRepresentation else {
+                return
+            }
+            
+            let environment: LMDB.Environment
```
ethan-kusters, line 934: This next chunk of code handles the creation of the LMDB environment with some unfortunate do/catch logic that's necessary because `finalize()` is non-throwing.

The actual LMDB specifics have not changed and are just moved from the `fileprivate init` that used to set this up ahead of time.

ethan-kusters:
@swift-ci please test

Sources/SwiftDocC/Indexing/IndexJSON/Index.swift
```diff
@@ -0,0 +1,282 @@
...
+/// The structure of this data is determined by the topic groups authored in
+/// the processed documentation content.
+///
+/// This is an alternative representation of the data that is also handled by the
+/// ``NavigatorIndex`` in a ``NavigatorTree``. This index is specifically designed to be emitted
+/// to disk as JSON file and implements the Index JSON spec.
+///
+/// An OpenAPI specification for Index is available in the repo at
+/// `Sources/SwiftDocC/SwiftDocC.docc/Resources/Index.spec.json`.
+public struct Index: Codable, Equatable {
```
franklinsch, line 24: This is too vague IMO. `RenderIndex`?

ethan-kusters: Agreed! I couldn't think of anything better though. `RenderIndex` makes a lot of sense to me. Thank!

Updated here: 6075a86.

franklinsch (approved):

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

PR #93, opened 2022-02-24
Fix non-deterministic default selectors

franklinsch:
Bug/issue #, if applicable: rdar://89355535

## Summary

 Fixes an issue where DocC picks the default selector for a unified symbol in a non-deterministic way. The new logic picks the default selector that has the most common platforms with other selectors, so that when documentation data variants are collected, the symbols for which we have the most data are used.

Note: the first commit just moves the `UnifiedSymbolGraph.Symbol` into its own file, and the second commit implements the bug fix.

## Dependencies

None.

## Testing

Convert a documentation catalog with multiple symbol graphs for different platforms, and verify that documentation variants from non-Swift symbol graphs is consistently emitted in the output render JSON.

## Checklist

- [x] Added tests


franklinsch:
@swift-ci please test

QuietMisdreavus (approved):
Thanks! I'm curious about the potential performance impact of doing this sorting here, but i also feel like we won't run into situations with a whole lot of selectors to choose from, at least at the moment. I think this is good to land as-is.

franklinsch:
That's a great point. I verified that there is no performance impact for single-language catalogs, but we'll want to make sure that multi-language builds aren't impacted before we enable the feature by default.

PR #95, opened 2022-03-02
Use targetFallback in inheritsFrom relationships

daniel-grumberg:
Bug/issue #, if applicable: rdar://80090675

## Summary

If the super class is in another framework, we can not get a
ResolvedTopicReference to create the inheritsFrom relationship. In this case we
use the targetFallback to populate the render node.
## Dependencies

None

## Testing

Build documentation for a symbol that inherits from a class in another framework and verify that the "Inherits From" section in the generated doccarchive contains the name of the super class.

## Checklist

- [X] Added tests


franklinsch (approved):
Looks great, thanks!

PR #81, opened 2022-02-07
Emit language variants for articles

franklinsch:
Bug/issue #, if applicable: rdar://88464797

## Summary

Articles that document a module available in multiple languages now inherit that module's languages. This ensures that we generate multiple language variants for article pages.

While investigating the performance impact of this change, I found that the repeated accesses to the `DocumentationContext.rootModules` (a computed property that iterates over all topic graph nodes) was causing a large performance regression. To resolve this, in the second commit I made the property a stored property instead that's initialised once during bundle registration, right after symbols have been loaded in the symbol graph.

### Performance impact

No statistically significant impact.

```
Size 10
+-----------------------------------------------------------------------------------------------------+
| Metric                                   | Change     | Before               | After                |
+-----------------------------------------------------------------------------------------------------+
| Compiled output size (MB)                | no change  | 114.68               | 114.68               |
| Duration for 'bundle-registration' (sec) | +1.58%     | 5.76                 | 5.85                 |
| Duration for 'convert-action' (sec)      | +2.93%     | 7.78                 | 8.00                 |
| Peak memory footprint (MB)               | -1.18%     | 354.33               | 350.14               |
| Topic Anchor Checksum                    | no change  | c8f03d4e81da8e3b6d7f | c8f03d4e81da8e3b6d7f |
| Topic Graph Checksum                     | no change  | edce22bce4f8de475a5e | edce22bce4f8de475a5e |
+-----------------------------------------------------------------------------------------------------+

Size 25
+-----------------------------------------------------------------------------------------------------+
| Metric                                   | Change     | Before               | After                |
+-----------------------------------------------------------------------------------------------------+
| Compiled output size (MB)                | no change  | 287.62               | 287.62               |
| Duration for 'bundle-registration' (sec) | +0.26%     | 14.75                | 14.79                |
| Duration for 'convert-action' (sec)      | -0.03%     | 20.09                | 20.08                |
| Peak memory footprint (MB)               | -0.68%     | 816.43               | 810.91               |
| Topic Anchor Checksum                    | no change  | 9a675b9ad6d69f8b7f0c | 9a675b9ad6d69f8b7f0c |
| Topic Graph Checksum                     | no change  | 665713509084b5131a37 | 665713509084b5131a37 |
+-----------------------------------------------------------------------------------------------------+

Size 5
+-----------------------------------------------------------------------------------------------------+
| Metric                                   | Change     | Before               | After                |
+-----------------------------------------------------------------------------------------------------+
[...]

PR #96, opened 2022-03-04
Filter out topics that aren't available in the page's language

franklinsch:
Bug/issue #, if applicable: rdar://89124291

## Summary

Filters out topics that aren't available in the language of the symbol that curates them. Note that this was already working for automatically curated topics, but not manually curated topics.

This PR also includes making tutorial content available on all the languages in which the framework is available (last commit),  

Also, since the original fix of this PR filters out topics that aren't available in the parent page's language, it was also causing DocC to filter out tutorial content from Objective-C pages. As such, the last commit of this PR resolves this issue by making tutorials multi-language as well, if the framework it's documenting is multi-language.

## Dependencies

None.

## Testing

1. Create symbol graph files with symbols that are only available in Swift and symbols only available in Objective-C, and curate them manually in a documentation extension file or in documentation comments.
2. Verify that the rendered documentation doesn't show the Objective-C–only symbols when browsing the Swift documentation, and vice-versa.

## Checklist

- [x] Added tests

## TODO

- [x] Make Tutorials multi-language so that they can be curated in Objective-C pages. They're currently hardcoded to be Swift-only


franklinsch:
@swift-ci please test

franklinsch:
@swift-ci please test

franklinsch:
@swift-ci please test

Sources/SwiftDocC/Utility/DataStructures/GroupedSequence.swift
```diff
@@ -0,0 +1,90 @@
...
+    
+    /// A closure that transforms an element into its key.
+    private var deriveKey: (Element) -> Key
+    
+    var description: String {
+        storage.values.description
+    }
+    
+    /// Creates an empty grouped sequence.
+    init(deriveKey: @escaping (Element) -> Key) {
```
daniel-grumberg, line 52: Nit: typo `derivedKey`

Sources/SwiftDocC/Utility/DataStructures/GroupedSequence.swift
```diff
@@ -0,0 +1,90 @@
...
+/// ```swift
+/// for item in groupedSequence {
+///     print(item)
+/// }
+///
+/// // Prints "aa"
+/// // Prints "a"
+/// // Prints "aaa"
+/// ```
+struct GroupedSequence<Key: Hashable, Element>: Sequence, CustomStringConvertible {
```
daniel-grumberg, line 41: Not needed for this PR but this already fulfills the requirements for `Collection` might be nice since we are making this a standalone data structure.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -823,6 +850,20 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 return ContentRenderSection(kind: .content, content: discussionContent, heading: "Discussion")
             }
             
+            /// Returns whether the topic with the given identifier is available in one of the traits in `allowedTraits`.
+            func isAvailableInAllowedTrait(identifier: String) -> Bool {
+                guard let reference = contentCompiler.collectedTopicReferences[identifier] else {
```
daniel-grumberg, line 855: Not sure I understand this logic. Could you explain why you return true if you couldn't find a topic reference for this identifier?

Tests/SwiftDocCTests/Utility/DataStructures/GroupedSequenceTests.swift
```diff
@@ -0,0 +1,48 @@
...
+class GroupedSequenceTests: XCTestCase {
+    /// A grouped sequence of strings, with their number of characters as the key.
+    var groupedSequence = GroupedSequence<Int, String>(deriveKey: \.count)
+    
+    func testAppends() {
+        groupedSequence.append("a")
+        groupedSequence.append("aa")
+        
+        XCTAssertEqual(groupedSequence[1], "a")
+        XCTAssertEqual(groupedSequence[2], "aa")
```
daniel-grumberg, line 23: you probably want to add the following to this:
```
groupedSequence.append("b")
XCTAssertEqual(groupedSequence[1], "b")
```

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
daniel-grumberg: Is this because you are conservatively not filtering out content you don't have a definition for in this framework?

Sources/SwiftDocC/Utility/DataStructures/GroupedSequence.swift
franklinsch: I think `deriveKey` as it is currently works better because its type is a function, so it should be name like an action.

Sources/SwiftDocC/Utility/DataStructures/GroupedSequence.swift
franklinsch: hm, yes, looks like we're close! But we need to implement a few more requirements, which I think we should implement when we get a use for them. This isn't intended to be a general-purpose API—it's just for DocC's purposes.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
franklinsch: Ah yes, I should clarify that in a comment, thanks for pointing that out. It's expected for `reference` to never be `nil` as it is a precondition for `contentCompiler` to have populated `collectionTopicReference` with the reference at this point. Otherwise, it is a bug in `contentCompiler`.

Tests/SwiftDocCTests/Utility/DataStructures/GroupedSequenceTests.swift
franklinsch: Yes, that's a great idea!

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
franklinsch: In fact, we can actually add an [`assertionFailure`](https://developer.apple.com/documentation/swift/1539616-assertionfailure) so that we get a crash in debug builds, but not release.

franklinsch: Actually, after taking a closer look, I'm wrong on this. It's possible for the content compiler to generate a reference identifier without registering it in `collectedTopicReferences`, for example when the reference is a non-doc URL (e.g., `https://` in which case it gets registered in `linkReferences` instead. I'll add a comment to clarify that.

franklinsch:
@swift-ci please teset

franklinsch:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -621,9 +624,19 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
...
                 
-                let sourceLanguage = SourceLanguage.swift
-                let node = DocumentationNode(reference: technologyResult.topicGraphNode.reference, kind: .technology, sourceLanguage: sourceLanguage, name: .conceptual(title: technology.intro.title), markup: technology.originalMarkup, semantic: technology)
-                documentationCache[technologyResult.topicGraphNode.reference] = node
+                let reference = technologyResult.topicGraphNode.reference.withSourceLanguages(sourceLanguages)
+                
+                let node = DocumentationNode(
+                    reference: reference,
+                    kind: .technology,
+                    sourceLanguage: Self.defaultLanguage(in: sourceLanguages),
```
ethan-kusters, line 632: Nit since I don't think this is directly related to this PR. But this should this be `defaultSourceLanguage`?

```suggestion
                    defaultSourceLanguage: Self.defaultLanguage(in: sourceLanguages),
```

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -621,9 +624,19 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
...
+                    availableSourceLanguages: sourceLanguages,
+                    name: .conceptual(title: technology.intro.title),
+                    markup: technology.originalMarkup,
+                    semantic: technology
+                )
+                documentationCache[reference] = node
+                topicGraph.nodes[reference]?.reference = reference
```
ethan-kusters, line 639: Is this a change from before? We weren't including technology references in the topic graph?

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -621,9 +624,19 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
                 // Add to document map
                 documentLocationMap[url] = technologyResult.topicGraphNode.reference
                 
-                let sourceLanguage = SourceLanguage.swift
-                let node = DocumentationNode(reference: technologyResult.topicGraphNode.reference, kind: .technology, sourceLanguage: sourceLanguage, name: .conceptual(title: technology.intro.title), markup: technology.originalMarkup, semantic: technology)
-                documentationCache[technologyResult.topicGraphNode.reference] = node
+                let reference = technologyResult.topicGraphNode.reference.withSourceLanguages(sourceLanguages)
```
ethan-kusters, line 627: Readability nit:
```suggestion
                let technologyReference = technologyResult.topicGraphNode.reference.withSourceLanguages(sourceLanguages)
```

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -621,9 +624,19 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
...
+                
+                let node = DocumentationNode(
```
ethan-kusters, line 629: Readability nit:
```suggestion
                let technologyNode = DocumentationNode(
```

Probably partly due to the GitHub UI but this diff was generally hard for me to follow because we have a lot of otherwise unspecified `node`s and `reference`.

Sources/SwiftDocC/Infrastructure/Topic Graph/TopicGraph.swift
```diff
@@ -70,7 +70,7 @@ struct TopicGraph {
         }
         
         /// The reference to the `DocumentationNode` this node represents.
-        let reference: ResolvedTopicReference
+        var reference: ResolvedTopicReference
```
ethan-kusters, line 73: It seems strange to me to make only the reference mutable. Should we just be making a new DocumentationNode when we need to modify the reference? That seems like a core part of defines the node.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -129,6 +129,9 @@ public struct RenderNodeTranslator: SemanticVisitor {
         node.hierarchy = hierarchy.hierarchy
         node.metadata.category = technology.name
         
+        let documentationNode = try! context.entity(with: identifier)
```
ethan-kusters, line 132: Definitely out-of-scope for this PR but it seems like the `visitTutorial` method should `throw` at this point.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -823,6 +850,22 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 return ContentRenderSection(kind: .content, content: discussionContent, heading: "Discussion")
             }
             
+            /// Returns whether the topic with the given identifier is available in one of the traits in `allowedTraits`.
+            func isAvailableInAllowedTrait(identifier: String) -> Bool {
```
ethan-kusters, line 854:
```suggestion
            func topicIsAvailableInAllowedTraits(identifier topicIdentifier: String) -> Bool {
```

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1011,7 +1068,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                     DeclarationsSectionTranslator(),
                     ReturnsSectionTranslator(),
                     ParametersSectionTranslator(),
-                    DiscussionSectionTranslator()
+                    DiscussionSectionTranslator(),
```
ethan-kusters, line 1071: 😃

ethan-kusters (commented):
This looks good to me! Just a few small things for your consideration.

Was tutorials the last thing we needed to support here or do we still need support for articles?

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
franklinsch: Hmm, possibly. Out of scope of this PR though, especially as `sourceLanguage` is public API. Thanks for bringing it up!

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
franklinsch: We are at an earlier stage, in `registerDocuments`. What we're doing here is updating the reference now that we have enough information to know in which languages the page is available in. I'll add a comment to clarify that.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
franklinsch: Makes sense, I see it's hard to follow in the GitHub diff as well. I'll take your suggestions!

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
franklinsch: In this case, it's actually a programmer error if `context.entity(with:)` throws, so the force-unwrap makes sense. It could be improved with a `preconditionFailure` with a more user-facing message, though.

Sources/SwiftDocC/Infrastructure/Topic Graph/TopicGraph.swift
franklinsch: I think since `Node` is a reference type, it makes sense to avoid accidental mutation. I added an API to create a new node given a new reference, and an API in `TopicGraph` to update a node's reference.

franklinsch:
> Was tutorials the last thing we needed to support here or do we still need support for articles?

I believe tutorials was the last thing we needed to make multi-language. Article support was added by #81.

franklinsch:
@swift-ci please test

franklinsch:
@swift-ci please test Linux platform

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
ethan-kusters: Deal. 🤝😃

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -621,15 +624,30 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
...
+                    sourceLanguage: Self.defaultLanguage(in: sourceLanguages),
+                    availableSourceLanguages: sourceLanguages,
+                    name: .conceptual(title: technology.intro.title),
+                    markup: technology.originalMarkup,
+                    semantic: technology
+                )
+                documentationCache[technologyReference] = technologyNode
+                
+                // Update the reference in the topic graph with the technology's available languages.
+                topicGraph.updateReference(
```
ethan-kusters, line 641: Nice! This looks great- much easier for me to follow.

ethan-kusters (approved):

PR #97, opened 2022-03-05
Deprecated symbols sidenav

daniel-grumberg:
Bug/issue #, if applicable: rdar://89564513

Summary

Introduce/change 3 boolean node properties in RenderIndex JSON:
- Change `isExternal` to `external`
- Add `deprecated`. Data for this is generated for render index nodes associated with a symbol. The node is marked as deprecated if we have an explicit deprecated platform version for the associated symbol
- Add `beta` to the RenderIndex JSON OpenAPI spec but doesn't actually generate anything for it.

Dependencies

None

Testing

Build documentation for a project with an explicitly deprecated symbol, and check that the RenderIndex JSON contains a `"deprecated" : true` property for the relevant node.

Checklist

[X]  Added tests
[X]  Ran the ./bin/test script and it succeeded
[X]  Updated documentation if necessary

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -504,13 +504,24 @@ extension NavigatorIndex {
         /// Maps an arbitrary Platform name string to a Platform.Name instance.
         private var nameToPlatform = [String: Platform.Name]()
         
+        // availabilityIDs and availabilityToID serve as a bidirectional map to lookup entries in the
+        // availabilityIndex of the NavigatorIndex. NavigatorItem stores an availabilityID that corresponds
+        // to the key of availabilityIDs. The associated value for that availabilityID corresponds
+        // to a list of IDs which is used to get the availability information from the availability index.
+        // availabilityToID is used to reuse the same availabilityID if two NavigatorItem have the same
+        // availability information.
+        
+        private static let availabilityIDWithNoAvailabilities = 0
+        
         /// The map of the availabilities from a single availabilityID to an array of availabilities inside the availability index.
         /// This approach gives us the opportunity to map multiple availabilities with the same entries using a single ID.
         /// Ex. An item with: iOS 13.0 and macOS 10.15 can share the same ID with other items having exactly the same availability.
-        private var availabilityIDs = [Int: [Int]]()
+        /// We use the `0` value to indicate that there are no associated availabilityIndex entries.
+        private var availabilityIDs: [Int: [Int]] = [availabilityIDWithNoAvailabilities: []]
         
         /// The map of the availabilities to their ID, the opposite of `availabilityIDs`.
-        private var availabilityToID = [[Int]: Int]()
+        /// Conversely we make sure that an empty list of IDs into the availabilityIndex maps back to the `0` value.
+        private var availabilityToID: [[Int]: Int] = [[]: availabilityIDWithNoAvailabilities]
         
         /// Indicates if the path component inside the navigator item needs to be persisted or not.
         private let writePathsOnDisk: Bool
@@ -687,7 +698,7 @@ extension NavigatorIndex {
                                               languageID: language.mask,
                                               title: title,
                                               platformMask: platformID,
-                                              availabilityID: UInt64(availabilityID))
+                                              availabilityID: UInt64(Self.availabilityIDWithNoAvailabilities))
                 groupItem.path = groupIdentifier.path + "#" + fragment
 
                 let navigatorGroup = NavigatorTree.Node(item: groupItem, bundleIdentifier: bundleIdentifier)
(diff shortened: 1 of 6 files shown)


daniel-grumberg:
@swift-ci please test

daniel-grumberg:
@swift-ci please test

Sources/SwiftDocC/SwiftDocC.docc/Resources/RenderIndex.spec.json
```diff
@@ -82,10 +82,18 @@
...
                         "default": "false"
                     },
+                    "external": {
+                        "type": "boolean",
+                        "default": "false"
+                    },
+                    "beta": {
+                        "type": "boolean",
+                        "default": "false"
+                    }
```
ethan-kusters, line 96: Looks like we're missing a comma here.

```suggestion
                    },
```

We don't (yet) have CI for validating our OpenAPI schemas so I always find it helpful to manually run them through https://apitools.dev/swagger-parser/online/ whenever making a change.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -1159,6 +1160,10 @@ extension NavigatorIndex {
             return problems
         }
         
+        public func getAvailabilityEntryIDs(for availabilityID: UInt64) -> [Int]? {
```
ethan-kusters, line 1163: Maybe this should be internal for now? Otherwise, if you think it should be public, we should document it.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -507,10 +506,12 @@ extension NavigatorIndex {
         /// The map of the availabilities from a single availabilityID to an array of availabilities inside the availability index.
         /// This approach gives us the opportunity to map multiple availabilities with the same entries using a single ID.
         /// Ex. An item with: iOS 13.0 and macOS 10.15 can share the same ID with other items having exactly the same availability.
-        private var availabilityIDs = [Int: [Int]]()
+        /// We use the `0` value to indicate that there are no associated availabilityIndex entries.
```
ethan-kusters, line 509: Is there a comment (just a regular non-documentation comment) we could add above these two properties (below `nameToPlatform`) to explain more generally what we these maps are used for and what the `0` value is reserved for?

Also- if it could briefly explain the bug you encountered and how the `0` fixes it.

I think I'll be confused about this next time I see it. 😃

ethan-kusters (commented):
This looks great! Thanks for figuring out that tricky non-determinism issue.

Just left a few comments/questions but I think this is nearly ready to go.

Sources/SwiftDocC/SwiftDocC.docc/Resources/RenderIndex.spec.json
daniel-grumberg: Thanks for the heads up! Maybe we should run this as part `./bin/test`?

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
daniel-grumberg: Not sure explaining the bug here is appropriate, but I will add a description for what these two dictionaries do exactly. Essentially what was happening is that some parts of the code assumed that an availability ID of 0 meant that there is no availability information in the `availabilityIndex` for that particular `NavigatorItem`. However, this wasn't a guarantee, because 0 wasn't reserved for that. If the first `RenderNode` to be indexed had availability information the availability data for that would be stored under ID 0. So that was why we were seeing all these symbols being marked as deprecated. Essentially all symbols that didn't define their own availability would wind up having the availability of the first RenderNode to be indexed.

Tests/SwiftDocCTests/Indexing/RenderIndexTests.swift
```diff
@@ -247,8 +249,116 @@ final class RenderIndexTests: XCTestCase {
...
+              "schemaVersion": {
+                "major": 0,
+                "minor": 1,
+                "patch": 0
+              }
+            }
+            """#
+        )
+        // Let's check that the "external" key is correctly parsed into the isExternal field of RenderIndex.Node.
+        XCTAssertTrue(renderIndexWithExternalNode.interfaceLanguages["swift"]![0].isExternal)
```
franklinsch, line 274:
```suggestion
        XCTAssertTrue(try XCTUnwrap(renderIndexWithExternalNode.interfaceLanguages["swift"])[0].isExternal)
```

daniel-grumberg:
@swift-ci please test

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
ethan-kusters: Ah got it. Agreed we don't need an explanation of the bug in that case then.

Thanks for explaining here! And I think the addition of `availabilityIDWithNoAvailabilities` makes this easy to follow.

ethan-kusters (approved):
Looks great. Thank you!

daniel-grumberg:
@swift-ci please test

PR #104, opened 2022-03-14
Filter topics languages for automatically-curated See Alsos

franklinsch:
Bug/issue #, if applicable: rdar://90168171

## Summary

As a follow-up to #103, filters out topics in automatically-generated See Also sections that are not available in the parent's current language.

## Dependencies

None.

## Testing

Build documentation for a mixed-language framework and verify that automatically-generated See Also sections only contain topics that are available in the current language of the parent page.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
+++ b/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
@@ -115,7 +115,14 @@ public struct AutomaticCuration {
     ///   - bundle: A documentation bundle.
     /// - Returns: A group title and the group's references or links.
     ///   `nil` if the method can't find any relevant links to automatically generate a See Also content.
-    static func seeAlso(for node: DocumentationNode, context: DocumentationContext, bundle: DocumentationBundle, renderContext: RenderContext?, renderer: DocumentationContentRenderer) throws -> TaskGroup? {
+    static func seeAlso(
+        for node: DocumentationNode,
+        withTrait variantsTrait: DocumentationDataVariantsTrait,
+        context: DocumentationContext,
+        bundle: DocumentationBundle,
+        renderContext: RenderContext?,
+        renderer: DocumentationContentRenderer
+    ) throws -> TaskGroup? {
         // First try getting the canonical path from a render context, default to the documentation context
         guard let canonicalPath = renderContext?.store.content(for: node.reference)?.canonicalPath ?? context.pathsTo(node.reference).first,
             !canonicalPath.isEmpty else {
@@ -124,6 +131,19 @@ public struct AutomaticCuration {
         }
         
         let parentReference = canonicalPath.last!
+        
+        func filterReferences(_ references: [ResolvedTopicReference]) throws -> [ResolvedTopicReference] {
+            try references
+                // Don't include the current node.
+                .filter { $0 != node.reference }
+            
+                // Filter out nodes that aren't available in the given trait.
+                .filter { reference in
+                    try context.entity(with: reference)
+                        .availableVariantTraits
+                        .contains(variantsTrait)
+                }
+        }
         
         // Look up the render context first
         if let taskGroups = renderContext?.store.content(for: parentReference)?.taskGroups,
@@ -134,7 +154,7 @@ public struct AutomaticCuration {
         {
             // Group match in render context, verify if there are any other references besides the current one.
             guard linkingGroup.references.count > 1 else { return nil }
-            return (title: linkingGroup.title, references: linkingGroup.references.filter { $0 != node.reference })
+            return (title: linkingGroup.title, references: try filterReferences(linkingGroup.references))
         }
         
         // Get the parent's task groups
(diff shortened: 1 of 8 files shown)


franklinsch:
@swift-ci please test

franklinsch:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
```diff
@@ -125,6 +132,19 @@ public struct AutomaticCuration {
         
         let parentReference = canonicalPath.last!
         
+        func filterReferences(_ references: [ResolvedTopicReference]) throws -> [ResolvedTopicReference] {
```
ethan-kusters, line 135: Nit: Is there a more specific name we can use to make this more readable at the call-site?

```suggestion
        func referencesAvailableInCurrentTrait(in references: [ResolvedTopicReference]) throws -> [ResolvedTopicReference] {
```

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
```diff
@@ -134,7 +154,7 @@ public struct AutomaticCuration {
         {
             // Group match in render context, verify if there are any other references besides the current one.
             guard linkingGroup.references.count > 1 else { return nil }
```
ethan-kusters, line 156: Nit since this is a little out-of-scope, but the existing logic seems a little hacky to me and would break if we ever stop including the current reference in the linking group.

```suggestion
            let references = try filterReferences(linkingGroup.references)
            guard !references.isEmpty else { return nil }
```

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
```diff
@@ -125,6 +132,19 @@ public struct AutomaticCuration {
...
         let parentReference = canonicalPath.last!
         
+        func filterReferences(_ references: [ResolvedTopicReference]) throws -> [ResolvedTopicReference] {
+            try references
+                // Don't include the current node.
+                .filter { $0 != node.reference }
+            
+                // Filter out nodes that aren't available in the given trait.
+                .filter { reference in
+                    try context.entity(with: reference)
```
ethan-kusters, line 142: Should we adopting this logic in the `topics(for:withTrait:context:)` function above? It seems cleaner than only relying on the kind being available for a given trait but maybe I'm missing something. (I'm not entirely sure what `availableVariantTraits` is pulled from.)

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
franklinsch: It also filters out the current reference from the task group, though. We could also consider making this function only handle the trait-based filtering, but the logic to filter out the current reference would then be duplicated.

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
franklinsch: The automatic curation logic here finds a task group that contains this reference, removing this reference from the task group. The `count > 1` is essentially to not emit the task group at all if the only reference it contains is this reference.

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
franklinsch: We could, but we need to get the appropriate language-specific kind of the symbol anyway in that case. If we ported this logic to `topics(for:withTrait:context:)`, we'd check the languages in `DocumentationNode. availableVariantTraits` but also need to verify that the symbol has `kindVariants` for each of these languages which seems redundant. Compilers are expected to produce the `kind` field for their symbols, so DocC shouldn't be in a case where a symbol is available in multiple languages, but the symbol only has 1 `kind` value.

`DocumentationNode.availableVariantTraits` is populated from this logic: https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Infrastructure/Symbol%20Graph/SymbolReference.swift#L174

franklinsch:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
ethan-kusters: Right I see that- I'm just suggesting we do the filtering ahead of time so we can gate on the already filtered group being empty or not.

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
ethan-kusters: > Compilers are expected to produce the kind field for their symbols, so DocC shouldn't be in a case where a symbol is available in multiple languages, but the symbol only has 1 kind value.

What if we have the same kind in multiple languages?

ethan-kusters (approved):

franklinsch:
@swift-ci please test

PR #106, opened 2022-03-16
Adopt the Swift-DocC SwiftPM plugin

ethan-kusters:
## Summary

Simplify the `bin/preview-docs` script by adopting the Swift-DocC SwiftPM plugin.

## Testing

Confirm that the documentation generatation and previewing works as expected from the command-line when using Swift 5.6 and that the package continues to build when using Swift 5.5.

## Checklist

- [ ] ~~Added tests~~ N/A


ethan-kusters:
@swift-ci please test

bin/preview-docs
```diff
@@ -48,78 +47,6 @@ while test $# -gt 0; do
   esac
 done
 
-# Create the output directory for the symbol graphs if needed.
```
franklinsch, line 51: I guess this means contributors will need Swift 5.6 to be able to build and preview documentation for Swift-DocC? Should we keep the previous way of building docs around for contributors that are still developing on 5.5 for a small period of time?

daniel-grumberg: I agree with this.

ethan-kusters: I agree it makes sense to move incrementally with this. Given that, I also updated our README to suggest people use Swift-DocC-Render-Artifact instead of building Swift-DocC-Render locally.

[efffe2a](https://github.com/apple/swift-docc/pull/106/commits/efffe2a0a268dcab69c6f15ea8baaf7ae1449417)

ethan-kusters:
@swift-ci please test

franklinsch (approved):

Kyle-Ye (approved):
When talking about downward compatibility (such as swift 5.5 and swift 5.6 here), we usually have two ways. I'd like to know what is the best practice in swift-docc related repos?
1. Using `Package.swift`(with swift-tools-version:5.5)  and `Package@swift-5.6.swift` like https://github.com/apple/swift-docc-symbolkit
2. doing some conditional check like `#if swift(>=5.6)`

franklinsch:
Personally I think the conditional check approach is cleaner when possible because it allows you to not duplicate the Package manifest, however there are cases where it's not possible to use a conditional check (e.g., in the SymbolKit example where what we wanted to increment was the tools version).

PR #132, opened 2022-04-08
use `default` branches for unknown symbol kinds

QuietMisdreavus:
## Summary

This PR tweaks how swift-docc uses `KindIdentifier` instances from SymbolKit. Instead of matching on an explicit `.unknown` case (and using an exhaustive match), it instead changes them to `default` branches, to catch any new cases as they're added. This allows us to add new symbol kinds to SymbolKit (e.g. https://github.com/apple/swift-docc-symbolkit/pull/21) without creating a source-breaking change.

## Dependencies

N/A

## Testing

Existing functionality should remain unchanged.

## Checklist

- [ n/a ] Added tests


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
I confirmed in a local build that this change still works with https://github.com/apple/swift-docc-symbolkit/pull/7 applied, which would have otherwise been a breaking change.

ethan-kusters (approved):

QuietMisdreavus:
@swift-ci Please test

PR #266, opened 2022-05-26
[5.7] Add the ability to add arbitrary attributes to a NavigatorIndex

daniel-grumberg:
Cherry-pick of https://github.com/apple/swift-docc/pull/258

This stores a handle to the attributes in the index, the tree, and the
individual nodes. Attributes are stored in a `[String: Any]` and the client is
expected to provided extensions with computed properties as needed.

Additionaly this commits refactors the API for reading a NavigatorIndex from an
on-disk LMDB database. The prveious API used a designated initializer to do
this. However, this initializer was complex as it opened database connections
etc...  Using a static function enables clients to be more resilient to API
changes, by letting them define protocols that specify the new API and providing
a no-op default implementation that delegates to old API. This is not currently
possible with the initializer base API as the initializer would need to be
marked required to ensure subclasses of NavigatorIndex implement this. This will
be made possible in the future as clients transition to the static function
based API.

rdar://93503303

- [X] Added tests

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -136,48 +136,134 @@ public class NavigatorIndex {
     }()
     
     /**
-     Initialize an `NavigatorIndex` from a given path.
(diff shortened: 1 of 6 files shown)


daniel-grumberg:
@swift-ci please test

PR #277, opened 2022-06-04
Fix error when writing multiple index.html files to the same location

ethan-kusters:
Bug/issue #, if applicable: rdar://94311195

## Summary

Addresses a regression introduced by #166 where `docc convert` would throw an error when writing multiple `index.html` files to the same location on disk.

This can occur if docc is writing to an output directory that already contains built documentation or when docc is given bad input such that multiple documentation pages have the same path on the filesystem.

## Details

There's some additional changes here ([ecc2774](https://github.com/apple/swift-docc/pull/277/commits/ecc2774c9f13526db2c315db2613244d7812061c)) to fix an issue with the diagnostic that was thrown when we found this issue:

```
DocC.symbols.json: error: “index.html” couldn’t be copied to “docc” because an item with the same name already exists.
```

The diagnostic was incorrectly associated with an input file because of the way DocumentationConverter currently records errors encountered when encoding RenderNodes. This made it a little more confusing to track down this issue.

## Testing

Build documentation for a small DocC catalog that includes multiple symbol graph files- one with a `"docc"` module name, and one with a `"DocC"` module name. Confirm that `docc convert` does not throw an error. (This is what the added test case does.)

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationConverter.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationConverter.swift
@@ -311,8 +311,7 @@ public struct DocumentationConverter: DocumentationConverterProtocol {
                         }
                     }
                 } catch {
-                    results.append(.init(description: error.localizedDescription, source: source))
-                    diagnosticEngine.emit(.init(description: error.localizedDescription, source: source))
+                    recordProblem(from: error, in: &results, withIdentifier: "render-node")
                 }
             }
         }
@@ -327,8 +326,7 @@ public struct DocumentationConverter: DocumentationConverterProtocol {
         do {
             try outputConsumer.consume(assetsInBundle: bundle)
         } catch {
-            conversionProblems.append(.init(description: error.localizedDescription, source: nil))
-            diagnosticEngine.emit(.init(description: error.localizedDescription, source: nil))
+            recordProblem(from: error, in: &conversionProblems, withIdentifier: "assets")
         }
         
         // Write various metadata
@@ -338,8 +336,7 @@ public struct DocumentationConverter: DocumentationConverterProtocol {
                 try outputConsumer.consume(indexingRecords: indexingRecords)
                 try outputConsumer.consume(assets: assets)
             } catch {
-                conversionProblems.append(.init(description: error.localizedDescription, source: nil))
-                diagnosticEngine.emit(.init(description: error.localizedDescription, source: nil))
+                recordProblem(from: error, in: &conversionProblems, withIdentifier: "metadata")
             }
         }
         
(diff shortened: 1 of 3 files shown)


Sources/SwiftDocCUtilities/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
```diff
@@ -116,6 +116,16 @@ class JSONEncodingRenderNodeWriter {
             attributes: nil
         )
         
-        try fileManager.copyItem(at: indexHTML, to: htmlTargetFileURL)
+        do {
+            try fileManager.copyItem(at: indexHTML, to: htmlTargetFileURL)
+        } catch let error as NSError where error.code == NSFileWriteFileExistsError {
```
ethan-kusters, line 121: This is the core fix for the actual issue here. `copyItem` doesn't have a way to specify that we're okay with overwriting the destination file so instead we just catch that specific error and try again.

Sources/SwiftDocC/Infrastructure/DocumentationConverter.swift
```diff
@@ -412,6 +406,34 @@ public struct DocumentationConverter: DocumentationConverterProtocol {
...
+    /// problem array.
+    ///
+    /// - Parameters:
+    ///   - error: The error that describes the problem.
+    ///   - problems: The array that the created problem should be appended to.
+    ///   - identifier: A unique identifier the problem.
+    private func recordProblem(
+        from error: Swift.Error,
+        in problems: inout [Problem],
+        withIdentifier identifier: String
```
ethan-kusters, line 422: This change just simplified the way we report errors in DocumentationConverter to make it less likely we'll introduce the problem of accidentally including a `source` location in the future.

It also updates the `identifier` used to actually be unique for each item. Currently, we're using the `org.swift.docc.DocumentationConverter` identifier for all of them.

Sources/SwiftDocC/Infrastructure/DocumentationConverter.swift
```diff
@@ -311,8 +311,7 @@ public struct DocumentationConverter: DocumentationConverterProtocol {
                         }
                     }
                 } catch {
-                    results.append(.init(description: error.localizedDescription, source: source))
```
ethan-kusters, line 314: Here's where we were accidentally including the `source` location.

ethan-kusters:
@swift-ci please test

franklinsch (approved):

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

PR #283, opened 2022-06-08
Don’t filter topic group items from unavailable languages

ethan-kusters:
Resolves rdar://94406023.

Bug/issue #, if applicable: 

## Summary

Addresses a regression where DocC began excluding manually curated topic group items that link to languages not otherwise locally included in the documentation content.

This is fallout from the work done to support multi-language variants of a page where we introduced filtering of topic group items based on the current variant.

So a catalog that only includes Swift content might link to a symbol that externally resolves to a Data or Objective-C source language. Then that symbol would be filtered out since the current page only includes a Swift-language variant.

The solution is to only filter topic group items based on languages that the parent page is actually available in. This ensures that we never filter out an item in a way that makes it impossible to actually view.

In other words, it’s okay to filter out Objective-C  items from the Swift variant of a Swift/Objective-C page. But it’s not okay to filter out Javascript items from this page since there is no Javascript version of that page for the reader to select in order to view that item.

## Testing

Build a Swift-only DocC catalog that curates externally resolved symbols to non-Swift content and ensure those items appear in the rendered documentation.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -633,6 +633,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                         topics,
                         allowExternalLinks: false,
                         allowedTraits: [trait],
+                        availableTraits: documentationNode.availableVariantTraits,
                         contentCompiler: &contentCompiler
                     )
                 )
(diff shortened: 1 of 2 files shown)


ethan-kusters:
@swift-ci please test

franklinsch:
Thanks! There are still some navigator index changes required here to support displaying Objective-C topics in the Swift navigator, in cases where you curate an Objective-C–only topic in a Swift-only topic:

Screen Shot 2022-06-08 at 10 17 28

I'll integrate your changes into #282 which includes the navigator index support. These changes are super helpful, thanks!

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

PR #287, opened 2022-06-09
Fix `ResolvedTopicReference` memory leak when not converting

ethan-kusters:
Bug/issue #, if applicable: rdar://92131596

## Summary

The caching mechanism in `ResolvedTopicReference` never releases references to the `ResolvedTopicReference`s it holds.

This has likely always been a problem since the introduction of the cache, but is just now appearing as a memory leak when profiling SwiftDocC because we recently added copy-on-write functionality to `ResolvedTopicReference`.

The current solution proposed here is to selectively enable ResolvedTopicReference caching for specific bundle IDs _only_ when running a conversion. This disables the caching by default when (for example) just using SwiftDocC as a library to load RenderNode’s in memory.

There’s more work to do here- but this resolves the immediate issue of caching (and never releasing) every reference when using SwiftDocC as a library and not to actively convert documentation.

## Testing

Use SwiftDocC to load and unload a large RenderNode in-memory and confirm that no `ResolvedTopicReference`s are held when profiling the process.

Confirm `docc convert` doesn't regress in performance when generating documentation for a large bundle.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Identifier.swift
+++ b/Sources/SwiftDocC/Model/Identifier.swift
@@ -96,6 +96,15 @@ public struct ResolvedTopicReference: Hashable, Codable, Equatable, CustomString
     static func purgePool(for bundleIdentifier: String) {
         sharedPool.sync { $0.removeValue(forKey: bundleIdentifier) }
     }
+    
+    /// Enables reference caching for any identifiers created with the given bundle identifier.
+    static func enableReferenceCaching(for bundleIdentifier: ReferenceBundleIdentifier) {
+        sharedPool.sync { sharedPool in
+            if !sharedPool.keys.contains(bundleIdentifier) {
+                sharedPool[bundleIdentifier] = [:]
+            }
+        }
+    }
 
     /// The URL scheme for `doc://` links.
     public static let urlScheme = "doc"
@@ -174,7 +183,10 @@ public struct ResolvedTopicReference: Hashable, Codable, Equatable, CustomString
         )
 
         // Cache the reference
-        Self.sharedPool.sync { $0[bundleIdentifier, default: [:]][key] = self }
+        Self.sharedPool.sync { sharedPool in
+            // If we have a shared pool for this bundle identifier, cache the reference
+            sharedPool[bundleIdentifier]?[key] = self
+        }
     }
     
     private static func cacheKey(
--- a/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
+++ b/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
@@ -65,10 +65,10 @@ class NavigatorIndexingTests: XCTestCase {
         return root
     }
     
-    func generateSmallTree() -> Node {
+    func generateSmallTree(bundleIdentifier: String = testBundleIdentifier) -> Node {
         var index = 1
         let rootItem = NavigatorItem(pageType: 1, languageID: Language.all.rawValue, title: "Root", platformMask: 1, availabilityID: 1)
-        let root = Node(item: rootItem, bundleIdentifier: "org.swift.docc.example")
+        let root = Node(item: rootItem, bundleIdentifier: bundleIdentifier)
         
         @discardableResult func addItems(n: Int, items: [Node], language: Language) -> [Node] {
             var leaves = [Node]()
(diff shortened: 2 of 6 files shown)


Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
```diff
@@ -243,6 +243,28 @@ Root
         XCTAssertTrue(validateTree(node: treeWithAttributesNonAtomic.root, validator: attributesValidator), "The tree lacks the correct attributes.")
     }
     
+    func testLoadingNavigatorIndexDoesNotCacheReferences() throws {
```
ethan-kusters, line 246: This actually passes even without the change in this PR (I think we recently removed any usage of `ResolvedTopicReference` in the navigator index in #166) but I think this is still a useful test to ensure we don't regress.

ethan-kusters:
Performance test results look good:

```
TestFramework-5
| Duration for 'bundle-registration' (sec) | +0.46%     | 2.61                 | 2.62                 |
| Duration for 'convert-total-time' (sec)  | -0.79%     | 4.07                 | 4.04                 |
| Duration for 'documentation-processing'  | -3.50%     | 1.28                 | 1.24                 |
| Duration for 'finalize-navigation-index' | +1.33%     | 0.07                 | 0.08                 |
| Peak memory footprint (MB)               | +3.78%     | 181.53               | 188.39               |
| Data subdirectory size (MB)              | no change  | 57.35                | 57.35                |
| Index subdirectory size (MB)             | no change  | 0.47                 | 0.47                 |
| Total DocC archive size (MB)             | no change  | 74.44                | 74.44                |
TestFramework-10
| Duration for 'bundle-registration' (sec) | +0.08%     | 5.25                 | 5.25                 |
| Duration for 'convert-total-time' (sec)  | +2.88%     | 8.15                 | 8.39                 |
| Duration for 'documentation-processing'  | +8.74%     | 2.59                 | 2.81                 |
| Duration for 'finalize-navigation-index' | +2.05%     | 0.15                 | 0.15                 |
| Peak memory footprint (MB)               | +1.31%     | 340.56               | 345.04               |
| Data subdirectory size (MB)              | no change  | 114.68               | 114.68               |
| Index subdirectory size (MB)             | no change  | 0.95                 | 0.95                 |
| Total DocC archive size (MB)             | no change  | 148.05               | 148.05               |
TestFramework-25
| Duration for 'bundle-registration' (sec) | -0.23%     | 13.32                | 13.29                |
| Duration for 'convert-total-time' (sec)  | +0.26%     | 21.68                | 21.73                |
| Duration for 'documentation-processing'  | +0.36%     | 7.56                 | 7.59                 |
| Duration for 'finalize-navigation-index' | +1.10%     | 0.36                 | 0.37                 |
| Peak memory footprint (MB)               | -1.46%     | 836.92               | 824.68               |
| Data subdirectory size (MB)              | no change  | 287.73               | 287.73               |
| Index subdirectory size (MB)             | no change  | 2.41                 | 2.41                 |
| Total DocC archive size (MB)             | no change  | 369.96               | 369.96               |
| Topic Anchor Checksum                    | no change  | 9a675b9ad6d69f8b7f0c | 9a675b9ad6d69f8b7f0c |
| Topic Graph Checksum                     | no change  | 665713509084b5131a37 | 665713509084b5131a37 |
TestFramework-50
| Duration for 'bundle-registration' (sec) | +0.13%     | 26.90                | 26.93                |
| Duration for 'convert-total-time' (sec)  | no change  | 46.95                | 46.95                |
| Duration for 'documentation-processing'  | -0.15%     | 18.38                | 18.35                |
| Duration for 'finalize-navigation-index' | +0.27%     | 0.75                 | 0.75                 |
| Peak memory footprint (MB)               | +0.77%     | 1572.60              | 1584.63              |
| Data subdirectory size (MB)              | no change  | 578.41               | 578.41               |
| Index subdirectory size (MB)             | no change  | 4.84                 | 4.84                 |
| Total DocC archive size (MB)             | no change  | 742.07               | 742.07               |
```

ethan-kusters:
@swift-ci please test

Sources/SwiftDocC/Model/Identifier.swift
```diff
@@ -96,6 +96,15 @@ public struct ResolvedTopicReference: Hashable, Codable, Equatable, CustomString
     static func purgePool(for bundleIdentifier: String) {
         sharedPool.sync { $0.removeValue(forKey: bundleIdentifier) }
     }
+    
+    /// Enables reference caching for any identifiers created with the given bundle identifier.
+    static func enableReferenceCaching(for bundleIdentifier: ReferenceBundleIdentifier) {
+        sharedPool.sync { sharedPool in
+            if !sharedPool.keys.contains(bundleIdentifier) {
+                sharedPool[bundleIdentifier] = [:]
```
franklinsch, line 104: Minor thing here for legibility, which could be left as a future improvement: since the behavior of cache is getting more complicated, it would be nice to refactor into its own type that exposes clearer operations like `enableCaching(forBundleIdentifier:)`. It's not immediately obvious what `sharedPool[bundleIdentifier] = [:]` does, and it looks like we're actually clearing a dictionary that potentially had values before.

franklinsch (approved):
Thanks! This seems fine as an incremental improvement. Longer term, I'd say the cache should be owned by a non-static value that is cleared between documentation builds (maybe a value owned by `DocumentationContext`?). And the creation of `ResolvedTopicReference`s would be done through that value, returning cached instances.

Sources/SwiftDocC/Model/Identifier.swift
ethan-kusters: > Minor thing here for legibility, which could be left as a future improvement: since the behavior of cache is getting more complicated, it would be nice to refactor into its own type that exposes clearer operations like enableCaching(forBundleIdentifier:)

I agree that would be good but I think we should leave that for the more comprehensive change we'll do here in the future (making the cache non-global). My general goal here was to make the change minimal and targeted to reduce risk for landing this on 5.7.

> it looks like we're actually clearing a dictionary that potentially had values before.

The if statement should prevent this, right? That was the intention.

franklinsch: Ah right yes, I meant more that the reason we're initializing an empty dictionary here is so that 188 is not a no-op. Generally, I'd say relying on the presence of a key being associated with an empty value is brittle, which is why we could formalise it in a wrapper time. But definitely for a future PR :)

ethan-kusters: Yeah agreed with all of that 👍🏻

PR #288, opened 2022-06-10
Remove spaces from test bundle identifiers

franklinsch:
Bug/issue #, if applicable: rdar://94811638

## Summary

Removes spaces from the bundle identifiers using in Swift-DocC tests to work around a bug in macOS Ventura Beta 1 (94747378).

## Dependencies

None.

## Testing

N/A, not a user facing change.

## Checklist

- [x] Added tests


franklinsch:
@swift-ci please test

ethan-kusters:
Should we add an XCTSkip'd test somewhere that is testing this? So we know if this regresses again in the future? (Unclear to me if having spaces in a bundle ID is supported or not)

franklinsch:
Given the URL RFC, I think we should actually treat bundle identifiers with spaces as invalid. The bug here is that DocC currently accepts them but shouldn't. I filed https://github.com/apple/swift-docc/issues/290

ethan-kusters (approved):

franklinsch:
@swift-ci please test

PR #304, opened 2022-06-15
Emit language-specific relationships

franklinsch:
Bug/issue #, if applicable: rdar://91627374

## Summary

Use SymbolKit's new `relationshipsByLanguage` API to emit language-specific relationships, e.g., for symbols that have different conformances in Swift vs. Objective-C.

## Dependencies

https://github.com/apple/swift-docc-symbolkit/pull/28

## Testing

Build documentation for the following class:

```
@interface MyClass : NSObject
@end
```

Verify that on the Swift documentation page, the relationships listed include the ones synthesized by the Swift compiler, e.g., a conformance to `Swift.Equatable`, and that on the Objective-C page, only `NSObject` is listed as an Inherited From symbol.

## Checklist

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

## TODO

- [x] Add tests for language-specific `defaultImplementationOf`, `requirementOf`, and `optionalRequirementOf` relationships

PR #325, opened 2022-06-29
Stop duplicating auto-generated collections per-platform

ethan-kusters:
Bug/issue #, if applicable: rdar://95882462

## Summary

Resolves a regression introduced by #304 where auto-generated "Protocol Implementations" collections were being duplicated per platform.

The immediate fix is to call the `createInheritedSymbolsAPICollection` method a single time with all of the relevant relationships, instead of once per platform. It tracks internal state while building up the API collections so it's important to provide all possible relationships when calling the method.

I think there's a likely a longer-term fix we should make here in the way we're handling multi-platform symbol graph information since currently we're handling it similarly to multi-language symbol graph information but don't actually emit variants for it.

## Testing

Build documentation for multiple platforms, save the symbol graphs, and pass them to `docc convert`. Confirm that any auto-generated "Protocol Implementations" sections are only created once. 

Attaching a [small DocC catalog](https://github.com/apple/swift-docc/files/9014799/r95882462-reproducer.docc.zip) that reproduces the issue created by following these instructions for the following Swift snippet:

```swift
public struct BestStruct: MyProtocol {}

public protocol MyProtocol {}

extension MyProtocol {
    public func bestFunc() {}
}
```

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -1261,13 +1261,11 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
             }
             
             // Create inherited API collections
-            for (_, relationships) in combinedRelationships {
-                try GeneratedDocumentationTopics.createInheritedSymbolsAPICollections(
-                    relationships: relationships,
-                    context: self,
-                    bundle: bundle
-                )
-            }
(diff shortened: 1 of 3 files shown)


ethan-kusters:
@swift-ci please test

QuietMisdreavus (approved):
Nice catch, thanks!

PR #332, opened 2022-07-07
load cross-import overlay module data, even for extensions

QuietMisdreavus:
Bug/issue #, if applicable: rdar://94736726

## Summary

Swift symbol graphs format the file names of cross-import overlays as if they were extension symbol graphs: `_OverlayModule@DeclaringModule.symbols.json`. However, if that overlay *also* declares some extensions, an extension symbol graph is also emitted: `_OverlayModule@SomeOtherModule.symbols.json`. In both cases, SymbolKit and Swift-DocC overwrite the symbol graph's module name with the name after the `@` in the file name. This creates a situation for cross-import overlays which declare extensions where the rendered module name is incorrect.

This PR changes how module information is loaded to preserve the parsed module name in the symbol graph if it indicates that it is from a cross-import overlay (i.e. it declares a "bystander" module). It also uses this information to emit the parsed module name alongside the bystander module(s) in the Render JSON for these symbols.

## Dependencies

https://github.com/apple/swift-docc-symbolkit/pull/46

## Testing

Generate a doc archive for a module which contains extensions from a cross-import overlay. (I couldn't get Xcode to generate symbol graphs that properly used cross-import overlay information, so you may need to use the symbol graphs i added as test data in this PR.) Make sure that symbols from the overlay properly have both the declaring and bystander modules in the Render JSON and/or on the page, instead of just the module being extended (which may be the same as the bystander module, as in the test data).

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
+++ b/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
@@ -139,7 +139,7 @@ struct SymbolGraphLoader {
         var symbolGraph = symbolGraphs[url]!
         let (moduleName, isMainSymbolGraph) = Self.moduleNameFor(symbolGraph, at: url)
         
-        if !isMainSymbolGraph {
+        if !isMainSymbolGraph && symbolGraph.module.bystanders == nil {
             // If this is an extending another module, change the module name to match the exteneded module.
             // This makes the symbols in this graph have a path that starts with the extended module's name.
             symbolGraph.module.name = moduleName
@@ -250,7 +250,7 @@ struct SymbolGraphLoader {
         let isMainSymbolGraph = !url.lastPathComponent.contains("@")
         
         let moduleName: String
-        if isMainSymbolGraph {
+        if isMainSymbolGraph || symbolGraph.module.bystanders != nil {
             // For main symbol graphs, get the module name from the symbol graph's data
             moduleName = symbolGraph.module.name
         } else {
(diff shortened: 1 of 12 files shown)


Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -115,7 +115,7 @@ public final class Symbol: Semantic, Abstracted, Redirected, AutomaticTaskGroups
     internal(set) public var extendedModule: String?
     
     /// Optional cross-import module names of the symbol.
-    internal(set) public var bystanderModuleNames: [String]?
```
franklinsch, line 118: Since this is public API, can we deprecated this instead, and make it return `crossImportOverlayModule.bystanderModules`?

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -139,7 +139,7 @@ struct SymbolGraphLoader {
         var symbolGraph = symbolGraphs[url]!
         let (moduleName, isMainSymbolGraph) = Self.moduleNameFor(symbolGraph, at: url)
         
-        if !isMainSymbolGraph {
+        if !isMainSymbolGraph && symbolGraph.module.bystanders == nil {
```
franklinsch, line 142: This line and 253 could benefit from comments to explain why we're checking the bystander modules. The logic of what name we use is pretty subtle.

QuietMisdreavus:
Pushed up the requested changes.

@swift-ci Please test

franklinsch (approved):

QuietMisdreavus:
Rebased onto the latest `main`.

@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test

PR #338, opened 2022-07-19
SE-0356: Implement snippet slicing

bitjammer:
Add an optional `slice` argument to the `@Snippet` directive, which
will only render a code listing comprised of the line range of the slice.

Remove snippet groups, as they are no longer used in the current design.

Replace logic that filters snippets and fake modules that hold snippets
to a more generic "is virtual" logic. A virtual symbol, topic, or documentation
node should not have its own page of documentation.

Testing: Moved snippet tests to their own test bundle as it was causing a lot
of thrash with navigator tests.

rdar://95220570

Requires SymbolKit model changes:
- https://github.com/apple/swift-docc-symbolkit/pull/50

Related change:
- https://github.com/apple/swift-docc-plugin/pull/20


bitjammer:
Looks like the CI can't fold in multiple PRs for this project, so the SymbolKit changes will need to land first.

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

Sources/SwiftDocC/Semantics/Snippets/Snippet.swift
```diff
@@ -18,18 +18,25 @@ public final class Snippet: Semantic, DirectiveConvertible {
...
             static let argumentName = "path"
         }
+        enum Slice: DirectiveArgument {
+            static let argumentName = "slice"
+        }
     }
 
     public let originalMarkup: BlockDirective
     public let path: String
+    public let slice: String?
```
franklinsch, line 28: Not specific to this PR, but having documentation comments for these public APIs would be great.

Tests/SwiftDocCTests/Semantics/SnippetTests.swift
```diff
@@ -58,10 +58,19 @@ class SnippetTests: XCTestCase {
         XCTAssertNotNil(snippet)
         XCTAssertTrue(problems.isEmpty)
     }
-
-    func testNoTopLevelPageForSnippetModule() throws {
-        let (_, context) = try testBundleAndContext(named: "TestBundle")
-        XCTAssertEqual(3, context.rootModules.count)
-        XCTAssertFalse(context.rootModules.contains { $0.path == "/documentation/Test" })
+    
+    func testSliceResolves() throws {
```
franklinsch, line 62: Are we missing a test that verifies a diagnostic is emitted if you're referring to a snippet that doesn't exist?

Sources/SwiftDocC/Semantics/MarkupReferenceResolver.swift
```diff
@@ -112,24 +118,19 @@ struct MarkupReferenceResolver: MarkupRewriter {
         return link
     }
 
-    mutating func resolveAbsoluteSymbolLink(unresolvedDestination: String, elementRange range: SourceRange?) -> String {
+    mutating func resolveAbsoluteSymbolLink(unresolvedDestination: String, elementRange range: SourceRange?) -> ResolvedTopicReference? {
         if let cached = context.documentationCacheBasedLinkResolver.referenceFor(absoluteSymbolPath: unresolvedDestination, parent: rootReference) {
             guard context.topicGraph.isLinkable(cached) == true else {
                 problems.append(disabledLinkDestinationProblem(reference: cached, source: source, range: range, severity: .warning))
-                return unresolvedDestination
+                return nil
```
franklinsch, line 125: If I understand correctly, this change doesn't change the behavior, because the previous was essentially setting `symbolLink.destination` to the same value it contained before?

Sources/SwiftDocCUtilities/Action/Actions/PreviewAction.swift
```diff
@@ -275,6 +275,8 @@ extension DocumentationContext {
     func previewPaths() throws -> [String] {
         let urlGenerator = PresentationURLGenerator(context: self, baseURL: URL(string: "/")!)
         
+        let rootModules = try rootModules.filter { try !entity(with: $0).isVirtual }
```
franklinsch, line 278: I think there might be other places in `DocumentationContext` where we want to apply the filtering. For example, I don't believe that `soleRootModuleReference` should consider virtual modules. It might be useful to expose a property like `nonVirtualRootModules` (is there a better word for non-virtual?) that we can use in more places.

Tests/SwiftDocCTests/Test Bundles/Snippets.docc/Test-snippets.symbols.json
```diff
@@ -0,0 +1,65 @@
...
+      "isVirtual": true,
+      "docComment": {
+        "lines": [
+          {
+            "text": "Does a foo. "
+          }
+        ]
+      },
+      "snippet": {
+	"language": "swift",
```
franklinsch, line 29: nit: Looks like the formatting is slightly off here (uses tabs instead of spaces like the rest of the file)

ktoso:
Just popping in to share how happy I am to see this happening already :) Even if it just lands in the packages we could be using it already in server projects -- really looking forward to it, thanks @bitjammer !

Sources/SwiftDocCUtilities/Action/Actions/PreviewAction.swift
bitjammer: I updated the `soleRootModuleReference` for the time being.

Tests/SwiftDocCTests/Semantics/SnippetTests.swift
bitjammer: Added a test.

Tests/SwiftDocCTests/Test Bundles/Snippets.docc/Test-snippets.symbols.json
bitjammer: Ah, yep! New machine, new vimrc and all that. Thanks

Sources/SwiftDocC/Semantics/MarkupReferenceResolver.swift
bitjammer: Correct, no functional change to the resolution logic. Just makes it easier to use the resolved reference in the snippet case.

Sources/SwiftDocC/Semantics/Snippets/Snippet.swift
bitjammer: No problem, added.

Sources/SwiftDocCUtilities/Action/Actions/PreviewAction.swift
bitjammer: I take that back, filtering in `soleRootModuleReference` would break snippet-only packages, so it should stay there for now.

bitjammer:
A pleasure, @ktoso, looking forward to seeing what you do with it!

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

bitjammer:
@swift-ci Please test

Sources/SwiftDocCUtilities/Action/Actions/PreviewAction.swift
franklinsch: Not updating `soleRootModuleReference` would cause wrong behavior for multi-language projects. For example, when building documentation for a Swift + Objective-C framework, the languages of article documents are inferred from `soleRootModuleReference`. When building a Swift + Objective-C framework that has snippets, there would actually be multiple modules and hence `soleRootModuleReference` would be nil and wouldn't be used to infer the languages of articles. 

This can probably be reproduced by adding snippets to `Tests/SwiftDocCTests/Test Bundles/MixedLanguageFramework.docc`, opening `/documentation/mixedlanguageframework/article` in the browser, and seeing that the language toggle at the top-right of the page would only show "Swift" rather than "Swift" and "Objective-C".

bitjammer: I see, so we need different logic here. For now, I can address this by filtering out virtual modules, but allow a virtual module to be the sole root module reference if it is the only one.

It might be good to rethink or at least rename this property in the future.

bitjammer:
@swift-ci Please test

ktoso:
> A pleasure, @ktoso, looking forward to seeing what you do with it!

Basically I want to throw out our workaround that collects all inline code pieces from a `md` file and makes a compile test for it.

We really need to have our docs code be compiled since we often end up with some old or partial code snippet and then users are utterly confused: like even yesterday there was such user:

> Don't get this example, .terminated is internal, you can't call it, or it's just an example what will happen when you creating a cluster system?

While the actual issue was... well, this method does not exist... and since the docc code is not verified, we missed it.

Really looking forward to this :) Thanks!

franklinsch (approved):

bitjammer:
Removed temporary commit to point to in-flight SymbolKit branch.

bitjammer:
@swift-ci Please test

PR #348, opened 2022-07-28
Trim snippet slice indentation

bitjammer:
When trimming a slice out of a snippet, the slice may span some
deeply nested scope that has indentation, so we don't want to take
those lines as is, because they may be unnecessarily indented when
presented as a slice.

rdar://97696391


bitjammer:
@swift-ci Please test

bitjammer:
Copyright header. Sigh.

bitjammer:
@swift-ci Please test

ethan-kusters (approved):

PR #349, opened 2022-07-28
Implement Snippet Slicing

bitjammer:
**Explanation**: This PR contains two commits to add a new `slice` parameter to the `@Snippet` directive, allowing authors to inline a named region of a snippet into documentation instead of the whole snippet if they like.

**Scope**: This implements changes made to the SE-0356 Swift Snippets proposal which added snippet slicing. These changes are only realized in the DocC Plugin and not Xcode.

**Risk**: Low. Snippets are already an opt-in feature that should only affect behavior when using DocC Plugin.

**Testing**: 
- [x] Included unit tests
- [x] Manual integration tests in Xcode by running the *Build Documentation* command on several projects to check for regressions
  - DocC itself (Swift)
  - https://developer.apple.com/documentation/xcode/slothcreator_building_docc_documentation_in_xcode (Swift)
  - https://github.com/mdiep/MMMarkdown (Objective-C)

**Reviewer**: @franklinsch @ethan-kusters

rdar://95220570&97696391


Package.resolved
```diff
@@ -6,8 +6,8 @@
         "repositoryURL": "https://github.com/apple/swift-argument-parser",
         "state": {
           "branch": null,
-          "revision": "d2930e8fcf9c33162b9fcc1d522bc975e2d4179b",
```
franklinsch, line 9: These revision updates for `swift-argument-parser` and `swift-crypto` seem unrelated, did you mean to include them?

bitjammer: This probably came through from a a full replacement of the Package.resolved file. I didn't change the definitions in Package.swift. Should these remain fixed?

franklinsch: Yes please, I'd prefer to have these dependencies updated separately so that we can test them in isolation.

bitjammer: Okay, I'll change that, one moment.

bitjammer: Fixed.

bitjammer:
@swift-ci Please test

franklinsch (approved):
Approving this for release/5.7 as part of SE-0356.

PR #358, opened 2022-08-10
refactors: turn RenderBlockContent's per-case data into separate structs

QuietMisdreavus:
Bug/issue #, if applicable: rdar://97382349

## Summary

This PR is a collection of refactors to RenderBlockContent that change any associated data in its enum cases to a separate struct. Without this change, any new data being added to these cases would become a source-breaking change to anything that used Swift-DocC as a library. As several of these are in the pipeline (https://github.com/apple/swift-cmark/pull/46, https://github.com/apple/swift-docc/pull/53, etc), i wanted to introduce One Big Breaking Change ahead of time, so that any further changes are much easier to adopt.

## Dependencies

None

## Testing

This should not affect the behavior of Swift-DocC. Any testing should ensure that behavior remains the same.

## Checklist

- [ n/a ] Added tests


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
After some discussion offline, i'm going to scrap this branch and try something else. The kinds of changes proposed in [this forums thread](https://forums.swift.org/t/supporting-more-dynamic-content-in-swift-docc-reference-documentation/59527) would still create breaking changes after this refactor, so we'll need to create a different architecture for `RenderBlockContent` to prevent them from being too disruptive.

QuietMisdreavus:
I tinkered around with an alternate design (turning `RenderBlockContent` into a protocol), but shortcomings in the language prevented that from being feasible. I instead opted to keep the enum structure and add a fake case that only serves to "unfreeze" the enum. This will allow us to expand the enum without waiting for https://github.com/apple/swift/issues/55110 to be fixed, or Library Evolution to be allowed for Swift Packages (rdar://78773361). With that in mind, i'm reopening this PR to keep the original implementation.

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Indexing/RenderBlockContent+TextIndexing.swift
```diff
@@ -11,53 +11,55 @@
...
         }
     }
 
     public func rawIndexableTextContent(references: [String : RenderReference]) -> String {
         switch self {
-        case let .aside(_, blocks):
-            return blocks.rawIndexableTextContent(references: references)
-        case let .orderedList(items):
-            return items.map {
+        case let .aside(a):
```
ethan-kusters, line 23: Nit: We don't have a lot of single-letter variable names in the codebase thus far. I think it probably fits in a little better to spell out the name of these types instead?

```suggestion
        case let .aside(aside):
```

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -37,29 +37,208 @@ import Markdown
...
     /// A list of terms.
-    case termList(items: [TermListItem])
+    case termList(TermList)
     /// A table that contains a list of row data.
-    case table(header: HeaderType, rows: [TableRow], metadata: RenderContentMetadata?)
+    case table(Table)
+
+    // Warning: If you add a new case to this enum, make sure to handle it in the Codable
+    // conformance at the bottom of this file, and in the `rawIndexableTextContent` method in
+    // RenderBlockContent+TextIndexing.swift!
```
ethan-kusters, line 66: Hm I hadn't considered this. I guess ideally we wouldn't want to mark the the `_nonfrozenEnum_useDefaultCase` case as deprecated _internally_ and only extenrally so that we can still switch on it within the library and make sure SwiftDocC itself is covering all of bases.

What do you think of the trade-off here? Should we consider not deprecating it so we can switch over it internally? Then we would move the deprecation message into a documentation comment instead.

ethan-kusters (approved):
Looks really good! I think this is going to be really nice moving forward. All my feedback is non-blocking.

ethan-kusters:
@swift-ci please test

PR #379, opened 2022-09-12
Add authoring support for @TabNavigator directive

ethan-kusters:
Bug/issue #, if applicable: rdar://97739641

## Summary

Add support for the tab navigator directive as described here: https://forums.swift.org/t/supporting-more-dynamic-content-in-swift-docc-reference-documentation/59527#tabnavigator-14

## Dependencies

- https://github.com/apple/swift-docc-render/pull/403

### DIFF
--- a/Tests/SwiftDocCTests/Indexing/RenderIndexTests.swift
+++ b/Tests/SwiftDocCTests/Indexing/RenderIndexTests.swift
@@ -657,6 +657,11 @@ final class RenderIndexTests: XCTestCase {
                         "title" : "My Article",
                         "icon" : "plus.svg",
                         "type" : "article"
+                      },
+                      {
+                        "path" : "\/documentation\/bestbook\/tabnavigatorarticle",
+                        "title" : "Tab Navigator Article",
+                        "type" : "article"
                       }
                     ]
                   },
--- a/Tests/SwiftDocCTests/Semantics/Reference/TabNavigatorTests.swift
+++ b/Tests/SwiftDocCTests/Semantics/Reference/TabNavigatorTests.swift
@@ -1,0 +1,201 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2022 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+import XCTest
+@testable import SwiftDocC
+import Markdown
+
+class TabNavigatorTests: XCTestCase {
+    func testNoTabs() throws {
+        let (renderBlockContent, problems, tabNavigator) = try parseDirective(TabNavigator.self) {
+            """
+            @TabNavigator
+            """
+        }
+        
+        XCTAssertNotNil(tabNavigator)
+        
+        XCTAssertEqual(
+            problems,
+            ["1: warning – org.swift.docc.HasAtLeastOne<TabNavigator, Tab>"]
+        )
+        
+        XCTAssertEqual(renderBlockContent.count, 1)
+        XCTAssertEqual(
+            renderBlockContent.first,
+            .tabNavigator(.init(tabs: []))
+        )
+    }
+    
+    func testEmptyTab() throws {
+        let (renderBlockContent, problems, tabNavigator) = try parseDirective(TabNavigator.self) {
+            """
+            @TabNavigator {
+                @Tab("hiya") {
+                    
+                }
+            }
+            """
+        }
+        
+        XCTAssertNotNil(tabNavigator)
+        XCTAssertEqual(
+            problems,
+            ["2: warning – org.swift.docc.Tab.HasContent"]
+        )
+        
+        XCTAssertEqual(renderBlockContent.count, 1)
+        XCTAssertEqual(
+            renderBlockContent.first,
+            .tabNavigator(RenderBlockContent.TabNavigator(
+                tabs: [RenderBlockContent.TabNavigator.Tab(title: "hiya", content: [])]
+            ))
+        )
+        
+    }
+    
+    func testInvalidParametersAndContent() throws {
+        let (renderBlockContent, problems, tabNavigator) = try parseDirective(TabNavigator.self) {
+            """
+            @TabNavigator(tabs: 3) {
+                @Tab("hi") {
+                    Hello there.
+                }
+            
+                @Tab("hey") {
+                    Hey there.
+            
+                    @TabNavigator(weird: true) {
+                        @Tab("bad") {
+                            @Unkown {
+                                
+                            }
+                        }
+                    }
+                }
+            }
+            """
+        }
+        
+        XCTAssertNotNil(tabNavigator)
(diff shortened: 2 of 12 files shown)


ethan-kusters:
@swift-ci please test

Tests/SwiftDocCTests/Indexing/RenderIndexTests.swift
```diff
@@ -657,6 +657,11 @@ final class RenderIndexTests: XCTestCase {
                         "title" : "My Article",
                         "icon" : "plus.svg",
                         "type" : "article"
+                      },
```
daniel-grumberg, line 660: I think the test is now poorly named, since it tests more than CustomIcons at this point. In my mind there are two alternatives. Either we break off the tests into individual tests and instead of `XCTAssertEqual` on the full string, we assert on the presence of individual elements. alternatively we just rename this test to `testRenderIndexGenerationWithBookLikeContent` although I am not sure I like to refer to all of these as being book-like since I can see things like this new tab navigator being useful for large amounts of related screenshots in an article within conceptual documentation for an app or something.

ethan-kusters: Ah I disagree. I think it's still only testing custom icons- if the navigator index doesn't include just regular article pages in it then something has gone _really_ wrong.

I just had to update the test to reflect the content of the DocC catalog since I added an additional page for a different test in `RenderNodeTranslatorTests` but the main thing this is testing for is still the special page that customizes an icon.

Tests/SwiftDocCTests/Semantics/Reference/TabNavigatorTests.swift
```diff
@@ -0,0 +1,202 @@
...
+    
+    func testInvalidParametersAndContent() throws {
+        let (renderBlockContent, problems, tabNavigator) = try parseDirective(TabNavigator.self) {
+            """
+            @TabNavigator(tabs: 3) {
+                @Tab("hi") {
+                    Hello there.
+                }
+            
+                @Tab("hey", bestTab: true) {
```
daniel-grumberg, line 74: Not sure this kind of test belongs in here. It's more checking the directive parser than the tab navigator directive itself. I think as you already went through so much effort to make directive arguments to content so declarative IMHO there is no point checking random keys like that.

Tests/SwiftDocCTests/XCTestCase+LoadingTestData.swift
```diff
@@ -138,11 +138,14 @@ extension XCTestCase {
         var analyzer = SemanticAnalyzer(source: nil, context: context, bundle: bundle)
         let result = analyzer.visit(blockDirectiveContainer)
         
-        let problemIDs = analyzer.problems.map { problem -> String in
-            let line = problem.diagnostic.range?.lowerBound.line.description ?? "unknown-line"
+        let problemIDs = try analyzer.problems.map { problem -> (line: Int, id: String) in
```
daniel-grumberg, line 141: This doesn't feel like it's specific to this change, maybe split it up into a separate commit?

Tests/SwiftDocCTests/Indexing/RenderIndexTests.swift
daniel-grumberg: I see what you mean, I think it's fine for this review then. FTR though I think that for someone new to the code base without the context of this PR, it is less clear what the test is checking for. We should come up with a better way of testing individual render index features without having to do this all the time.

Tests/SwiftDocCTests/Semantics/Reference/TabNavigatorTests.swift
ethan-kusters: Hm yeah that's fair. I'm asserting on the generated RenderBlockContent when an unknown directive is encountered as well below which is a little more interesting but I agree that `bestTab: true` is kind of useless at this point. I can remove it.

daniel-grumberg: The unknown directive case is more interesting in the sense that it maps more closely to the structure we expect from the content specifically for tab navigators.

Tests/SwiftDocCTests/XCTestCase+LoadingTestData.swift
ethan-kusters: I thought about it but it's needed for the test changes in this PR so I think it belongs here.

Some of the diagnostics being asserted on in `TabNavigatorTests` go beyond line `10` so they were being mis-sorted (`11 -` coming before `1 -`). This just resolves it by sorting on the line number specifically instead of only the string.

daniel-grumberg: OK I see, fair enough.

Tests/SwiftDocCTests/XCTestCase+LoadingTestData.swift
```diff
@@ -138,11 +138,14 @@ extension XCTestCase {
...
+            let line = try XCTUnwrap(problem.diagnostic.range).lowerBound.line
             
-            return "\(line): \(problem.diagnostic.severity) – \(problem.diagnostic.identifier)"
-        }.sorted()
+            return (line, "\(line): \(problem.diagnostic.severity) – \(problem.diagnostic.identifier)")
+        }
+        .sorted(by: \.id)
```
daniel-grumberg, line 146: The double sort feels really confusing. there must be a better way of achieving this. If we just need them sorted by line number then let's just keep the second sort. If we then need ordering on other criteria let's make it explicit and do a simple, `.sorted`.

ethan-kusters: Ah I'm open to suggestions. We need it sorted on the line number so that the tests are readable but then we need the sort to be stable between test runs for multiple diagnostics that appear on the same line.

I considered dropping into a full `sorted` but I think those are inherently much more confusing to read.

ethan-kusters: Oh I just hadn't considered sorting the tuple itself – will that work? Agreed that's a much nicer solution if so.

ethan-kusters: Hm yeah a (Int, String) isn't comparable so just `sorted()` doesn't work unfortunately. I'll implement it manually and see how it is.

ethan-kusters: Okay I moved to a custom `.sorted`. Let me know what you think.

ethan-kusters:
@swift-ci please test

Tests/SwiftDocCTests/XCTestCase+LoadingTestData.swift
daniel-grumberg: Interestingly enough you could write the sorted as `.sorted { $0 < $1 } because tuples somehow implement `<` without conforming to comparable apparently. I still think your version is better though

daniel-grumberg:
LGTM

ethan-kusters:
@swift-ci please test

daniel-grumberg (approved):

ethan-kusters:
@swift-ci please test

PR #384, opened 2022-09-14
curate type-subscript symbols in automatic curation

QuietMisdreavus:
Bug/issue #, if applicable: rdar://99921042

## Summary

Currently, we don't curate static subscript symbols in the automatic curation logic. In Xcode, this causes the symbol to "fall out" of curation, appearing at the top level alongside the package itself. This PR adds "type subscripts" to the automatic curation listing, so that these symbols can be properly displayed.

## Dependencies

None

## Testing

The following Swift code will trigger this bug:

```swift
public struct SomeStruct {
    public static subscript(_ idx: Int) -> Int {
        idx
    }
}
```

Steps:
1. Build documentation for a package containing the above code.
2. Ensure that the `static subscript(Int) -> Int` symbol is correctly organized underneath `SomeStruct`.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
+++ b/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
@@ -204,8 +204,6 @@ extension AutomaticCuration {
             case .`method`: return "Instance Methods"
             case .`property`: return "Instance Properties"
             case .`protocol`: return "Protocols"
-            case .snippet: return "Snippets"
-            case .snippetGroup: return "Snippets"
             case .`struct`: return "Structures"
             case .`subscript`: return "Subscripts"
             case .`typeMethod`: return "Type Methods"
@@ -225,9 +223,10 @@ extension AutomaticCuration {
     }
 
     /// The order of symbol kinds when grouped automatically.
+    ///
+    /// Add a symbol kind to `KindIdentifier.noPageKinds` if it should not generate a page in the
+    /// documentation hierarchy.
     static let groupKindOrder: [SymbolGraph.Symbol.KindIdentifier] = [
-        .module,
-
         .`class`,
         .`protocol`,
         .`struct`,
@@ -259,8 +258,5 @@ extension AutomaticCuration {
         .unknownExtendedType,
 
         .extension,
-
-        .snippet,
-        .snippetGroup
     ]
 }
--- a/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
+++ b/Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
@@ -195,6 +195,7 @@ extension AutomaticCuration {
             case .`deinit`: return "Deinitializers"
             case .`enum`: return "Enumerations"
             case .`case`: return "Enumeration Cases"
+            case .extension: return "Extensions"
             case .`func`: return "Functions"
             case .`operator`: return "Operators"
             case .`init`: return "Initializers"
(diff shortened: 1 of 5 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
```diff
@@ -232,6 +232,7 @@ extension AutomaticCuration {
         .`typealias`,
         .`typeProperty`,
         .`typeMethod`,
+        .`typeSubscript`,
```
ethan-kusters, line 235: It looks like `SymbolGraph.Symbol.KindIdentifier` is `CaseIterable` - should we add a test that just confirms we're covering all possible cases here?

QuietMisdreavus: That's a really good idea! I'll write one up.

QuietMisdreavus: I changed the existing `testAutomaticTopics` test, which revealed that a few symbol kinds were also not being curated correctly.

QuietMisdreavus:
@swift-ci Please test

Tests/SwiftDocCTests/Infrastructure/AutomaticCurationTests.swift
```diff
@@ -16,7 +16,7 @@ import XCTest
 class AutomaticCurationTests: XCTestCase {
     func testAutomaticTopics() throws {
         // Create each kind of symbol and verify it gets its own topic group automatically
-        for kind in AutomaticCuration.groupKindOrder where kind != .module {
+        for kind in SymbolGraph.Symbol.KindIdentifier.allCases where kind != .module {
```
ethan-kusters, line 19: Nice 👍🏻

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
```diff
@@ -240,6 +242,14 @@ extension AutomaticCuration {
...
-        .`enum`
+        .`typeSubscript`,
+        .`enum`,
+
+        .unknown,
+
+        .extension,
+
+        .snippet,
+        .snippetGroup
```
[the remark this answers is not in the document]
QuietMisdreavus: Right, i wasn't sure about this one. `snippet` had a group title already, but it wasn't clear whether it should also be curated into the hierarchy. Note that if snippet symbols do exist and they're not filtered out like `module` symbols are, they wouldn't be curated at all, much like the static subscript method in the original issue.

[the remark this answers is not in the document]
bitjammer: Correct, snippets shouldn’t have pages or appear in curations or the navigator at this time. Snippet groups are also not used in the latest design.

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Infrastructure/Topic Graph/AutomaticCuration.swift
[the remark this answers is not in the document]
QuietMisdreavus: I've pushed up a new commit that abstracts out the idea of "this kind of symbol shouldn't generate a page". Right now there's a place where the `module` kind is hardcoded, and i assume this is how those symbols are dropped from the page structure entirely. (I would like confirmation that this guess is correct! 😅) Snippets and snippet groups were added to this list, which is now consulted in the "symbol should get automatically curated" test.

d-ronnqvist (approved):

QuietMisdreavus:
Rebased.

@swift-ci Please test

PR #389, opened 2022-09-23
emit table column alignments into Render JSON

QuietMisdreavus:
Bug/issue #, if applicable: rdar://77749746

## Summary

This PR takes the table alignment info from Swift-Markdown and emits it into Render JSON, so that Markdown tables' column alignment information can be reflected in the final rendering.

## Dependencies

https://github.com/apple/swift-docc-render/pull/441

## Testing

Use the following table as test information:

```markdown
| Column One | Column Two | Column Three | Column Four |
| :--------- | ---------: | :----------: | ----------- |
| one        | two        | three        | four        |
```

Steps:
1. Add the above Markdown to `Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC.md`.
2. Clone and build the Swift-DocC-Render PR.
3. `DOCC_HTML_DIR=/path/to/swift-docc-render/dist bin/preview-docs`
4. Ensure that the table's columns have the proper alignment when rendered.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -244,6 +244,9 @@ public enum RenderBlockContent: Equatable {
         /// The style of header in this table.
         public var header: HeaderType
         /// The text alignment of each column in this table.
+        ///
+        /// A `nil` value for this property is the same as all the columns being
+        /// ``RenderBlockContent/ColumnAlignment/unset``.
         public var alignments: [ColumnAlignment]?
         /// The rows in this table.
         public var rows: [TableRow]
@@ -253,12 +256,21 @@ public enum RenderBlockContent: Equatable {
         public var metadata: RenderContentMetadata?
 
         /// Creates a new table with the given data.
-        public init(header: HeaderType, alignments: [ColumnAlignment]? = nil, rows: [TableRow], extendedData: Set<TableCellExtendedData>, metadata: RenderContentMetadata? = nil) {
+        ///
+        /// - Parameters:
+        ///   - header: The style of header in this table.
+        ///   - rawAlignments: The text alignment for each column in this table. If all the
+        ///     alignments are ``RenderBlockContent/ColumnAlignment/unset``, the ``alignments``
+        ///     property will be set to `nil`.
+        ///   - rows: The cell data for this table.
+        ///   - extendedData: Any extended information that describes cells in this table.
+        ///   - metadata: Additional metadata for this table, if necessary.
+        public init(header: HeaderType, rawAlignments: [ColumnAlignment]? = nil, rows: [TableRow], extendedData: Set<TableCellExtendedData>, metadata: RenderContentMetadata? = nil) {
             self.header = header
             self.rows = rows
             self.extendedData = extendedData
             self.metadata = metadata
-            if let alignments = alignments, !alignments.allSatisfy({ $0 == .unset }) {
+            if let alignments = rawAlignments, !alignments.allSatisfy({ $0 == .unset }) {
                 self.alignments = alignments
             }
         }
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -618,7 +618,14 @@ extension RenderBlockContent.Table: Codable {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         self.header = try container.decode(RenderBlockContent.HeaderType.self, forKey: .header)
-        self.alignments = try container.decodeIfPresent([RenderBlockContent.ColumnAlignment].self, forKey: .alignments)
+
+        let rawAlignments = try container.decodeIfPresent([RenderBlockContent.ColumnAlignment].self, forKey: .alignments)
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -241,11 +243,14 @@ public enum RenderBlockContent: Equatable {
         public var metadata: RenderContentMetadata?
 
         /// Creates a new table with the given data.
-        public init(header: HeaderType, rows: [TableRow], extendedData: Set<TableCellExtendedData>, metadata: RenderContentMetadata? = nil) {
+        public init(header: HeaderType, alignments: [ColumnAlignment]?, rows: [TableRow], extendedData: Set<TableCellExtendedData>, metadata: RenderContentMetadata? = nil) {
```
franklinsch, line 246: We can set a default value of `nil` here to avoid API breakage.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -241,11 +243,14 @@ public enum RenderBlockContent: Equatable {
...
             self.header = header
             self.rows = rows
             self.extendedData = extendedData
             self.metadata = metadata
+            if let alignments = alignments, !alignments.allSatisfy({ $0 == .unset }) {
```
franklinsch, line 251: It seems odd to not assign to the property if they're all `unset` here. Is there a technical reason that's needed? As a user of the API, I'd expect my argument to get assigned to the property.

Tests/SwiftDocCTests/Model/RenderContentMetadataTests.swift
```diff
@@ -157,6 +157,47 @@ class RenderContentMetadataTests: XCTestCase {
 
         try assertRoundTripCoding(renderedTable)
     }
+
```
franklinsch, line 160: Should we also test the extra behavior in the Codable implementation where the `alignments` value is not emitted at all if they're all `unset`?

franklinsch (dismissed):

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
QuietMisdreavus: This is mainly for backwards compatibility with existing DocC-Render - the absence of the `alignments` list is the same as when they're all `unset`, so i decided to drop it to save on encoding space.

Tests/SwiftDocCTests/Model/RenderContentMetadataTests.swift
QuietMisdreavus: This is effectively tested in the existing `testRenderingTableSpans` test, at least the sense that the decoded representation is ensured to be the same. I could extend it to double-check that the decoded `alignments` property is `nil`, if that would be useful.

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
franklinsch: Not emitting it in the encoder seems to be handled in the encode function though, right? It doesn't seem like the extra check here is required.

franklinsch: I see, then it seems like we should implement this filtering in the decoder, rather than the initializer, that way clients that use the Swift model directly aren't affected by this.

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -236,6 +236,11 @@ public enum RenderBlockContent: Equatable {
     public struct Table: Equatable {
         /// The style of header in this table.
         public var header: HeaderType
+        /// The text alignment of each column in this table.
+        ///
+        /// A `nil` value for this property is the same as all the columns being
+        /// ``RenderBlockContent/ColumnAlignment/unset``.
```
franklinsch, line 242: This could be inferred as this type's Equatable conformance implementing this behavior. Maybe we can mention that this is the behavior _in practice_ or something like that.

franklinsch: Or we can actually override `==` and implement the behavior there.

franklinsch (approved):

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
Kyle-Ye: Should a Table (with 3 rows and alignments are [.unset, .unset, .unset]) equal to a Table (with the same 3 rows but alignments are nil)?

Their actual behavior is the same but the current default Equatable implementation is false.

QuietMisdreavus: I've pushed up a new commit that implements this behavior into the Equatable conformance, and adds some tests to ensure that the behavior holds.

QuietMisdreavus:
@swift-ci Please test

Kyle-Ye (approved):

QuietMisdreavus:
@swift-ci Please test

PR #418, opened 2022-11-07
Add Backtrace for non-darwin platfrom

### LINKED ISSUE #414: docc crash traces are not symbolicated on Linux
### Description

When docc crashes on Linux, crash traces are not symbolicated (i.e., symbol names aren't present and appear as `<unavailable>` instead) which makes it difficult to investigate the crash. We should look into enabling symbolication on Linux, possible using https://github.com/swift-server/swift-backtrace.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.

### Expected Behavior

_No response_

### Actual behavior

_No response_

### Steps To Reproduce

_No response_

_No response_

_No response_

Kyle-Ye:
Bug/issue #, if applicable:

Close #414 

## Summary

Add swift-backtrace dependency to add symbolicated docc crashes on non-Darwin platform.(eg. Linux)

## Dependencies

Add dependency https://github.com/swift-server/swift-backtrace for non-Darwin platform (Currently only Linux and Android since it was the current status of Command line interface)

https://github.com/apple/swift-docc/blob/6e35675cd6173426a87e1ba6d9ab4cf4e43c74af/Sources/docc/main.swift#L11-L17

## Testing

Currently, I do not find an effective way to properly test this PR.

## Checklist

- [ ] Added tests


Kyle-Ye:
@swift-ci please test

franklinsch (commented):
Thanks for opening this @Kyle-Ye! This will really help investigate issues on Linux platforms. I believe there will be additional work required here to land swift-backtrace to the Swift toolchains (@d-ronnqvist can probably provide more details). If you trigger a toolchain build from the Swift repo with these changes, I suspect `docc` will crash on launch on Linux because Backtrace is not in the toolchain.

Kyle-Ye:
> Thanks for opening this @Kyle-Ye! This will really help investigate issues on Linux platforms. I believe there will be additional work required here to land swift-backtrace to the Swift toolchains (@d-ronnqvist can probably provide more details). If you trigger a toolchain build from the Swift repo with these changes, I suspect `docc` will crash on launch on Linux because Backtrace is not in the toolchain.

Yeah,I forgot the fact Swift-DocC is part of the toolchain. I'll try to build the toolchain from the PR and find the missing part.

Saying that does it mean we need to include [swift-server/swift-backtrace](https://github.com/swift-server/swift-backtrace)'s source code into part of the swift-toolchain on non-Darwin platform?

ktoso:
I think we need to solve this "properly", and suggest not trying to solve it for docc here right now.

Kyle-Ye:
> I think we need to solve this "properly", and suggest not trying to solve it for docc here right now.

Got it. Do you mean we should better fix it in Swift language runtime level since [swift-server/swift-backtrace](https://github.com/swift-server/swift-backtrace) only servers as a temporary transition solution for some end-user applications. And it should not be depended by swift toolchain infrastructure component (eg. swift-docc)

> The library is designed to fill a gap in backtraces support for Swift on non-Darwin platforms. When this gap is closed at the language runtime level, this library will become redundant and be deprecated.

ktoso:
Yeah that's right; swift-backtrace is a stop-gap that we would not want to get into the toolchain. We hear the feedback that this needs a real fix loud and clear though :+1:

PR #435, opened 2022-11-30
add @CallToAction metadata directive to reference downloadable content

QuietMisdreavus:
Bug/issue #, if applicable: rdar://57847199

## Summary

This PR adds a new metadata directive for articles: `@CallToAction`. Specifying this directive for an article adds a prominent button to the article header, like this:

The `@CallToAction` directive has four possible arguments, in two pairs: `file`/`url` and `purpose`/`label`. One argument from each pair is required.

This directive was proposed in this thread: https://forums.swift.org/t/supporting-sample-code-pages-in-swift-docc/61143

## Dependencies

None

## Testing

Use the following article as test data:

```markdown
# My Cool Sample

@Metadata {
    @CallToAction(url: "https://github.com/apple/swift-docc", purpose: link)
}

This is something cool you can do with this package.
```

Steps:
1. Add the above article as `Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/MyCoolSample.md`.
2. Run `bin/preview-docs`, open the preview, and navigate to "My Cool Sample".
3. Ensure that the "My Cool Sample" article has a button labeled "Visit" that links to the Swift-DocC GitHub repository.

## Checklist

- [x] Added tests


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+    public var label: String? = nil
+
+    static var keyPaths: [String : AnyKeyPath] = [
+        "url"      : \CallToAction._url,
+        "file"     : \CallToAction._file,
+        "purpose"  : \CallToAction._purpose,
+        "label"    : \CallToAction._label,
+    ]
+
+    public func buttonLabel() -> String {
```
ethan-kusters, line 76: Nit: This API might be a little nicer/swifty-er as a computed property:
```suggestion
    public var buttonLabel: String {
```

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+    /// The location of the associated download.
+    @DirectiveArgumentWrapped
+    public var url: URL? = nil
+
+    @DirectiveArgumentWrapped(
+        parseArgument: { bundle, argumentValue in
+            ResourceReference(bundleIdentifier: bundle.identifier, path: argumentValue)
+        }
+    )
+    public var file: ResourceReference? = nil
```
ethan-kusters, line 61: Should we add docs for the other public API in this struct?

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+    public let originalMarkup: Markdown.BlockDirective
+
+    @available(*, deprecated, message: "Do not call directly. Required for 'AutomaticDirectiveConvertible'.")
+    init(originalMarkup: Markdown.BlockDirective) {
+        self.originalMarkup = originalMarkup
+    }
+}
+
+extension CallToAction.Purpose {
+    public func defaultLabel() -> String {
```
ethan-kusters, line 161: Nit: This could also be a computed property

```suggestion
    public var defaultLabel: String {
```

ethan-kusters, line 161: Should this be public API?

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+        "label"    : \CallToAction._label,
+    ]
+
+    public func buttonLabel() -> String {
+        if let label = label {
+            return label
+        } else if let purpose = purpose {
+            return purpose.defaultLabel()
+        } else {
+            fatalError("A valid CallToAction should have either a purpose or label")
```
ethan-kusters, line 82: I was a little confused by this at first. Maybe a comment on the line above explaining that the `validate` function is intended make this state impossible would be helpful?

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+        for bundle: DocumentationBundle,
+        in context: DocumentationContext,
+        problems: inout [Problem]
+    ) -> Bool {
+        var isValid = true
+
+        if self.url == nil && self.file == nil {
+            problems.append(.init(diagnostic: .init(
+                source: source,
+                severity: .error,
```
ethan-kusters, line 97: Generally we want most of the directive parsing issues to be warnings, not errors, since the build can continue even with them.

```suggestion
                severity: .warning,
```

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+                identifier: "org.swift.docc.\(CallToAction.self).missingLink",
+                summary: "\(CallToAction.directiveName.singleQuoted) directive requires `url` or `file` argument",
+                explanation: "The Call to Action requires a link to direct the user to."
+            )))
+
+            isValid = false
+        } else if self.url != nil && self.file != nil {
+            problems.append(.init(diagnostic: .init(
+                source: source,
+                severity: .error,
```
ethan-kusters, line 108:
```suggestion
                severity: .warning,
```

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -0,0 +1,169 @@
...
+                explanation: "Both the `url` and `file` arguments specify the link in the heading; specifying both of them creates ambiguity in where the call should link."
+            )))
+
+            isValid = false
+        }
+
+        if self.purpose == nil && self.label == nil {
+            problems.append(.init(diagnostic: .init(
+                source: source,
+                severity: .error,
```
ethan-kusters, line 121:
```suggestion
                severity: .warning,
```

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
```diff
@@ -344,6 +344,12 @@ struct ReferenceResolver: SemanticVisitor {
         let newDeprecationSummary = article.deprecationSummary.flatMap {
             visitMarkupContainer($0) as? MarkupContainer
         }
+        // If there's a call to action with a local-file reference, change its context to `download`
+        if let downloadFile = article.metadata?.callToAction?.file,
+            var resolvedDownload = context.resolveAsset(named: downloadFile.path, in: bundle.rootReference) {
+            resolvedDownload.context = .download
```
ethan-kusters, line 350: Is this effectively overriding the case where the user provides a local file but with a `link` purpose? (That seems like the right behavior for me – just checking.)

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -779,13 +779,39 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+            }
+        }
         
         collectedTopicReferences.append(contentsOf: contentCompiler.collectedTopicReferences)
         node.references = createTopicRenderReferences()
 
         addReferences(imageReferences, to: &node)
         addReferences(videoReferences, to: &node)
         addReferences(linkReferences, to: &node)
+        addReferences(downloadReferences, to: &node)
```
ethan-kusters, line 814: I guess we need this because we've never had download references in article pages before?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -779,13 +779,39 @@ public struct RenderNodeTranslator: SemanticVisitor {
             
             return seeAlsoSections
         } ?? .init(defaultValue: [])
+
+        if let callToAction = article.metadata?.callToAction {
```
ethan-kusters, line 783: Will this work for symbol pages too? (That can probably be left for a future PR if not, just wondering. I think we already have the limitation where `@Metadata` is only supported in doc extension files fwiw.)

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -779,13 +779,39 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+                node.sampleDownload = .init(
+                    action: .reference(
+                        identifier: downloadIdentifier,
+                        isActive: true,
+                        overridingTitle: callToAction.buttonLabel(),
+                        overridingTitleInlineContent: nil))
+                downloadReferences[url.description] = DownloadReference(
+                    identifier: downloadIdentifier,
+                    renderURL: url,
+                    sha512Checksum: ""
```
ethan-kusters, line 795: Should we make this `nil` instead of an empty string?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -779,13 +779,39 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
         } ?? .init(defaultValue: [])
+
+        if let callToAction = article.metadata?.callToAction {
+            if let url = callToAction.url {
+                let downloadIdentifier = RenderReferenceIdentifier(url.description)
+                node.sampleDownload = .init(
+                    action: .reference(
+                        identifier: downloadIdentifier,
+                        isActive: true,
+                        overridingTitle: callToAction.buttonLabel(),
```
ethan-kusters, line 790: What happens if we don't provide an overriding title? Should we be allowing the renderer to pick the title in some cases?

ethan-kusters (commented):
Looks great! Just left a few thoughts/questions.

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
QuietMisdreavus: I feel like if `CallToAction.Purpose` and `CallToAction.buttonLabel` are public, this might as well be too. It's not important for this implementation either way, though.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: Correct; currently only tutorials use the `downloadReferences` collection.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
QuietMisdreavus: Basically yeah; this makes it so that if someone uses the `file` argument to `@CallToAction`, that file is treated as a download instead of whatever it might otherwise have been. (This let me test using an SVG file as a download, which otherwise is treated as an image asset.)

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: Not with the current implementation, but it's not that much more work to add it there. (I would need to add this block to `visitSymbol()` as well for it to be picked up on symbol pages.)

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: Currently the `DownloadReference` type (and the corresponding Render JSON spec) requires that `sha512Checksum` is non-nullable. We could add that (i remember a conversation with @mportiz08 saying that we likely should) but i haven't done that yet.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: The `buttonLabel` method (now property) checks the `label` parameter as well as the `purpose` parameter, so the "default" label is determined by what `purpose` an author writes. We could make these parameters totally optional and default all the way down to "Download", but i think forcing the `purpose` parameter is good enough.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
ethan-kusters: Makes sense. Thank you!

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
ethan-kusters: If it's small I think it might be worth doing here just to reduce confusion – otherwise we can open a bug for it so that folks are aware.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
ethan-kusters: I suspect DocC-Render will support a null value already. (Maybe @mportiz08 can confirm?) If so – I'd rather make the spec change as part of this PR since providing an empty string is a little confusing.

Tests/SwiftDocCTests/Test Bundles/SampleBundle.docc/MySample.md
```diff
@@ -0,0 +1,9 @@
+# MySample
+
+@Metadata {
+    @CallToAction(url: "https://example.com/sample.zip", purpose: download)
```
ethan-kusters, line 4: Should/do we have a test for the `link` purpose?

QuietMisdreavus: The `testValidDirective` test in `CallToActionTests.swift` at least ensures that the `link` purposes parses correctly. It looks like i don't check for the expected button label, though, so i could extend it to check for that if you think it would be helpful.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
mportiz08: DocC-Render does not use this field today, so it should be safe to make it optional.

Tests/SwiftDocCTests/Test Bundles/SampleBundle.docc/MySample.md
ethan-kusters: Ah I was more concerned about making sure we have a test for ensuring an external link makes it all the way through to the RenderJSON – but actually this test case already covers that doesn't it. I was just confusing the `url` versus `file` case with the `purpose`. So I think we're good – covering the `purpose` couldn't hurt though! Up to you.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
ethan-kusters: @QuietMisdreavus what do you think? Personally I'd rather we just do that here.

QuietMisdreavus: I think it works to have it as part of this PR. I have this implemented locally already, i'll push it up as soon as i can get it working on symbol pages.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: Adding it to symbol pages turned out to be a bit more complicated than i thought, so i'm going to punt on it for now.

QuietMisdreavus:
Rebased to match the latest `main`.

@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
ethan-kusters: Sounds good 👍🏻 I think a GitHub Issue to track this would be helpful so we don't lose track.

ethan-kusters (approved):

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
QuietMisdreavus: I've filed https://github.com/apple/swift-docc/issues/436 to track this.

PR #437, opened 2022-12-07
Improve reliability of WebKit communication bridge

ethan-kusters:
Bug/issue #, if applicable: rdar://103089222

## Summary

Before this change we would send WebKit JSON messages directly by using backtick string literal syntax. This proved problematic since we now need to escape any backticks in the JSON. Instead we can use `JSON.stringify` which _does_ accept a string literal and convert that literal into a JSON object before calling `JSON.parse`

This removes the need for any escaping and fixes some edge cases in WebKit communication.

## Checklist

- [x] Added tests


ethan-kusters:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Communication/WebKitCommunicationBridge.swift
```diff
@@ -49,10 +49,8 @@ public struct WebKitCommunicationBridge: CommunicationBridge {
         do {
             let encodedMessage = try JSONEncoder().encode(message)
             let messageJSON = String(data: encodedMessage, encoding: .utf8)!
-                // Escape backticks to avoid conflicting with JavaScript's template string syntax.
-                .replacingOccurrences(of: "`", with: "\\`")
             
-            evaluateJavaScript("window.bridge.receive(JSON.parse(`\(messageJSON)`))") { _, _ in }
+            evaluateJavaScript("window.bridge.receive(JSON.parse(JSON.stringify(\(messageJSON))))") { _, _ in }
```
franklinsch, line 53: Doesn't this mean that what we're passing to `window.bridge.receive` is a string instead of an object now? Since DocC-Render expects an object, I suspect this change will break things?

dobromir-hristov: No. `JSON.prase` parses it to an object.

franklinsch: Aah I see, the confusion for me was that `messageJSON` isn't actually a JavaScript string literal, it's a JSON object. In that case why not pass it directly with `window.bridge.receive(\(messageJSON))`?

dobromir-hristov: was just about to ask :)

dobromir-hristov (commented):
I think we can send the JSON directly over the bridge without parsing back and forth.

Sources/SwiftDocC/Infrastructure/Communication/WebKitCommunicationBridge.swift
ethan-kusters: Ah right good point... Curious why that wasn't done originally then? But yes that seems to work as well – I'll update the PR.

ethan-kusters:
@swift-ci please test

hqhhuang (approved):
Didn't know that `messageJSON` was already formatted correctly 😁

dobromir-hristov (approved):

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test

ethan-kusters:
@swift-ci please test macOS platform

PR #447, opened 2023-01-03
Autogenerate directive documentation based on models in SwiftDocC framework

ethan-kusters:
## Summary

Adds autogeneration of documentation for Swift-DocC directives. This includes both autogenerating of declarations based on support parameters and child directives, as well as the extraction of documentation comments from what is written in the SwiftDocC framework doc comments.

The most immediate user-facing benefit here is improvements to detail in declarations. But this will also help with the maintainability of Swift-DocC moving forward as directives will have a single definition and source of truth for both documentation and parsing.

|Screenshot 2023-01-02 at 10 02 31 PM | Screenshot 2023-01-02 at 10 02 44 PM |
|--------------|-----------|
| Screenshot 2023-01-02 at 10 03 15 PM | Screenshot 2023-01-02 at 10 03 29 PM |

## Testing

I've updated the gh-pages site on my fork with a preview: https://ethan-kusters.github.io/swift-docc/documentation/docc.

Note that this initial PR is just to set the foundation for autogeneration – there will be future PRs to clean up the documentation for the directives added as a result of this.

### DIFF
--- a/Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
+++ b/Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
@@ -15,6 +15,19 @@ import Markdown
 ///
 /// A "Call to Action" has two main components: a link or file path, and the link text to display.
 ///
+/// The link path can be specified in one of two ways:
+/// - The `url` parameter specifies a URL that will be used verbatim. Use this when you're linking
+///   to an external page or externally-hosted file.
+/// - The `path` parameter specifies the path to a file hosted within your documentation catalog.
+///   Use this if you're linking to a downloadable file that you're managing alongside your
+///   articles and tutorials.
+///
+/// The link text can also be specified in one of two ways:
+/// - The `purpose` parameter can be used to use a default button label. There are two valid values:
+///   - `download` indicates that the link is to a downloadable file. The button will be labeled "Download".
+///   - `link` indicates that the link is to an external webpage. The button will be labeled "Visit".
+/// - The `label` parameter specifies the literal text to use as the button label.
+///
 /// `@CallToAction` requires one of `url` or `path`, and one of `purpose` or `label`. Specifying both
 /// `purpose` and `label` is allowed, but the `label` will override the default label provided by
 /// `purpose`.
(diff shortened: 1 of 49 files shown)


Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
```diff
@@ -15,19 +15,6 @@ import Markdown
 ///
 /// A "Call to Action" has two main components: a link or file path, and the link text to display.
 ///
-/// The link path can be specified in one of two ways:
```
daniel-grumberg, line 18: Did you mean to remove this or does this PR need rebasing?

Sources/SwiftDocC/Semantics/Reference/Row.swift
```diff
@@ -59,10 +63,6 @@ public final class Row: Semantic, AutomaticDirectiveConvertible, MarkupContainin
     ]
     
     /// The number of columns in this row.
-    ///
```
daniel-grumberg, line 62: Should we keep this as a code comment?

daniel-grumberg (approved):
Looks pretty good to me although I would be more comfortable with it if we had some tests to check that the custom symbol graph additions we are doing here work as expected (I assume not everyone will check that the directive documentation when making a change).

daniel-grumberg (approved):

Sources/SwiftDocC/Semantics/Metadata/CallToAction.swift
ethan-kusters: I did mean to remove it since this information is a little redundant with the autogenerated Parameters section – but on a second look I don't think it's fully redundant so I'll add it back for now. We can restructure some of these docs in a future PR.

Sources/SwiftDocC/Semantics/Reference/Row.swift
ethan-kusters: Good idea! Done.

ethan-kusters:
> Looks pretty good to me although I would be more comfortable with it if we had some tests to check that the custom symbol graph additions we are doing here work as expected (I assume not everyone will check that the directive documentation when making a change).

Agreed. I think it makes sense to prioritize landing this now so that we can have up-to-date documentation for Swift 5.8 but I'll plan a follow-up PR that splits this code into a unit-testable library. It's grown a little beyond a simple Swift script now 😊

ethan-kusters:
@swift-ci please test

PR #467, opened 2023-01-23
make CallToAction `url` arguments encode verbatim in Render JSON

QuietMisdreavus:
Bug/issue #, if applicable: rdar://104558809

## Summary

The `url` argument of the `@CallToAction` directive was meant to be copied in verbatim as the `href` attribute of the resulting button. However, due to using the `DownloadReference` type to store this information, URLs that did not include a protocol and domain (i.e. those meant to reference items outside the documentation catalog/archive but ultimately stored on the same server) would be rewritten to be `/downloads/<filename>`. This PR introduces a `CallToActionReference`, compatible with `DownloadReference`, that bypasses the latter's `renderURL` logic, so that the path written into the `url` argument will be copied verbatim.

## Dependencies

None

## Testing

Use the following directive to test:

```markdown
@Metadata {
    @CallToAction(url: "/files/Sample.zip", purpose: download)
}
```

Steps:
1. Add the above metadata to `Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/Directives.md`.
2. Run `bin/preview-docs`.
3. Navigate to the "Directives" article in the preview.
4. Ensure that the "Download" button links to `http://localhost:8080/files/Sample.zip` (and not `http://localhost:8080/downloads/Sample.zip`).

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
@@ -1,0 +1,53 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+/// A specialized ``DownloadReference`` used for ``CallToAction`` directives.
+///
+/// `@CallToAction` directives can link either to a local file or to a URL, whether relative or
+/// absolute. Directives that use the `file` argument will create a ``DownloadReference`` and copy
+/// the file from the catalog into the resulting archive.
+///
+/// A `CallToActionReference` is intended to encode to Render JSON compatible with a
+/// ``DownloadReference``, but with the `url` set to the text given in the `@CallToAction`'s `url`
+/// argument.
+public struct CallToActionReference: RenderReference, URLReference {
+    public static var baseURL: URL = .init(string: "/\(DownloadReference.locationName)/")!
+
+    public var type: RenderReferenceType = .download
+
+    public var identifier: RenderReferenceIdentifier
+
+    enum CodingKeys: String, CodingKey {
+        case type
+        case identifier
+        case url
+    }
+
+    public init(identifier: RenderReferenceIdentifier) {
+        self.identifier = identifier
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
+    }
+
+    public func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(type.rawValue, forKey: .type)
+        try container.encode(identifier, forKey: .identifier)
+
+        // Enter the given URL verbatim into the Render JSON
+        try container.encode(identifier.identifier, forKey: .url)
+    }
+}
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -20,7 +20,7 @@ import Foundation
 /// ``DownloadReference``, but with the `url` set to the text given in the `@CallToAction`'s `url`
 /// argument.
 public struct ExternalLocationReference: RenderReference, URLReference {
-    public static var baseURL: URL = .init(string: "/\(DownloadReference.locationName)/")!
+    public static var baseURL: URL = DownloadReference.baseURL
 
     public private(set) var type: RenderReferenceType = .download
 
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -22,7 +22,7 @@ import Foundation
 public struct ExternalLocationReference: RenderReference, URLReference {
     public static var baseURL: URL = .init(string: "/\(DownloadReference.locationName)/")!
 
-    public var type: RenderReferenceType = .download
+    public private(set) var type: RenderReferenceType = .download
 
     public var identifier: RenderReferenceIdentifier
 
@@ -40,6 +40,7 @@ public struct ExternalLocationReference: RenderReference, URLReference {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
+        self.type = try container.decode(RenderReferenceType.self, forKey: .type)
     }
 
     public func encode(to encoder: Encoder) throws {
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -10,7 +10,7 @@
 
 import Foundation
 
-/// A specialized ``DownloadReference`` used for ``CallToAction`` directives.
+/// A specialized ``DownloadReference`` used for references to external links.
 ///
 /// `@CallToAction` directives can link either to a local file or to a URL, whether relative or
 /// absolute. Directives that use the `file` argument will create a ``DownloadReference`` and copy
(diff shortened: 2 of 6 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public init(identifier: RenderReferenceIdentifier) {
+        self.identifier = identifier
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
```
ethan-kusters, line 42: Will this round-trip correctly?

Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
```diff
@@ -0,0 +1,53 @@
...
+/// A specialized ``DownloadReference`` used for ``CallToAction`` directives.
+///
+/// `@CallToAction` directives can link either to a local file or to a URL, whether relative or
+/// absolute. Directives that use the `file` argument will create a ``DownloadReference`` and copy
+/// the file from the catalog into the resulting archive.
+///
+/// A `CallToActionReference` is intended to encode to Render JSON compatible with a
+/// ``DownloadReference``, but with the `url` set to the text given in the `@CallToAction`'s `url`
+/// argument.
+public struct CallToActionReference: RenderReference, URLReference {
```
ethan-kusters, line 22: Should this be an `ExternalLocationReference` or something along those lines? `CallToActionReference` feels wrong since sometimes we _do_ use `DownloadReference` for `CallToAction` directives.

Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
QuietMisdreavus: If you go `CallToActionReference -> DownloadReference -> DownloadReference`, the second round-trip will come back differently, since the URL will be different. However, if you go `CallToActionReference -> CallToActionReference` or `CallToActionReference -> DownloadReference -> CallToActionReference`, then the information should be preserved. (The second one will have a different JSON encoding when going back to `CallToActionReference`, but the final decoded instance should be the same as the original one, since the only valuable piece of information is the identifier, which remains untouched.)

Should i add more tests to this effect?

Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
QuietMisdreavus: That could work. I wrote the name before i realized that i wanted to keep using `DownloadReference` for `file` arguments, and i never revisited it. 😅

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -0,0 +1,53 @@
...
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+
+/// A specialized ``DownloadReference`` used for ``CallToAction`` directives.
```
ethan-kusters, line 13: Maybe we should update this to explain this is specifically for external links? It seems like we'll likely have a use for this outside of `CallToAction` directives at some point – so might be good to mention it's currently only used there but it doesn't seem overly specialized to me.

Sources/SwiftDocC/Model/Rendering/References/CallToActionReference.swift
ethan-kusters: I think that would be good. This is tricky behavior that seems like it would be easy to break – having a test to enforce it will likely help us down the road.

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    public init(identifier: RenderReferenceIdentifier) {
+        self.identifier = identifier
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+
+        self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
```
ethan-kusters, line 42: We have a separate discussion about the round-tripping already – but shouldn't we be encoding the type here as well?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -789,11 +789,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                         isActive: true,
                         overridingTitle: callToAction.buttonLabel,
                         overridingTitleInlineContent: nil))
-                downloadReferences[url.description] = DownloadReference(
-                    identifier: downloadIdentifier,
-                    renderURL: url,
-                    checksum: nil
-                )
+                callToActionReferences[url.description] = ExternalLocationReference(identifier: downloadIdentifier)
```
ethan-kusters, line 792:
```suggestion
                externalLocationReferences[url.description] = ExternalLocationReference(identifier: downloadIdentifier)
```

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1626,6 +1623,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
     var linkReferences: [String: LinkReference] = [:]
     var requirementReferences: [String: XcodeRequirementReference] = [:]
     var downloadReferences: [String: DownloadReference] = [:]
+    var callToActionReferences: [String: ExternalLocationReference] = [:]
```
ethan-kusters, line 1626:
```suggestion
    var externalLocationReferences: [String: ExternalLocationReference] = [:]
```

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
QuietMisdreavus: I can change the abstract to generalize it, but leave in the full discussion that references `@CallToAction` specifically, since that's what it's used for right now.

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
QuietMisdreavus: Encode the type how? The idea behind this type is that it should encode to JSON compatible with a `DownloadReference`, including the `type` property being set to `"download"`. Are you suggesting that we add an extra field, not used by Swift-DocC-Render, that would indicate the ultimate source of a `"download"` reference?

QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -0,0 +1,53 @@
...
+///
+/// `@CallToAction` directives can link either to a local file or to a URL, whether relative or
+/// absolute. Directives that use the `file` argument will create a ``DownloadReference`` and copy
+/// the file from the catalog into the resulting archive.
+///
+/// An `ExternalLocationReference` is intended to encode to Render JSON compatible with a
+/// ``DownloadReference``, but with the `url` set to the text given in the `@CallToAction`'s `url`
+/// argument.
+public struct ExternalLocationReference: RenderReference, URLReference {
+    public static var baseURL: URL = .init(string: "/\(DownloadReference.locationName)/")!
```
ethan-kusters, line 23:
```suggestion
    public static var baseURL = DownloadReference.baseURL
```

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -0,0 +1,53 @@
...
+
+    public var type: RenderReferenceType = .download
```
ethan-kusters, line 25: Should we require this to be RenderReferenceType.download?

```suggestion
    public private(set) var type: RenderReferenceType = .download
```

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
ethan-kusters: No – sorry I had a type above. I mean *decode* not *encode*.

We're already encoding the property below, shouldn't we be decoding it here?

```suggestion
        self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
        self.type = try container.decode(RenderReferenceType.self, forKey: .type)
```

QuietMisdreavus:
@swift-ci Please test

ethan-kusters (approved):
Looks great!

PR #520, opened 2023-03-26
Add authoring support for the `@PageColor` directive

ethan-kusters:
Bug/issue #, if applicable: rdar://106153042

## Summary

Adds a new `@PageColor` metadata directive that allows for customizing the color used to represent a given page.

`@PageColor` gives authors control over the color used when rendering a page – initially this will affect the background color Swift-DocC-Render uses in the page's introduction (hero) section.

## Example:

    # What's New in SlothCreator

    @Metadata {
        @PageColor(red: 233, green: 58, blue: 43)
    }

    A sloth on a tree wearing a fedora.

    Let's check out what's new in SlothCreator!

    ...

## Details:

`@PageColor` accepts the following parameters:

  - `red`: An integer value between `0` and `255` that represents the amount of red in the color.

  - `green`: An integer value between `0` and `255` that represents the amount of green in the color.

  - `blue`: An integer value between `0` and `255` that represents the amount of blue in the color.

  - `opacity`: An optional floating-point integer value between `0.0` and `1.0` that represents the opacity of the color.

    Defaults to `1.0`.

This forums post for this change is still pending.

## Dependencies

Swift-DocC-Render PR TBD.

## Testing

Add `@PageColor` to the `@Metadata` directive in a Swift-DocC article or extension file and confirm that the resulting RenderNode includes a `color` metadata property.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/References/TopicColor.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/TopicColor.swift
@@ -1,0 +1,36 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2023 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+/// A custom authored color that can be associated with a documentation topic.
+public struct TopicColor: Codable, Hashable {
+    /// An integer value between `0` and `255` that represents the
+    /// amount of red in the color.
+    public let red: Int
+    
+    /// An integer value between `0` and `255` that represents the
+    /// amount of green in the color.
+    public let green: Int
+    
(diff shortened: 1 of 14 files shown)


ethan-kusters:
@swift-ci please test

SamLanier:
> * Why dont we allow them to specify a css color string? - `#fafafa`, `rgb(244,244,244)` or `linear-gradient(to right top, #051937, #004d7a);`?
> * Should we have light/dark mode support?
> * Because we dont allow them to change the icon/text color, someone will surely end up with colors that are not accessible by any means. Should we give them that option or do we try to be smart about it and do some science figure that out 👨‍🔬

1. Yes, it would be great if we could allow for any of these color options (#fafafa, rgb(244,244,244) or linear-gradient(to right top, #051937, #004d7a);)
2. Dark mode support would be a nice to have, imo
3. We should just provide the customization abilities (and suggest best practices separately ; ) )

ethan-kusters:
Thank you for the feedback here!

Closing in favor of https://github.com/apple/swift-docc/pull/525.

PR #599, opened 2023-05-29
build: fix dependencies

compnerd:
We need this to use the new initialiser `init(url:exact:)` for updating the standalone build to match the CI build.

When building in CI, we would build against swift-crypto 2.5.0 as per update-checkout-config.json in apple/swift.  Adjust the standalone builds to match.

When building on CI, we would build against swift-argument-parser 1.2.2 as per update-checkout-config.json in apple/swift.  Make the standalone builds match.


compnerd:
@swift-ci please tes

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #600, opened 2023-05-30
LMDB: introduce a typealias for `mode_t`

compnerd:
`mode_t` is a non-portable extension to C (POSIX).  Introduce a new typealias `ModeType` that represents the underlying type that is used. Provide an alternate definition for the default mode on Windows as it does not have non-user permissions on a file but rather relies on proper ACLing.


compnerd:
@swift-ci please test

compnerd:
CC: @franklinsch

daniel-grumberg (approved):

PR #601, opened 2023-05-30
Model: enable concurrent rendering on Windows

compnerd:
Assume that concurrent rendering on Windows is the proper thing to do. It is unclear why Linux (but not Android) uses serial rendering.  This enables us to make progress with the Windows port.


daniel-grumberg (approved):

compnerd:
@swift-ci please test

PR #602, opened 2023-05-30
FoundationExtensions: enable autoreleasepool on Windows

compnerd:
The `autoreleasepool` is technically required on all non-ObjC runtimes. It is likely better to use the unstable `_runtime(_ObjC)` check but for now simply extend the OS list with Windows.


compnerd:
@swift-ci please test

daniel-grumberg (approved):

PR #603, opened 2023-05-31
Utility: add a Windows path to the synchronisation support

compnerd:
Use a `SRWLOCK` on Windows to support locking.  Slim Reader-Writer locks are lightweight mutexes that closely resemble `pthread_mutex_t` in functionality.  While possible to use `WaitOnAddress` to implement `os_unfair_lock`, that would be more complex and a slightly heavier mutex as a starting point seems reasonable.  If it turns out that the performance really requires an even lighter weight locking primitive, we can refine this to `WaitOnAddress` subsequently.

### DIFF
--- a/Sources/SwiftDocC/Utility/Synchronization.swift
+++ b/Sources/SwiftDocC/Utility/Synchronization.swift
@@ -9,6 +9,9 @@
 */
 
 import Foundation
+#if os(Windows)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

daniel-grumberg (approved):

PR #604, opened 2023-05-31
build: remove the PreviewServer on Windows

compnerd:
This is the critical change to enable Windows support in DocC.  The dependency on NIO is restricted to the PreviewServer and as NIO is incapable of supporting non-POSIX models (or emulation of POSIX models), we cannot use this dependency on Windows.  As a temporary solution, remove the NIO dependency on Windows by removing the PreviewServer.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -63,29 +63,6 @@ let package = Package(
             swiftSettings: swiftSettings
         ),
         
-        // Command-line tool library
-        .target(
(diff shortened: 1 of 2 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

daniel-grumberg (approved):

PR #605, opened 2023-05-31
Utility: disable `Signal` on Windows

compnerd:
Windows does not support the concept of signals and has no equivalent to it.  Simply disable the interface to ensure that we do not accidentally rely on it somewhere.


daniel-grumberg (approved):

compnerd:
@swift-ci please test

PR #608, opened 2023-06-01
build: split the SwiftDocCUtilitiesTest for Windows

compnerd:
We need to replicate the target definition to exclude some files when building for Windows as they depend on NIO to create the PreviewServer. Rather than adding the files and conditionally removing the content, remove the files from being built.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -105,7 +105,7 @@ let package = Package(
 
 // Command-line tool library
 #if os(Windows)
-package.targets.append(
+package.targets.append(contentsOf: [
     .target(
         name: "SwiftDocCUtilities",
         dependencies: [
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

daniel-grumberg (approved):

Kyle-Ye:
IMO, adding `#if os` check on Manifest file is not a good idea because it only indicates the host platform not the destination platform.

eg.
1. We can't use `#if os(iOS)` to exclude some file because the host OS will be `macOS`
2. Later when we support cross build in SwiftPM. eg. Build Linux product on Windows host, it will be a problem too.

In general, I think a conditional check for NIO's available on the destination side is more acceptable for me.

PR #614, opened 2023-06-07
Remove `#if os` check on Package.swift file

Kyle-Ye:
Bug/issue #, if applicable: 

## Summary

A follow-up PR for #608 (Revert the change of Package.swift in that PR)

See https://github.com/apple/swift-docc/pull/608#issuecomment-1580891971

TLDR: Adding #if os check on Manifest file is not a good practice because it only indicates the host platform not the destination platform.

Imagine the following two scenarios: 
1. We have added the cross compile ability to SwiftPM (Build Linux product on Windows) then the check here will give us a wrong answer. 

2. We have added a new platform support which NIO also does not support and NIO support for Windows is ready.

For the original code, we need to remove the exclude for Windows and then add another exclude for the new platform. 

- In this simple case only one dependency NIO is considered, it would be simple to change `#if os(Windows)` to `#if os(newOS)`

- But if there are multi dependency and Windows is getting support for only one dependency then  we will have to maintain such code for the 2 platform

For the updated code, no major changed is needed. If NIO adds Windows support, we just add it to the `condition: .when(platforms: [.macOS, .linux, .android])` array. And that's it.

## Testing

- [ ] Windows Platform build is not checked.

## Checklist

- [x] Added tests

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -62,6 +62,29 @@ let package = Package(
             ],
             swiftSettings: swiftSettings
         ),
+        // Command-line tool library
+        .target(
+            name: "SwiftDocCUtilities",
+            dependencies: [
+                .target(name: "SwiftDocC"),
+                .product(name: "NIOHTTP1", package: "swift-nio", condition: .when(platforms: [.macOS, .iOS, .linux, .android])),
+                .product(name: "ArgumentParser", package: "swift-argument-parser")
+            ],
+            swiftSettings: swiftSettings
+        ),
+        .testTarget(
+            name: "SwiftDocCUtilitiesTests",
+            dependencies: [
+                .target(name: "SwiftDocCUtilities"),
+                .target(name: "SwiftDocC"),
+                .target(name: "SwiftDocCTestUtilities"),
+            ],
+            resources: [
+                .copy("Test Resources"),
+                .copy("Test Bundles"),
+            ],
+            swiftSettings: swiftSettings
+        ),
         
         // Test utility library
         .target(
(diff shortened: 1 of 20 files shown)


compnerd (approved):
Sure, this seems reasonable.  I might want a quick follow up if this regresses the build on Windows (we don't have CI coverage yet :-().

compnerd (approved):
Sure, this seems reasonable.  I might want a quick follow up if this regresses the build on Windows (we don't have CI coverage yet :-().

Kyle-Ye:
@swift-ci please test

Kyle-Ye:
> Sure, this seems reasonable. I might want a quick follow up if this regresses the build on Windows (we don't have CI coverage yet :-().

I'll get it a try on my Windows machine before merging it.

compnerd:
> > Sure, this seems reasonable. I might want a quick follow up if this regresses the build on Windows (we don't have CI coverage yet :-().
> 
> I'll give it a try on my Windows machine before merging it.

Awesome, thank you!  I'm hoping to get the rest of the changes for the tests merged soon.  There is some work that needs to happen with Foundation to improve long paths to improve the test coverage pass rate, but the non-test components should build at least.

Kyle-Ye:
@swift-ci please test

compnerd:
> Sorry swift 5.8 on my Windows computer can only build simple packages. When it comes to swift-docc, it will always fail for some unknown error. 😭

Yeah, there were some recent changes to SPM to support it, so it needs main.

> Maybe we can merge it first and if this regresses the build on Windows I'll do a quick follow up/

I think that is reasonable.

Kyle-Ye:
> > Sorry swift 5.8 on my Windows computer can only build simple packages. When it comes to swift-docc, it will always fail for some unknown error. 😭
> 
> Yeah, there were some recent changes to SPM to support it, so it needs main.
> 
> > Maybe we can merge it first and if this regresses the build on Windows I'll do a quick follow up/
> 
> I think that is the right thing here.

Got it. I'll try it will Swift 5.9 or the latest nightly toolchain later on.

PR #616, opened 2023-06-07
Tests: replace `sleep` with `Thread.sleep(forTimeInterval:)`

compnerd:
`sleep` is not portable as it is not a standard C function.  Use the Foundation function to provide a single codepath across all the targets.

### DIFF
--- a/Tests/SwiftDocCTests/Utility/SynchronizationTests.swift
+++ b/Tests/SwiftDocCTests/Utility/SynchronizationTests.swift
@@ -73,7 +73,7 @@ class SynchronizationTests: XCTestCase {
 
         // Schedule the synchronized block of work asynchronously
         testQueue.async {
-            synced.sync { _ in _ = sleep(5) }
+            synced.sync { _ in Thread.sleep(forTimeInterval: 5) }
         }
         
         // Asynchronously perform a check after 0.25 secs that the lock is locked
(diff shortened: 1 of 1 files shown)


Kyle-Ye (commented):
On the Windows part this is totally fine.

But for macOS or Linux side, the implementation will change from `Darwin.sleep` to a `Thread.sleep` implementation on Foundation.

I am not familiar with the implementation of both and does not know whether it will give us different behavior.

Source: https://github.com/apple/swift-corelibs-foundation/blob/bd2e810a3ff5adf12410666cee74725d94f2dd25/Sources/Foundation/Thread.swift#L121-L156

Should we use a if directive to make sure other platform's behavior is not effected or just use `Thread.sleep` for all platforms? cc @franklinsch 
```
#if os(Windows)
Thread.sleep(xx)
#else
_ = sleep(xx)
#endif

```

compnerd:
`Thread.sleep(forTimeInterval:)` gives a uniform way to handle sleeping.  This is used pretty extensively in the Swift project.  There is a `Sleep` function on Windows that does sleep for the number of seconds.

daniel-grumberg (approved):
Looks good to me since this is just in testing code.

compnerd:
@swift-ci please test

PR #619, opened 2023-06-09
Tests: use `XCTUnwrap` over forced unwrappng

compnerd:
Failure to unwrap would trigger an assertion failure whereas the use of `XCTUnwrap` will fail the test case and permit the remainder of the test suite to be processed.  This allows us to better understand the state of the test suite coverage when porting.


compnerd:
@swift-ci please test

d-ronnqvist (approved):
Thanks for fixing this

daniel-grumberg (approved):

PR #620, opened 2023-06-09
Tests: treat Windows as Linux

compnerd:
Treat Windows similarly to Linux as the functionality is near similar in completeness.  While it is possible to support the DirectoryMonitor on Windows, it is not yet implemented in DocC, so simply treat it as we do Linux.


compnerd:
@swift-ci please test

daniel-grumberg (approved):

Kyle-Ye (approved):

PR #621, opened 2023-06-09
keep the original URL when decoding Download/ExternalLocationReference

QuietMisdreavus:
Bug/issue #, if applicable: rdar://110536969

## Summary

When a DownloadReference or ExternalLocationReference are decoded from JSON, they both modify URL when re-encoded back to JSON. This can cause issues if the URL is different from what either of these types use as their default (`/downloads/file` and the identifier, respectively). This PR tweaks the Codable logic for both of these types to always use the decoded URL when re-encoding.

## Dependencies

None

## Testing

This bug manifests when the Render JSON is modified after rendering, and then re-encoded after the fact. As such, testing is restricted to the automated test added in this PR.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -24,7 +24,16 @@ public struct ExternalLocationReference: RenderReference, URLReference {
 
     public private(set) var type: RenderReferenceType = .externalLocation
 
-    public var identifier: RenderReferenceIdentifier
+    public var identifier: RenderReferenceIdentifier {
+        didSet {
+            // Keep the url property in sync if it was just a copy of the reference identifier
+            if self.url == oldValue.identifier {
+                self.url = identifier.identifier
+            }
+        }
+    }
+
+    private var url: String
 
     enum CodingKeys: String, CodingKey {
         case type
@@ -34,12 +43,19 @@ public struct ExternalLocationReference: RenderReference, URLReference {
 
     public init(identifier: RenderReferenceIdentifier) {
         self.identifier = identifier
+        self.url = identifier.identifier
     }
 
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
+        let url = try container.decodeIfPresent(String.self, forKey: .url)
+        if let url = url {
+            self.url = url
+        } else {
+            self.url = self.identifier.identifier
+        }
         self.type = try container.decode(RenderReferenceType.self, forKey: .type)
     }
 
@@ -47,8 +63,6 @@ public struct ExternalLocationReference: RenderReference, URLReference {
         var container = encoder.container(keyedBy: CodingKeys.self)
         try container.encode(type.rawValue, forKey: .type)
         try container.encode(identifier, forKey: .identifier)
-
-        // Enter the given URL verbatim into the Render JSON
-        try container.encode(identifier.identifier, forKey: .url)
+        try container.encode(url, forKey: .url)
     }
 }
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -24,16 +24,9 @@ public struct ExternalLocationReference: RenderReference, URLReference {
 
     public private(set) var type: RenderReferenceType = .externalLocation
 
-    public var identifier: RenderReferenceIdentifier {
-        didSet {
-            // Keep the url property in sync if it was just a copy of the reference identifier
-            if self.url == oldValue.identifier {
-                self.url = identifier.identifier
-            }
-        }
-    }
+    public let identifier: RenderReferenceIdentifier
 
-    private var url: String
+    private let url: String
 
     enum CodingKeys: String, CodingKey {
         case type
--- a/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
@@ -43,12 +43,7 @@ public struct ExternalLocationReference: RenderReference, URLReference {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
-        let url = try container.decodeIfPresent(String.self, forKey: .url)
-        if let url = url {
-            self.url = url
-        } else {
-            self.url = self.identifier.identifier
-        }
+        self.url = try container.decode(String.self, forKey: .url)
         self.type = try container.decode(RenderReferenceType.self, forKey: .type)
     }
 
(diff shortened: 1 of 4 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -34,21 +43,26 @@ public struct ExternalLocationReference: RenderReference, URLReference {
...
     public init(from decoder: Decoder) throws {
         let container = try decoder.container(keyedBy: CodingKeys.self)
 
         self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
+        let url = try container.decodeIfPresent(String.self, forKey: .url)
+        if let url = url {
+            self.url = url
+        } else {
+            self.url = self.identifier.identifier
+        }
```
ethan-kusters, line 58: Is this actually optional? In the [spec it's not](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/SwiftDocC.docc/Resources/RenderNode.spec.json#L1823) so I think we can depend on it always being there.

Maybe we should just have this?

```suggestion
        self.url = try container.decode(String.self, forKey: .url)
```

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
```diff
@@ -24,7 +24,16 @@ public struct ExternalLocationReference: RenderReference, URLReference {
 
     public private(set) var type: RenderReferenceType = .externalLocation
 
-    public var identifier: RenderReferenceIdentifier
+    public var identifier: RenderReferenceIdentifier {
+        didSet {
+            // Keep the url property in sync if it was just a copy of the reference identifier
+            if self.url == oldValue.identifier {
```
ethan-kusters, line 30: I like this as a way to maintain the existing behavior but it does seem a little odd to me. Just from looking at the JSON spec – I don't think there's any implication that the URL is at all related to the identifier, it just happens to be in our current implementation.

Maybe we should have a `FIXME:` to remove this at some future point and explain in a comment that it's here for backwards compatibility? Or do you think this is something we should support longer-term? 


(To clarify a bit – it makes total sense that we set the `url` from the `identifier` in the `init` of this struct, I'm just not sure we should be providing the syncing behavior.)

Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
```diff
@@ -60,15 +62,41 @@ public struct DownloadReference: RenderReference, URLReference, Equatable {
...
+        case url
+        case checksum
+    }
+
+    public init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.type = try container.decode(RenderReferenceType.self, forKey: .type)
+        self.identifier = try container.decode(RenderReferenceIdentifier.self, forKey: .identifier)
+        self.url = try container.decode(URL.self, forKey: .url)
+        self.urlWasDecoded = true
```
ethan-kusters, line 78: Similar to the other reference, I'm wondering if we need this. It seems like we always encode a URL, so I don't see how we would ever fail to decode one.

QuietMisdreavus: The `urlWasDecoded` property doesn't denote that it was successfully decoded, it's a note to the `encode(to:)` function to not pass the result through `renderURL(for:)`. If i didn't include this property, the existing round-trip tests failed because the URL was being rewritten by `DownloadReference` before being reloaded by `ExternalLocationReference`.

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
QuietMisdreavus: I was just being overly paranoid here. 😅 I can take it out, this shouldn't happen in practice.

Sources/SwiftDocC/Model/Rendering/References/ExternalLocationReference.swift
QuietMisdreavus: I added this to keep the existing behavior when you modify the `identifier` property between initialization and encoding. I took a guess that this was happening in tests somewhere (or in proper code), but i didn't actually investigate whether it was necessary. I'll take a look to see if that assumption holds.

Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
ethan-kusters: Ohhh got it. Okay makes sense. I think a doc comment on the `urlWasDecoded` might help me when I come across this in the future then but not a blocker.

Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
```diff
@@ -60,15 +62,41 @@ public struct DownloadReference: RenderReference, URLReference, Equatable {
...
         // Render URL
-        try container.encode(renderURL(for: url), forKey: .url)
+        if !urlWasDecoded {
+            try container.encode(renderURL(for: url), forKey: .url)
+        } else {
+            try container.encode(url, forKey: .url)
+        }
+    }
+
+    static public func ==(lhs: DownloadReference, rhs: DownloadReference) -> Bool {
```
ethan-kusters, line 96: Off-topic but I wonder if it's possible to make a `@EquatableIgnored` macro like https://docs.devpubs.apple.com/releases/rainbow-beta-1/documentation/observation/observationignored...

ethan-kusters, line 96: Off-topic but I wonder if it's possible to make an `@EquatableIgnored` macro like: https://developer.apple.com/documentation/observation/observationignored...

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/Tutorial/References/DownloadReference.swift
```diff
@@ -26,7 +26,15 @@ public struct DownloadReference: RenderReference, URLReference, Equatable {
     
     /// The location of the downloadable resource.
     public var url: URL
-    
+
+    /// Indicates whether the ``url`` property was loaded from the regular initializer or from the
+    /// `Decodable` initializer.
+    ///
+    /// This is used during encoding to determine whether to filter ``url`` through the
```
ethan-kusters, line 33: This is awesome – thank you!

Tests/SwiftDocCTests/Rendering/SampleDownloadTests.swift
```diff
@@ -240,4 +240,22 @@ class SampleDownloadTests: XCTestCase {
         let reference = try XCTUnwrap(renderNode.references[identifier.identifier])
         XCTAssert(reference is ExternalLocationReference)
     }
+
+    /// Ensure that a DownloadReference where the URL is different from the reference identifier
+    /// can still round-trip through an ExternalLocationReference with the URL and reference identifier intact.
+    func testRoundTripWithDifferentUrl() throws {
```
ethan-kusters, line 246: Could we add a full test fixture for this too? I like this as a targeted test but I think it would be good to include a RenderJSON test fixture where we've manually modified the URL property and ensure it round-trips as expected. Just to be extra-safe.

QuietMisdreavus:
@swift-ci Please test

ethan-kusters (approved):
🚀

QuietMisdreavus:
@swift-ci Please test

PR #629, opened 2023-06-15
Tests: add a missed case of Windows exclusion

compnerd:
This mirrors the Linux and Android paths and excludes this test on Windows.  With this change (and 1 workaround), we can fully execute the test suite (though not yet pass it completely).


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

PR #634, opened 2023-06-18
Tests: disable the LogHandle tests on Windows

compnerd:
`FileHandle` transacts in `HANDLE`s on Windows rather than file descriptors.  This makes it impossible to replace the end point of the pipe after it has been created as the conversion from a HANDLE to a file descriptor transfers ownership and as such the FileHandle does not work with file descriptors at all.  Replicating that behaviour here would detach the `FileHandle` associated with the `Pipe` and thus cannot be done inline.


compnerd:
@swift-ci please test

PR #635, opened 2023-06-21
don't crash on an extension with an empty declaration

QuietMisdreavus:
Bug/issue #, if applicable: rdar://110683602

## Summary

In current development builds of the Swift compiler, some extension block symbols are emitted with an empty `declarationFragments` array. This causes the extension block symbol converter to crash with an out-of-bounds access when it tries to strip away the `extension SomeType` part of the declaration to find its where clause. This PR fixes this logic by guarding against the length of the array as well as its existence.

## Dependencies

None

## Testing

I haven't isolated how to properly reproduce this (other than "export-import Foundation in a project"), so the test is restricted to the sample data i added in this branch.

Steps:
1. `swift run docc convert 'Tests/SwiftDocCTests/Test Bundles/ModuleWithEmptyDeclarationFragments.docc'`
2. Ensure that docc does not crash when processing the included extension symbol graph.

## Checklist

- [x] Added tests


QuietMisdreavus:
@swift-ci Please test

d-ronnqvist (approved):
LGTM

PR #636, opened 2023-06-21
[5.9] don't crash on an extension with an empty declaration

QuietMisdreavus:
- **Explanation**: Current development versions of the Swift compiler emit some extension-block symbols with an empty `declarationFragments` array, causing Swift-DocC to crash. This PR adds a guard to prevent the out-of-bounds access.
- **Scope**: Affects projects which export-import Foundation, or other packages/frameworks which include public underscored protocols.
- **Issue**: rdar://110683602
- **Risk**: Low. The fix does not change anything about the output, as the converted extension symbols get a synthesized declaration anyway.
- **Testing**: An automated test has been added to ensure the issue is fixed. I've also performed manual testing with the original reproduction case and ensured that it stopped crashing.
- **Reviewer**: @theMomax and @d-ronnqvist


QuietMisdreavus:
@swift-ci Please test

PR #638, opened 2023-06-21
deprecate ExternalLocationReference in favor of DownloadReference

QuietMisdreavus:
Bug/issue #, if applicable: rdar://111121386

## Summary

This PR refactors `DownloadReference` and `ExternalLocationReference` to deprecate the latter by incorporating it into the former. Now that there is a property on `DownloadReference` to prevent rewriting the URL upon encoding, this refactor takes advantage of that to incorporate the original uses of `ExternalLocationReference` back into `DownloadReference`.

## Dependencies

None

## Testing

This is primarily an internal refactor, but there are two major changes in the output: `external-assets.json` will no longer be generated (its previous contents will now be folded into the `downloads` property of `assets.json`, and references that would have used the `externalLocation` reference type (i.e. `@CallToAction(url:_:)`) now instead use the `download` reference type.

Steps:
1. `DOCC_JSON_PRETTYPRINT=YES swift run docc convert --emit-digest 'Tests/SwiftDocCTests/Test Bundles/SampleBundle.docc'`
2. Inspect the resulting built archive:
   a. The `external-assets.json` file should not exist.
   b. The `assets.json` file should contain a `download` reference with a `url` of `"https://www.example.com/source-repository.git"`.
   c. The rendered JSON for `MyExternalSample` should contain a reference to `"https://www.example.com/source-repository.git"` with a reference type of `download`.

## Checklist

- [ n/a ] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNode/CodableRenderReference.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNode/CodableRenderReference.swift
@@ -37,14 +37,12 @@ struct CodableRenderReference: Codable {
             reference = try TopicRenderReference(from: decoder)
         case .section:
             reference = try TopicRenderReference(from: decoder)
-        case .download:
+        case .download, .externalLocation:
             reference = try DownloadReference(from: decoder)
         case .unresolvable:
             reference = try UnresolvedRenderReference(from: decoder)
         case .link:
             reference = try LinkReference(from: decoder)
-        case .externalLocation:
-            reference = try ExternalLocationReference(from: decoder)
         }
     }
     
--- a/Tests/SwiftDocCTests/Rendering/SampleDownloadTests.swift
+++ b/Tests/SwiftDocCTests/Rendering/SampleDownloadTests.swift
@@ -152,15 +152,9 @@ class SampleDownloadTests: XCTestCase {
         XCTAssertEqual(ident.identifier, "files/ExternalSample.zip")
 
         // Ensure that the encoded URL still references the entered URL
-        let downloadReference = try XCTUnwrap(renderNode.references[ident.identifier] as? ExternalLocationReference)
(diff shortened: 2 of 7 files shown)


QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/RenderNode/CodableRenderReference.swift
```diff
@@ -37,14 +37,12 @@ struct CodableRenderReference: Codable {
             reference = try TopicRenderReference(from: decoder)
         case .section:
             reference = try TopicRenderReference(from: decoder)
-        case .download:
+        case .download, .externalLocation:
```
daniel-grumberg, line 40: Not sure if this is important, but IIUC the line below won't work because `DownloadReference(from: Decoder)` re-decodes the `type` key?

QuietMisdreavus: Right, i edited this method because i didn't want to introduce a deprecation warning to our builds. I can put it back if that's what y'all would prefer.

Tests/SwiftDocCTests/Rendering/SampleDownloadTests.swift
ethan-kusters, line 1: Sorry – just thought of an additional thing.

Can we add a test that calls `projectFiles()` on a RenderNode that was created with `@CallToAction(...download...)`? Clients expect that calling that method will return the `DownloadReference` for any sample code pages.

QuietMisdreavus:
@swift-ci Please test

ethan-kusters (approved):

ethan-kusters:
@swift-ci please test

QuietMisdreavus:
@swift-ci Please test

PR #647, opened 2023-06-26
Indexing: add a workaround for clearing the index

compnerd:
The unit tests will occasionally desire to clean up the generated index while retaining a reference to the NavigatorIndex instance.  This is prevented on Windows as the LMDB environment will preserve an open handle which results in the index not being removed.

Adding an internal API to support this use case as this is used for tests only AFAICT.

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -285,6 +285,11 @@ public class NavigatorIndex {
             throw Error.missingBundleIndentifier
         }
     }
+
+    internal func clearIndex() throws {
+        self.environment?.close()
+        try FileManager.default.removeItem(at: self.url)
+    }
     
     /// Indicates the page type of a given item inside the tree.
     /// - Note: This information is stored as UInt8 to decrease the required size to store it and make
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -285,11 +285,6 @@ public class NavigatorIndex {
             throw Error.missingBundleIndentifier
         }
     }
-
-    internal func clearIndex() throws {
-        self.environment?.close()
-        try FileManager.default.removeItem(at: self.url)
-    }
     
     /// Indicates the page type of a given item inside the tree.
     /// - Note: This information is stored as UInt8 to decrease the required size to store it and make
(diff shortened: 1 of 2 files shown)


compnerd:
I'm not sure if this is the best approach for this, and I'm open to additional approaches - @d-ronnqvist @daniel-grumberg

compnerd:
@swift-ci please test

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1981,7 +1981,7 @@ class ConvertActionTests: XCTestCase {
         let indexFromConvertAction = try NavigatorIndex.readNavigatorIndex(url: indexURL)
         XCTAssertEqual(indexFromConvertAction.count, 37)
         
-        try FileManager.default.removeItem(at: indexURL)
+        try indexFromConvertAction.clearIndex()
```
Kyle-Ye, line 1984: Since we do not need the logic in normal env and we already have `@testable import`, I think we can directly call it and remove `clearIndex ` defined on `NavigatorIndex.swift`.

```suggestion
        indexFromConvertAction.environment?.close()
        try FileManager.default.removeItem(at: self.url)
```

Kyle-Ye, line 1984: Since we do not need the logic in normal env and we already have `@testable import`, I think we can directly call it and remove `clearIndex ` defined on `NavigatorIndex.swift`.

```suggestion
        indexFromConvertAction.environment?.close()
        try FileManager.default.removeItem(at: self.url)
```

Kyle-Ye (changes requested):

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -285,6 +285,11 @@ public class NavigatorIndex {
             throw Error.missingBundleIndentifier
         }
     }
+
+    internal func clearIndex() throws {
+        self.environment?.close()
+        try FileManager.default.removeItem(at: self.url)
+    }
```
Kyle-Ye, line 292:
```suggestion
```

compnerd:
Thanks @Kyle-Ye I definitely like this approach much better, it is cleaner.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -1981,7 +1981,8 @@ class ConvertActionTests: XCTestCase {
         let indexFromConvertAction = try NavigatorIndex.readNavigatorIndex(url: indexURL)
         XCTAssertEqual(indexFromConvertAction.count, 37)
         
-        try FileManager.default.removeItem(at: indexURL)
+        indexFromConvertAction.environment?.close()
+        try FileManager.default.removeItem(at: self.url)
```
Kyle-Ye, line 1985:
```suggestion
        try FileManager.default.removeItem(at: indexURL)
```

compnerd:
@swift-ci please test

Kyle-Ye (approved):

compnerd:
@swift-ci please test

PR #655, opened 2023-07-07
SwiftDocCUtilitiesTests: adjust `TestFileSystem`'s handling of paths

compnerd:
When constructing a `URL` with a file path, we should always use `init(fileURLWithPath:)` as not all paths are represented as a URL and require additional encoding.  On Windows, this would result in us stripping the drive prefix (`C:`) which would corrupt the path and thus fail to lookup the item.  In the second instance, this allowed us to simply avoid an object copy entirely.

This helps improve the test pass rate significantly on Windows.


compnerd:
@swift-ci please test

d-ronnqvist (approved):

PR #665, opened 2023-07-21
only use the last file component when deciding decoding strategy

QuietMisdreavus:
Bug/issue #, if applicable: rdar://112664398

## Summary

Swift CI has started failing on Swift-DocC's tests, because its workspace folder includes an `@` sign. This confuses the SymbolGraphLoader into thinking that there are no base-module symbol graphs, and thus refusing to use the `concurrentlyAllFiles` decoding strategy. This PR updates that logic to only inspect the last path component instead of the full file path, allowing these `@` sign directories to successfully detect base-module and extension symbol graphs.

## Dependencies

None

## Testing

Steps:
1. Generate symbol graphs for multiple modules (or copy out two symbol graphs from the "Test Resources" directory) into a directory whose name contains `@`.
2. Run `docc convert` in a debugger and ensure that `SymbolGraphLoader` correctly uses the `concurrentlyAllFiles `decoding strategy to load symbol graph files in parallel.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
+++ b/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
@@ -113,7 +113,7 @@ struct SymbolGraphLoader {
         // This strategy benchmarks better when we have multiple
         // "larger" symbol graphs.
         #if os(macOS) || os(iOS)
-        if bundle.symbolGraphURLs.filter({ !($0.pathComponents.last ?? "").contains("@") }).count > 1 {
+        if bundle.symbolGraphURLs.filter({ !$0.lastPathComponent.contains("@") }).count > 1 {
             // There are multiple main symbol graphs, better parallelize all files decoding.
             decodingStrategy = .concurrentlyAllFiles
         }
--- a/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
+++ b/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
@@ -113,7 +113,7 @@ struct SymbolGraphLoader {
         // This strategy benchmarks better when we have multiple
         // "larger" symbol graphs.
         #if os(macOS) || os(iOS)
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -113,7 +113,7 @@ struct SymbolGraphLoader {
         // This strategy benchmarks better when we have multiple
         // "larger" symbol graphs.
         #if os(macOS) || os(iOS)
-        if bundle.symbolGraphURLs.filter({ !$0.path.contains("@") }).count > 1 {
+        if bundle.symbolGraphURLs.filter({ !($0.pathComponents.last ?? "").contains("@") }).count > 1 {
```
ethan-kusters, line 116: This seems kind of fragile to me. How confident are we in the stability of the symbol graph file names that are passed to DocC?

Maybe we should file an issue to look into removing the need for this? Based on the comment it seems like we're interested in only concurrently decoding when we have multiple large symbol graphs so maybe we should instead be looking at the size of each graph.

ethan-kusters (approved):
Looks great! This is making me question are file-path-name-based logic here in general but we would address that separately from this PR.

Thanks for jumping on this.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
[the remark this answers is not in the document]
QuietMisdreavus: I thought i was missing something! I went ahead and used your suggestion.

QuietMisdreavus: There's been no reason to change the symbol graph naming format, and i can only assume that other symbol graph users have come to depend on it by now as well. It's probably worth adding your suggested change, but i'm also looking to get this change landed in release/5.9, so that's probably best as a separate issue/PR.

QuietMisdreavus:
@swift-ci Please test

PR #670, opened 2023-07-24
Tests: standardize both sides of the URL

compnerd:
When testing with a rebound volume mount (i.e. using `subst` on Windows), the paths are compared without canonicalisation for the rebound path.  This causes a test failure as although the paths are logically equivalent, they are not physically equivalent.  The `standardizedPath` here ensures that we resolve the junction allowing the path comparison to succeed.


d-ronnqvist (approved):
LGTM

compnerd:
@swift-ci please test

PR #673, opened 2023-07-29
SwiftDocC: handle arc separator on Windows

compnerd:
The file system representation of the URL is OS dependent.  Since we do not have an easy to get the file system representation of a relative path, perform the arc separator translation explicitly through string manipulation.  This helps fix a few errors on Windows.

### DIFF
--- a/Sources/SwiftDocC/Servers/FileServer.swift
+++ b/Sources/SwiftDocC/Servers/FileServer.swift
@@ -198,7 +198,10 @@ public class MemoryFileServerProvider: FileServerProvider {
     @discardableResult
     public func addFile(path: String, data: Data) -> Bool {
         guard !path.isEmpty else { return false }
-        let trimmed = path.trimmingCharacters(in: slashCharSet)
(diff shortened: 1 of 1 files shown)


compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

d-ronnqvist (approved):
This looks good to me. 

Just a question for my own understanding:
I would have thought that this in-memory file server would work the same on all platforms since it doesn't read files from the file system. 
Is the reason why this didn't work before that the path arguments use a different path component separator on Windows?

compnerd:
The path representation is still used for textual content as well as for the mapping.  The map of the files will sometimes have the primary and sometimes the secondary and in order to compare them, the path must be canonicalised.  Once you have done so, you need to ensure that the path is canonicalised universally.  We could use a single representation - `URL`, but that would be a larger change I suspect.

compnerd:
@swift-ci please test

PR #626, opened 2023-06-14
`@PageKind` sets the kind for the documentation node

QuietMisdreavus:
Bug/issue #, if applicable: rdar://110182993

## Summary

When `@PageKind` is used to set an article's kind, it doesn't actually set the kind for the `DocumentationNode` created from that article, only the `RenderNode` role. This PR sets the node kind for the `DocumentationNode` when the article is loaded, so that it can be used e.g. in the `linkable-entities.json` file emitted from the node.

## Dependencies

None

## Testing

To test, use a documentation catalog that includes an article that uses `@PageKind(sampleCode)`. The bundle at `Tests/SwiftDocCTests/Test Bundles/SampleBundle.docc` has two such articles, `MyExternalSample` and `MyLocalSample`, and is used in the following instructions.

Steps:
1. `swift run docc convert --output-path .build/output.docarchive --emit-digest 'Tests/SwiftDocCTests/Test Bundles/SampleBundle.docc'`
2. Inspect `.build/output.docarchive/linkable-entities.json` and ensure that the pages for `MyExternalSample` and `MyLocalSample` have their `kind` set to `org.swift.docc.kind.sampleCode`.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [ n/a ] Updated documentation if necessary

PR #674, opened 2023-07-31
Fix regression where sample code pages stopped respecting manual curation

ethan-kusters:
Bug/issue #, if applicable: rdar://112223641

## Summary

Fixes a regression where sample code pages stopped respecting manual curation. Sample code pages that were manually curated were also being automatically curated – meaning they would show up twice in the technology's topic structure.

This regression was introduced by:
- #626

The proposed solution is to revert the fix from #626 and instead introduce a targeted fix for ensuring that sample code pages use the correct kind in the link destination summary. This allows sample code pages to continue to pretend to be articles during the compilation process and receive correct autocuration behavior because of that – but still create correct link destination entries.

This is likely not the right long-term fix and we should consider how to handle specialized articles like sample code pages separately. The intent here is to address the regression in a low-risk way that can potentially be brought over to 5.9. 

## Testing

Curate an article with the `@PageKind(sampleCode)` metadata directive. Ensure that article does not appear twice in the navigation sidebar.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -1854,14 +1854,13 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
         in bundle: DocumentationBundle
     ) -> DocumentationContext.Articles {
         articles.map { article in
-            let kind = article.value.metadata?.pageKind?.kind.documentationNodeKind ?? .article
             guard let (documentation, title) = DocumentationContext.documentationNodeAndTitle(
                 for: article,
                 // By default, articles are available in the languages the module that's being documented
                 // is available in. It's possible to override that behavior using the `@SupportedLanguage`
                 // directive though; see its documentation for more details.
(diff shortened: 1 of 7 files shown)


ethan-kusters:
@swift-ci please test

d-ronnqvist (commented):
This seems fine as a fix for the regression but I don't like it as a long term solution. Specifically I don't like that PageKind pages have different "kind" values locally and externally.

So that we don't lose track of this, can we add TODO/FIXME comments in both these places?

ethan-kusters:
> This seems fine as a fix for the regression but I don't like it as a long term solution. Specifically I don't like that PageKind pages have different "kind" values locally and externally.
> 
> So that we don't lose track of this, can we add TODO/FIXME comments in both these places?

Yeah agreed. Good idea – thank you! Added a comment here: https://github.com/apple/swift-docc/pull/674/files#diff-a72178c5b706bcb7382fd00322f921a0b9ba03e844f63a8d5c11398c4b807fabR756.

ethan-kusters:
@swift-ci please test

d-ronnqvist (approved):

ethan-kusters:
@swift-ci please test

PR #677, opened 2023-08-02
Fix `ValidationError` text in `DirectoryPathOption.swift`

MaxDesiatov:
It should either read as `No documentation directories exist` for plural or `No documentation directory exists` for singular. I've updated the text to use the latter.


d-ronnqvist (approved):

MaxDesiatov:
@swift-ci test

PR #712, opened 2023-09-22
Check symbols are deprecated on at least one platform before marking unconditionally deprecated

mayaepps:
Bug/issue #, if applicable: rdar://112197357

## Summary

Currently, DocC checks whether a symbol is either deprecated or unavailable for all platforms it is aware of to determine whether a symbol is unconditionally deprecated.

However, if a symbol is marked unavailable on all the platforms DocC has availability information about, this check causes DocC to mark the symbol unconditionally deprecated.

This PR adds a check that the symbol is deprecated on at least one platform to prevent symbols that are marked _unavailable_ from being displayed as unconditionally _deprecated_.

## Testing

Steps:
1. Create a project that does not give DocC any default platform availability information.
2. Mark a symbol as unavailable.
3. Build documentation and check that the symbol is not deprecated.

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Test Bundles/AvailabilityOverrideBundle.docc/ComplexAvailable.md
```diff
@@ -0,0 +1,11 @@
+# Complex Available Article
```
ethan-kusters, line 1: Is this being used in the test? I'm only seeing `MyClass` being tested above – not a huge deal if we include this as well, just curious if maybe a test case got lost along the way.

ethan-kusters (approved):
Looks great! Nice find. I had one question but it's non-blocking so happy to land this as-is or re-review if you'd like to make adjustments.

Tests/SwiftDocCTests/Test Bundles/AvailabilityOverrideBundle.docc/ComplexAvailable.md
mayaepps: Nope, it's not being used, just ended up being copied in with the symbol graph. I'll remove it since it doesn't really need to be there.

mayaepps:
@swift-ci please test

QuietMisdreavus (approved):
Nice catch, thanks!

PR #807, opened 2024-01-19
Support `unavailable` platform status and updated fallback platforms.

sofiaromorales:
Bug/issue #, if applicable: rdar://88643064

## Summary

DocC adds Mac Catalyst availability when there's iOS availability. With this change, DocC will also add iPadOS as a platform when there is iOS availability since a symbol available for iOS will almost always be available to use on iPad.

In case the user wants to opt out of this behaviour, the Info.plist default availability section now supports the new property `unavailable`, which allows you, on a per-platform basis, to remove the availability information.

```
<key>CDAppleDefaultAvailability</key>
<dict>
<key>MyModule</key>
<array>
	<dict>
		<key>unavailable</key>
		<true/>
		<key>name</key>
		<string>MacCatalyst</string>
	</dict>
</array>
</dict>
```

## Dependencies

N/A

## Testing

1. Preview documentation with the attached catalog and SGF.
2. Assert that the availability section does not show the Mac Catalyst platform.

[Catalog.docc.zip](https://github.com/apple/swift-docc/files/14452386/Catalog.docc.zip)

## Checklist

- [X] Added tests (Changed existing tests)


d-ronnqvist (commented):
> DocC adds Mac Catalyst availability when there's iOS availability. This was necessary before having the ability to emit SGFs for Mac Catalyst, but now that the compiler can emit availability information for this platform, the iOS fallback logic is no longer needed.

This description makes it sound like the Swift compiler gained a new capability but as long as Mac Catalyst has been a platform, the compiler have been able to emit a symbol graph for Mac Catalyst by building for that platform.

Since nothing has changed, this PR is removing an existing behavior without providing an argument for why that behavior should change. It is possible that this fallback behavior should be removed but saying that it's because it's no longer needed is incorrect.

d-ronnqvist (changes requested):
We talked a bit about this offline. There are some bugs with the way that DocC infers Mac Catalyst availability but the reason for this feature is still valid. That said, there may be cases when a developer wants to opt out of this behavior. 

This PR shouldn't remove the feature completely but instead give developers a mechanism to out out. 

Since this behavior only applies to the "default availability" that developers can specify in the Info.plist, my preferred mechanism to out out would be to specify a default availability item in the Info.plist that marks the project as "unconditionally unavailable" for a given platform.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,32 +44,82 @@ public struct DefaultAvailability: Codable, Equatable {
         enum CodingKeys: String, CodingKey {
             case platformName = "name"
             case platformVersion = "version"
+            case unavailable = "unavailable"
```
d-ronnqvist, line 47: minor: It's not necessary to specify the raw value if it's the same as the case name
```suggestion
            case unavailable
```

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,32 +44,82 @@ public struct DefaultAvailability: Codable, Equatable {
...
+        }
+        
+        /// The diferent availability states that can be declared.
+        /// Unavailable or Available with an introduced version.
+        public enum State: Equatable, Hashable, Encodable {
```
d-ronnqvist, line 52: Does this _need_ to be public? If this is declared public then changing its cases will be source breaking.

If it can be made `internal` then that would be the preferred solution. 

If that's not possible, I think that we should change the `available` case to `case available(introduced: String, deprecated: String?, obsoleted: String?)` so that we can hopefully avoid some source breaking changes in the future if we add support for more information. We could also introduce a struct and use that as the enum value instead:
```swift
public struct AvailableInfo { 
   public var introducedVersion: String
}

public enum State {
   case unavailable
   case available(AvailableInfo)
}
```

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,32 +44,82 @@ public struct DefaultAvailability: Codable, Equatable {
...
 
         /// The name of the platform, e.g. "macOS".
         public var platformName: PlatformName
+        
+        /// The availability state, e.g unavailable
+        public var state: State
 
         /// A string representation of the version for this platform.
-        public var platformVersion: String
+        @available(*, deprecated, message: "Use `introducedVersion` instead.")
```
d-ronnqvist, line 64: Please include information about how long this API will be around.

Also, if you provide a `renamed` value then the compiler can create a fixit that updates the call site to the new API.

```suggestion
        @available(*, deprecated, renamed: "introducedVersion", message: "Use 'introducedVersion' instead. This deprecated API will be removed after 5.11 is released")
```

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,32 +44,82 @@ public struct DefaultAvailability: Codable, Equatable {
...
+            }
+        }
 
         /// Creates a new module availability with a given platform name and platform version.
         ///
         /// - Parameters:
         ///   - platformName: A platform name, such as "iOS" or "macOS"; see ``PlatformName``.
-        ///   - platformVersion: A 2- or 3-component version string, such as `"13.0"` or `"13.1.2"`
+        ///   - platformVersion: A 2- or 3-component version string, such as `"13.0"` or `"13.1.2"`.
+        @available(*, deprecated, message: "Use `init(platformName:platformState:) instead.")
```
d-ronnqvist, line 85: We don't need to deprecate this initializer. It could be used as a way to initialize a value without needing to expose `State` as a public type.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -73,7 +131,9 @@ public struct DefaultAvailability: Codable, Equatable {
     /// Creates a default availability module.
     /// - Parameter modules: A map of modules and the default platform availability for symbols in that module.
     public init(with modules: [String: [ModuleAvailability]]) {
-        self.modules = modules
+        self.modules = modules.mapValues { platformAvailabilities -> [DefaultAvailability.ModuleAvailability] in
+            platformAvailabilities
+        }
```
d-ronnqvist, line 136: Why is this change necessary? The closure passed to `mapValues` returns the unmodified value.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
d-ronnqvist: Also, I wonder if there a more descriptive type name that we could use here. The word "state" to me implies that it's a value that changes but this value doesn't. If it was an internal type I would maybe name it `Storage` or `Value` or something along those lines instead.

Sources/SwiftDocC/Utility/Version.swift
```diff
@@ -35,7 +35,8 @@ public struct Version: Codable, RandomAccessCollection, ExpressibleByArrayLitera
     /// - warning: Returns `nil` if the version string contains non-integer or
     ///   negative numeric components; for example the strings "1.2.3-beta6" and "1.-2.3"
     ///   are invalid inputs.
-    public init?(versionString: String) {
+    public init?(versionString: String?) {
```
d-ronnqvist, line 38: Why is this change necessary? It's a breaking change.

Tests/SwiftDocCTests/Rendering/DefaultAvailabilityTests.swift
```diff
@@ -374,4 +380,72 @@ class DefaultAvailabilityTests: XCTestCase {
         
         XCTAssertEqual(renderNode.deprecationSummary?.firstParagraph, expected)
     }
+    
+    func testUnconditionallyUnavailable() throws {
+        let infoPlist = """
+            <plist version="1.0">
+            <dict>
+                <key>CFBundleName</key>
+                <string>MyModule</string>
```
d-ronnqvist, line 389: minor: This property list key isn't used by DocC.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -26,6 +28,12 @@ import Foundation
 ///             <key>version</key>
 ///             <string>Version Number</string>
 ///         </dict>
+///         <dict>
+///             <key>name</key>
+///             <string>Platform Name</string>
```
d-ronnqvist, line 33: minor: I don't think we support specifying the same platform more than once

```suggestion
///             <string>Other Platform Name</string>
```

d-ronnqvist:
This mostly looks good. Just a couple of questions about how we can avoid source breaking changes now and in the future.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: I cannot set it to internal because it's used by a public initializer. I like the struct approach tho.

d-ronnqvist: We don't need to add a new public initializer with a state argument. There's already an initializer that takes a version string and creates a `State.available` value. All we need is an initializer to create an unavailable value. We could do something like
```
public init(unavailablePlatformName: PlatformName) {
    self.platformName = platformName
    self.state = .unavailable
}
```
and not have to expose the `State` in public API

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: Fixed here https://github.com/apple/swift-docc/pull/807/commits/e4a747366cb1985e176f33eabbed1766c591ab35

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: Thx David, resolved

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: Deprecation tag removed 👍

Sources/SwiftDocC/Utility/Version.swift
sofiaromorales: Back to the original init

Tests/SwiftDocCTests/Rendering/DefaultAvailabilityTests.swift
sofiaromorales: Resolved here https://github.com/apple/swift-docc/pull/807/commits/c2a8b16dc10f2e9e624d9ebbb59964ad7b5a4db5

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: Resolved https://github.com/apple/swift-docc/pull/807/commits/768efb21f647046dd13622f1658062f931ef9bd0

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: ✅

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
sofiaromorales: I updated this to be:
```swift
 public init(with modules: [String: [ModuleAvailability]]) {
            self.modules = modules.mapValues { platformAvailabilities -> [DefaultAvailability.ModuleAvailability] in
            // If a module doesn't contain default availability information for any of the fallback platforms,
            // infer it from the corresponding mapped value.
            platformAvailabilities + DefaultAvailability.fallbackPlatforms.compactMap { (platform, fallbackPlatform) in
                if !platformAvailabilities.contains(where: { $0.platformName == platform }),
                   let fallbackAvailability = platformAvailabilities.first(where: { $0.platformName == fallbackPlatform }),
                   let fallbackIntroducedVersion = fallbackAvailability.introducedVersion
                {
                    return DefaultAvailability.ModuleAvailability(
                        platformName: platform,
                        platformVersion: fallbackIntroducedVersion
                    )
                }
                return nil
            }
        }
    }
```

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -146,6 +146,36 @@ struct SymbolGraphLoader {
...
+                    $0.state == .unavailable
+                }.map(\.platformName)
+                // Availability platforms declared in the Info.plist as `available`.
+                defaultAvailableAvailabilities = defaultAvailabilities.filter {
+                    $0.state != .unavailable
+                }
+            }
+            
+            // Platform names registered across the SGFs for each module.
+            let registeredPlatforms: [PlatformName] = unifiedGraph.moduleData.compactMap {
```
d-ronnqvist, line 167: I don't understand the word "registered" in this context. The information that this variable represents is the platforms that have specific symbol graph files. Would a name like `platformsFoundInSymbolGraphs` be clearer about what this is (maybe it doesn't even need that comment to explain what it is)?

```suggestion
            let platformsFoundInSymbolGraphs: [PlatformName] = unifiedGraph.moduleData.compactMap {
```

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -146,6 +146,36 @@ struct SymbolGraphLoader {
         
         self.symbolGraphs = loadedGraphs.mapValues(\.graph)
         (self.unifiedGraphs, self.graphLocations) = graphLoader.finishLoading()
+        
+        unifiedGraphs.forEach { (_, unifiedGraph) in
+            var unifiedGraph = unifiedGraph
```
d-ronnqvist, line 151: fyi: this can be written more simply using a regular for loop
```suggestion
        for var unifiedGraph in unifiedGraphs.values {
```

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -146,6 +146,36 @@ struct SymbolGraphLoader {
...
+            var defaultUnconditionallyUnavailablePlatforms = [PlatformName]()
+            var defaultAvailableAvailabilities = [DefaultAvailability.ModuleAvailability]()
```
d-ronnqvist, line 153: I find both of these variable names hard to read but I'm struggling to come up with better names without making them equally wordy. 

I think the best I've got is `platformsToMarkUnavailableAsFallback` (focusing on what it's used for) and `defaultIntroducedVersions` (trying to describe the specific availability information that this represents).

I used different words ("fallback" and "default") to mean the same thing on those variable names. It may be clearer with the same word in both variables? What do you think of those alternatives?

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -146,6 +146,36 @@ struct SymbolGraphLoader {
...
+
+            if let defaultAvailabilities = bundle.info.defaultAvailability?.modules[unifiedGraph.moduleName] {
+                // Availability platforms declared in the Info.plist as `unavailable`.
+                defaultUnconditionallyUnavailablePlatforms = defaultAvailabilities.filter {
+                    $0.state == .unavailable
+                }.map(\.platformName)
+                // Availability platforms declared in the Info.plist as `available`.
+                defaultAvailableAvailabilities = defaultAvailabilities.filter {
+                    $0.state != .unavailable
+                }
```
d-ronnqvist, line 163: minor: We have a nice `categorize(where:)` extension that simplifies cases like this where one value represent all the `true` cases and the other all the `false` cases 
```suggestion
                let (unavailablePlatforms, availablePlatforms) = defaultAvailabilities.categorize(where: { $0.state == .unavailable })
                
                platformsToMarkUnavailableAsFallback = unavailablePlatforms.map(\.platformName)
                defaultIntroducedVersions = availablePlatforms
```

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -176,34 +206,80 @@ struct SymbolGraphLoader {
...
+            !unconditionallyUnavailablePlatformNames.contains($0.key) &&
+            registeredPlatforms.contains($0.value)
+        }
+        // Platforms that are defined in the Info.plist that had no corresponding SGF
+        // and are not being added as fallback of another platform.
+        let missingAvailabilities = defaultAvailabilities.filter {
+                !missingFallbackPlatforms.keys.contains($0.platformName) &&
+                !registeredPlatforms.contains($0.platformName)
+        }
+        let symbolsWithFallbackAvailability = unifiedGraph.symbols.compactMapValues { symbol  -> UnifiedSymbolGraph.Symbol in
```
d-ronnqvist, line 233: minor: Mapping all the values and then separately replacing the dictionary could result in potentially holding two versions of all the data for a brief moment unless the compiler can optimize it away. 

Since `UnifiedSymbolGraph.Symbol` is a class this mostly means holding two copies of a number of pointers but it could be avoided by directly iterating over the symbols and updating them in place.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -176,34 +206,80 @@ struct SymbolGraphLoader {
...
+            for (selector, _) in symbol.mixins {
+                if var symbolAvailability = (symbol.mixins[selector]?["availability"] as? SymbolGraph.Symbol.Availability) {
+                    guard !symbolAvailability.availability.isEmpty else { continue }
+                    // Add fallback availability.
+                    for (fallbackPlatform, inheritedPlatform) in missingFallbackPlatforms {
+                        if !symbolAvailability.availability.contains(where: { $0.domain?.rawValue == fallbackPlatform.rawValue }) {
+                            symbolAvailability.availability.forEach {
+                                // Add the platform fallback to the availability mixin the platform is inheriting from.
+                                // The added availability only copies the introduced version.
+                                if $0.domain?.rawValue.lowercased() == inheritedPlatform.rawValue.lowercased() {
```
d-ronnqvist, line 244: There are a few different places that make this check, some with a case sensitive comparison and some with a case <em><b>in</b></em>sensitive comparison. 

Maybe it would be worth adding a couple of private extensions in this file to make all these places use the same case sensitivity? For example:
```swift
private extension SymbolGraph.Symbol.Availability {
    func contains(_ platform: PlatformName) -> Bool {
        availability.contains(where: { $0.matches(platform) })
    }
}

private extension SymbolGraph.Symbol.Availability.AvailabilityItem {
    func matches(_ platform: PlatformName) -> Bool {
        domain?.rawValue.lowercased() == platform.rawValue.lowercased()
    }
}
```

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -176,34 +206,80 @@ struct SymbolGraphLoader {
...
+                                    var fallbackAvailability = $0
+                                    fallbackAvailability.deprecatedVersion = nil
+                                    fallbackAvailability.obsoletedVersion = nil
```
d-ronnqvist: Why does the fallback platform only inherit the introduced version? If the symbol is deprecated or obsoleted, wouldn't we want to display that information for the fallback platform as well?

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -176,34 +206,80 @@ struct SymbolGraphLoader {
...
+                        }
+                    }
+                    // Add the missing default platform availability.
+                    missingAvailabilities.forEach { missingAvailability in
+                        if !symbolAvailability.availability.contains(where: { $0.domain?.rawValue == missingAvailability.platformName.rawValue }) {
+                            guard let defaultAvailability = AvailabilityItem(missingAvailability) else { return }
+                            symbolAvailability.availability.append(defaultAvailability)
+                        }
+                    }
+                    symbol.mixins[selector]!["availability"] = symbolAvailability
```
d-ronnqvist, line 259: This key is defined on each mixin kind
```suggestion
                    symbol.mixins[selector]![SymbolGraph.Symbol.Availability.mixinKey] = symbolAvailability
```

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -509,6 +509,22 @@ extension Symbol {
             }
         }
     }
+    
+    /// Merge the different availability variants defined in the unified symbol,
+    /// and update the availability of the canonical symbol to consider all the different availability mixins instead of only the first one.
```
d-ronnqvist, line 514: I don't understand why the symbol semantic was only considering the first availability mixing before this. Can you help explain that?

d-ronnqvist (commented):
This generally looks good. There are a few places where I think more explicit type names or variable names could make the code easier to read and understand.

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,53 +44,112 @@ public struct DefaultAvailability: Codable, Equatable {
         enum CodingKeys: String, CodingKey {
             case platformName = "name"
             case platformVersion = "version"
+            case unavailable
+        }
+        
+        /// The diferent availability states that can be declared.
+        /// Unavailable or Available with an introduced version.
+        internal enum State: Hashable {
```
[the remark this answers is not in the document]
sofiaromorales: Does `AvailabilityInformation` works? `VersionInformation` sound like there's going to be a version number when it could also be unavailable.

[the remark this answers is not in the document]
d-ronnqvist: Yes, it works. 

I feel that it's not _as_ descriptive since this is already a nested type inside of `DefaultAvailability.ModuleAvailability` so it's _almost_ as descriptive as just `Information` except it doesn't make the reader wonder what type of information it is. I viewed `VersionInformation` a little bit like an optional value with a named `none` case. Either there is version information or there isn't.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: I prefer using `default` for both because they refer to information from the default availability section in the Info.plist. Using `fallback` could mistakenly associate it with the term fallback platforms, which we already use for referring to fallback platforms for availability.

I agree that the current names are verbose and very similar, so I suggest renaming them to `defaultUnavailablePlatforms` and `defaultAvailableInformation`.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: Yes this is correct, fallback platforms should inherit all the availability info, I fixed this here https://github.com/apple/swift-docc/pull/807/commits/ee27d5ed540e33ba9136ba7e6a333763ebf51872

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -509,6 +509,22 @@ extension Symbol {
...
+    func mergeAvailabilities(unifiedSymbol: UnifiedSymbolGraph.Symbol) throws {
+        for (selector, mixins) in unifiedSymbol.mixins {
+            let trait = DocumentationDataVariantsTrait(for: selector)
+            if let unifiedSymbolAvailability = mixins[SymbolGraph.Symbol.Availability.mixinKey] as? SymbolGraph.Symbol.Availability {
+                unifiedSymbolAvailability.availability.forEach { availabilityItem in
+                    if (availabilityVariants[trait]!.availability.contains(where: { $0.domain?.rawValue == availabilityItem.domain?.rawValue })) {
```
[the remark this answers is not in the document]
sofiaromorales: It was for the same reason we don't do it here https://github.com/apple/swift-docc/pull/807#discussion_r1513401514, but if the guard statement applies to this case then it also applies to the other.

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
sofiaromorales: I can't really tell, aside from assuming that it went unnoticed.

From the `availability` declaration
```swift
/// The platforms on which the first variant of the symbol is available.
/// - note: Updating this property recalculates `isDeprecated`.
public var availability: SymbolGraph.Symbol.Availability? {
    get { availabilityVariants.firstValue }
    set { availabilityVariants.firstValue = newValue }
}
```
seems like the possibility of having different availability information in each of the unified symbol availability mixing wasn't considered.

If the test were always conducted using an Info.plist, it would be challenging to notice incorrect availability information since all platforms are displayed, but with incorrect versions in some cases.

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -509,6 +509,22 @@ extension Symbol {
             }
         }
     }
+    
+    /// Merge the different availability variants defined in the unified symbol,
+    /// and update the availability of the canonical symbol to consider all the different availability mixins instead of only the first one.
+    func mergeAvailabilities(unifiedSymbol: UnifiedSymbolGraph.Symbol) throws {
```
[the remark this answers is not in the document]
sofiaromorales: Resolved https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -14,7 +14,9 @@ import Foundation
 ///
 /// Default availability is used as a fallback value for symbols without explicit availability information.
 ///
-/// This information can be authored in the bundle's Info.plist file, as a dictionary of module names to arrays of platform "name" and "version" pairs:
+/// This information can be authored in the bundle's Info.plist file, as a dictionary of module names to arrays of platform "name" and "version" pairs,
+/// or in the case where the platform in uncodinionally unavailable, "name" and "unavailable" pairs:
```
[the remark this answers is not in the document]
sofiaromorales: Fixed here https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift
```diff
@@ -36,53 +44,112 @@ public struct DefaultAvailability: Codable, Equatable {
...
+        }
+        
+        public func encode(to encoder: Encoder) throws {
+            var container = encoder.container(keyedBy: CodingKeys.self)
+            try container.encode(platformName, forKey: .platformName)
+            guard let introducedVersion = introducedVersion else {
+                try container.encode(true, forKey: .unavailable)
+                return
+            }
+            try container.encode(introducedVersion, forKey: .platformVersion)
```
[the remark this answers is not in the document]
sofiaromorales: Updated: https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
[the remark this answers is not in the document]
sofiaromorales: https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: Nice, I didn't know about this, thx! https://github.com/apple/swift-docc/pull/807/commits/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: This makes it more concise and easy to read, thx for the suggestion! https://github.com/apple/swift-docc/commit/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: Implemented here https://github.com/apple/swift-docc/commit/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: https://github.com/apple/swift-docc/commit/c55abc329895424275b96a3c47afef3f8c5bb3f9

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
sofiaromorales: Fixed here https://github.com/apple/swift-docc/commit/c55abc329895424275b96a3c47afef3f8c5bb3f9

sofiaromorales:
Thanks for the review and the suggestions @d-ronnqvist 😄 I addressed all of them. 

This pr is ready for another round of review.

sofiaromorales:
@swift-ci please test

d-ronnqvist (approved):

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

Teapane (approved):

PR #812, opened 2024-01-31
Fix "missing technology root" warning for tutorials

sofiaromorales:
Bug/issue #, if applicable: rdar://117866037

## Summary

When either previewing or converting a tutorial, the diagnostic engine was emitting a warning about a missing technology-root, even when the tutorial was well structured and with the appropriate root. This fix resolved the problem by also checking  the context root technology.

## Testing

Steps:
1. Preview the attached tutorial.
2. Assert that is not showing a `No TechnologyRoot to organize article-only documentation.`warning.

[Tutorial.docc.zip](https://github.com/apple/swift-docc/files/14115674/Tutorial.docc.zip)

## Checklist

- [X] Added tests


Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -462,14 +462,18 @@ public struct ConvertAction: Action, RecreatingContext {
...
                     diagnostic: Diagnostic(
                         severity: .warning,
                         identifier: "org.swift.docc.MissingTechnologyRoot",
-                         summary: "No TechnologyRoot to organize article-only documentation.",
-                         explanation: "Article-only documentation needs a TechnologyRoot page (indicated by a `TechnologyRoot` directive within a `Metadata` directive) to define the root of the documentation hierarchy."
+                         summary: "No TechnologyRoot to organize the documentation.",
+                         explanation: """
+                        There was no root found for this documentation catalog.
+                        - For Article-only documentation please add a TechnologyRoot page (indicated by a `TechnologyRoot` directive within a `Metadata` directive) to define the root of the documentation hierarchy.
+                        - For Tutorials please add a table of contents page (indicated by a `Tutorials` directive with the corresponding `Intro` and `Chapters` directive) to define the root of the documentation hierarchy.\n
```
d-ronnqvist, line 475: Nice adding the description for tutorial-only documentation here. 

I think we could make the tutorial-only experience even nicer if we checked if the context had any tutorials and used different diagnostics for article-only and tutorial-only content.  

What do you think?

d-ronnqvist: Actually, now that I think about this some more... since TechnologyRoot is optional now; this error message is only applicable for tutorial-only documentation, right?

sofiaromorales: > Actually, now that I think about this a bit more... since TechnologyRoot is optional now; this error message is only applicable for tutorial-only documentation, right? So I think that this warning only needs to cover tutorial-only documentation.

I considered this, but since the synthesised technology root behaviour is not documented yet (I think) I was not sure if I wanted to remove the warning.

sofiaromorales: What about this? 
```
 There was no root found for this documentation catalog.

For Tutorials please add a table of contents page (indicated by a `Tutorials` directive with the corresponding `Intro` and `Chapters` directive) to define the root of the documentation hierarchy.
```

Emitting the warning when there's no root could be beneficial to give to the user an idea of the problem independently of the type of catalog they're trying to build.

d-ronnqvist: We could possibly do something like 

```swift
let hasTutorial = context.knownPages.contains(where: {
    guard let kind = try? context.entity(with: $0).kind else {
        return false
    }
    return kind == .tutorial || kind == .tutorialArticle
})
```

and customize the error message depending on if we know that the catalog has a tutorial or not.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -462,14 +462,18 @@ public struct ConvertAction: Action, RecreatingContext {
         }
 
         var didEncounterError = analysisProblems.containsErrors || conversionProblems.containsErrors
-        if try context.renderRootModules.isEmpty {
+        if (try context.renderRootModules + context.rootTechnologies).isEmpty {
```
patshaughnessy, line 465: I was reading the code behind `renderRootModules` and `rootTechnologies`, and looks like they return the nodes in the topic graph that have kind=module or kind=technology. But there's also a static DocumentationNode.Kind called `.root`:

```
/// Root-level documentation.
public static let root = DocumentationNode.Kind(name: "Documentation", id: "org.swift.docc.kind.root", isSymbol: false)
```

Does DocC use this node type? Do we need to consider those root nodes here?

sofiaromorales: > Does DocC use this node type? Do we need to consider those root nodes here?

I didn't find anywhere that this kind was assigned to a node, so maybe it was implemented there originally but never used (?). Maybe @d-ronnqvist can better explain if that kind is meant to be used anywhere.

I'm following the same logic used in the previewPaths() method, so I don't think we're missing any other root type.

d-ronnqvist: I don't _think_ that we use `root` for anything. We only have 3 types of top-level pages:
- modules (frameworks/libraries/apps)
- technology root pages
- tutorial table of content pages

Modules use `.module` and IIRC tutorial table of content pages use `.technology` (The implementation calls them "technologies" and it's a bit confusing that it's different from the user-facing terminology). 

AFAICT technology root pages also use `.module` but I don't know if that's a bug or if it's intended behavior.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
sofiaromorales: Implemented here https://github.com/apple/swift-docc/pull/812/commits/71cd2a6ac63320a99807f74770aa7fbed7402cbb

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -3142,13 +3182,13 @@ class ConvertActionTests: XCTestCase {
...
-            "No TechnologyRoot to organize article-only documentation.",
+            "There was no root found for this documentation catalog.",
             "No symbol matched 'ModuleThatDoesNotExist'. Can't resolve 'ModuleThatDoesNotExist'."
-        ])
+        ].sorted())
         
         let logLines = logStorage.text.splitByNewlines
         XCTAssertEqual(logLines.filter { ($0 as NSString).contains("warning:") }.count, 2, "There should be two warnings printed to the console")
-        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("No TechnologyRoot to organize article-only documentation.") }.count, 1, "The root page warning shouldn't be repeated.")
+        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("here was no root found for this documentation catalog.") }.count, 1, "The root page warning shouldn't be repeated.")
```
mayaepps, line 3191:
```suggestion
        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("There was no root found for this documentation catalog.") }.count, 1, "The root page warning shouldn't be repeated.")
```

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -469,14 +469,22 @@ public struct ConvertAction: Action, RecreatingContext {
...
             postConversionProblems.append(
                 Problem(
                     diagnostic: Diagnostic(
                         severity: .warning,
                         identifier: "org.swift.docc.MissingTechnologyRoot",
-                         summary: "No TechnologyRoot to organize article-only documentation.",
-                         explanation: "Article-only documentation needs a TechnologyRoot page (indicated by a `TechnologyRoot` directive within a `Metadata` directive) to define the root of the documentation hierarchy."
+                         summary: "There was no root found for this documentation catalog.",
+                         explanation: hasTutorial ? """
+                        For Tutorials please add a table of contents page (indicated by a `Tutorials` directive with the corresponding `Intro` and `Chapters` directive) to define the root of the documentation hierarchy. \n
```
mayaepps, line 486: Out of curiosity, do you think it would be helpful to add a link to https://www.swift.org/documentation/docc/tutorials?

sofiaromorales: > Is there a precedent with adding documentation links in our diagnostics? I am not sure that is something we want to do, it would mean we need to make sure the link remains up to date. @d-ronnqvist what do you think?

From https://github.com/apple/swift-docc/pull/664#discussion_r1275305566

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -469,14 +469,22 @@ public struct ConvertAction: Action, RecreatingContext {
...
+                return kind == .tutorial || kind == .tutorialArticle
+            })
             postConversionProblems.append(
                 Problem(
                     diagnostic: Diagnostic(
                         severity: .warning,
                         identifier: "org.swift.docc.MissingTechnologyRoot",
-                         summary: "No TechnologyRoot to organize article-only documentation.",
-                         explanation: "Article-only documentation needs a TechnologyRoot page (indicated by a `TechnologyRoot` directive within a `Metadata` directive) to define the root of the documentation hierarchy."
+                         summary: "There was no root found for this documentation catalog.",
```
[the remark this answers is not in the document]
d-ronnqvist: I don't know if fixits/solutions support defining new files or if they can only modify existing files. If it's possible to define a new file in a fixit/solution then maybe we could reuse the init command code to fill in most of the necessary information for the developer

```swift
let tableOfContentsFilename = "table-of-contents.tutorial"
let source = rootURL?.appendingPathComponent(tableOfContentsFilename)

Solution(
    summary: "Create a `@Tutorials` table of content page",
    replacements: [
        Replacement(
            range: .init(line: 1, column: 1, source: source) ..< .init(line: 1, column: 1, source: source),
            replacement: CatalogTemplateKind.tutorialTemplateFiles(converter.firstAvailableBundle()?.displayName ?? "Project Name")[tableOfContentsFilename]! // this file name is known to exist in the dictionary
        )
    ]
)
```

d-ronnqvist (changes requested):

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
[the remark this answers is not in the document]
sofiaromorales: I added your suggestion and tweaked how `formattedDiagnosticSource` works in order to be able to suggest addition of new content files without having to refactor the `Replacement` struct.

```swift
guard let diagnosticRange = diagnostic.range
      else {
          // If the replacement operation involves adding new files,
          // emit the file content as an addition instead of a replacement.
          //
          // Example:
          // --> /path/to/new/file.md
          // Summary
          // suggestion:
          // 0 + Addition file and
          // 1 + multiline file content.
          var addition = ""
          solutions.forEach { solution in
              addition.append("\n" + solution.summary)
              solution.replacements.forEach { replacement in
                  let solutionFragments = replacement.replacement.split(separator: "\n")
                  addition += "\nsuggestion:\n" + solutionFragments.enumerated().map {
                      "\($0.offset) + \($0.element)"
                  }.joined(separator: "\n")
              }
          }
          return "\n--> \(formattedSourcePath(url))\(addition)"
      }
```

Since the diagnostic does not belongs to an existing range in the source code the `diagnostic.range` will be nil and  we log the suggestion.

This allows us to use fixits/solutions for defining new files instead of only modify existing files.

The other possible improvement to do here is making `range` in `Replacement` optional, but I don't know if that might be a breaking API change. 

Implemented here https://github.com/apple/swift-docc/pull/812/commits/d7720d96b668f7df2ac407f334948fcceb1d4e69.

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

patshaughnessy (changes requested):

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -483,15 +483,39 @@ public struct ConvertAction: Action, RecreatingContext {
         }
 
         var didEncounterError = analysisProblems.containsErrors || conversionProblems.containsErrors
-        if try context.renderRootModules.isEmpty {
+        let hasTutorial = context.knownPages.contains(where: {
+            guard let kind = try? context.entity(with: $0).kind else { return false }
+            return kind == .tutorial || kind == .tutorialArticle
```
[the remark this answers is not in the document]
sofiaromorales: `Volume`, `chapter`, etc., are all considered child directives. The only permissible top-level directives are `Tutorials` `Tutorial` and `Article` as outlined in the documentation: https://www.swift.org/documentation/docc/tutorial-syntax.

If the user adds any other directive as a top-level directive, they will encounter the following error:

```
warning: No valid content was found in this file.
A '.tutorial' file must contain a top-level directive ('Tutorials', 'Tutorial', or 'Article')
and valid child content. Only '.md' files support content without a top-level directive.
```
This warning already specifies the valid directives.

I don't believe this conditional logic needs to account for any other page types.

Regarding the addition of a method to categorize kinds based on the documentation type, I propose something like the following:

```swift
public static let tutorialValues: [DocumentationNode.Kind] = [
    .tutorials, .tutorial, .article, .volume, (etc.)
]
```
This introduces more organization into the code, but I don't see it as a blocker for the pull request.

@patshaughnessy, could you please clarify why this is considered a blocker for the bug we're trying to solve?

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -483,15 +483,39 @@ public struct ConvertAction: Action, RecreatingContext {
...
                     diagnostic: Diagnostic(
+                        source: source,
                         severity: .warning,
-                        identifier: "org.swift.docc.MissingTechnologyRoot",
-                         summary: "No TechnologyRoot to organize article-only documentation.",
-                         explanation: "Article-only documentation needs a TechnologyRoot page (indicated by a `TechnologyRoot` directive within a `Metadata` directive) to define the root of the documentation hierarchy."
-                     )
+                        identifier: "org.swift.docc.MissingTableOfContents",
+                        summary: "Missing tutorial table of contents page.",
+                        explanation: "`@Tutorial` and `@Article` pages require a `@Tutorials` table of content page to define the documentation hierarchy."
```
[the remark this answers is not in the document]
sofiaromorales: I'm not entirely sure I understand this 🤔. Are you suggesting adding a note to the warning indicating that they are converting a tutorial catalog?

If the user receives this warning, it's because they are attempting to convert a `.tutorial` file. Additionally, the `@Article` directive is only supported in tutorials. Therefore, by this point, the user may already realize that these are tutorial-related directives.

What additional clarification do you believe is necessary?

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -483,15 +483,39 @@ public struct ConvertAction: Action, RecreatingContext {
...
+                    ),
+                    possibleSolutions: [
+                        Solution(
+                            summary: "Create a `@Tutorials` table of content page.",
+                            replacements: [
+                                Replacement(
+                                    range: .init(line: 1, column: 1, source: source) ..< .init(line: 1, column: 1, source: source),
+                                    replacement: CatalogTemplateKind.tutorialTemplateFiles(converter.firstAvailableBundle()?.displayName ?? "Project Name")[tableOfContentsFilename]! // this file name is known to exist in the dictionary
```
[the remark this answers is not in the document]
sofiaromorales: I'm not really up for tweaking the `CatalogTemplateKind`  because its content doesn't change frequently unless there are updates to the directives or how tutorials need to be set up. It's hard-coded and doesn't rely on user input so we can assure that it does exists. Plus, if there is a change, the unit test would catch the error before it becomes a major issue.

Still, with this change, we can ensure that we don't break DocC if this unlikely scenario does occur:

```swift
var replacements = [Replacement]()
            if let tableOfContentsTemplate = CatalogTemplateKind.tutorialTemplateFiles(converter.firstAvailableBundle()?.displayName ?? "Tutorial Name")[tableOfContentsFilename] {
                replacements.append(
                    Replacement(
                        range: .init(line: 1, column: 1, source: source) ..< .init(line: 1, column: 1, source: source),
                        replacement: tableOfContentsTemplate
                    )
                )
            }
            postConversionProblems.append(
                Problem(
                    diagnostic: Diagnostic(
                        source: source,
                        severity: .warning,
                        identifier: "org.swift.docc.MissingTableOfContents",
                        summary: "Missing tutorial table of contents page.",
                        explanation: "`@Tutorial` and `@Article` pages require a `@Tutorials` table of content page to define the documentation hierarchy."
                    ),
                    possibleSolutions: [
                        Solution(
                            summary: "Create a `@Tutorials` table of content page.",
                            replacements: replacements
                        )
                    ]
                )
            )
```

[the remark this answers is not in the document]
sofiaromorales: Fixed here https://github.com/apple/swift-docc/pull/812/commits/cd0aad7fb2f4ade7bc5e3def5bd6f815da04d940

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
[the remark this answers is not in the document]
patshaughnessy: Great explanation - thanks! No, categorizing these is not a blocker but might be nice to do someday.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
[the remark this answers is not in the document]
patshaughnessy: This avoids the force-unwrap, thank you. Could we at least convert the keys in `CatalogTemplateKind` to constants? Something like:

```
let tableOfContentsFileName = "table-of-contents.tutorial"
```

etc.. and refer to it here, instead of copy-pasting the string literal?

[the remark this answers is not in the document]
sofiaromorales: Extensions don't allow stored properties, so this would have to be a computed property.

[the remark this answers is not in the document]
patshaughnessy: Or just another static function could work.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
```diff
@@ -3135,21 +3163,20 @@ class ConvertActionTests: XCTestCase {
...
             "No symbol matched 'ModuleThatDoesNotExist'. Can't resolve 'ModuleThatDoesNotExist'."
-        ])
+        ].sorted())
         
         let logLines = logStorage.text.splitByNewlines
-        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("warning:") }.count, 2, "There should be two warnings printed to the console")
-        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("No TechnologyRoot to organize article-only documentation.") }.count, 1, "The root page warning shouldn't be repeated.")
+        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("warning:") }.count, 1, "There should be two warnings printed to the console")
+        XCTAssertEqual(logLines.filter { ($0 as NSString).contains("There was no root found for this documentation catalog.") }.count, 0, "The root page warning shouldn't be repeated.")
         XCTAssertEqual(logLines.filter { ($0 as NSString).contains("No symbol matched 'ModuleThatDoesNotExist'. Can't resolve 'ModuleThatDoesNotExist'.") }.count, 1, "The link warning shouldn't be repeated.")
```
d-ronnqvist, line 3180: minor: Now that there's only one expected warning we could probably make a single assertion here

```suggestion
        XCTAssertEqual(logLines.filter { $0.hasPrefix("warning: No symbol matched 'ModuleThatDoesNotExist'. Can't resolve 'ModuleThatDoesNotExist'.") }.count, 1)
```

Otherwise, the first assertion message says "two warnings" but expects a single warning.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -483,15 +483,43 @@ public struct ConvertAction: Action, RecreatingContext {
...
+        let hasTutorial = context.knownPages.contains(where: {
+            guard let kind = try? context.entity(with: $0).kind else { return false }
+            return kind == .tutorial || kind == .tutorialArticle
+        })
+        // Warn the user if the catalog is a tutorial but does not contains a table of contents
+        // and provide template content to fix this problem.
+        if (
+            context.rootTechnologies.isEmpty &&
+            hasTutorial &&
+            !analysisProblems.contains(where: {$0.diagnostic.identifier == "org.swift.docc.HasExactlyOne<Tutorials, Intro>.Missing"})
```
d-ronnqvist, line 495: Is there a test that checks this behavior? I'm assuming that this check is here so that if the developer has an _incomplete_ table of contents file we don't make a warning with a suggestion to create a new file?

d-ronnqvist (approved):
This looks good to me. 

One comment about potentially adding one more test but I don't consider that blocking.

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
sofiaromorales: I ended up removing this check.

Tests/SwiftDocCUtilitiesTests/ConvertActionTests.swift
sofiaromorales: Fixed here https://github.com/apple/swift-docc/pull/812/commits/55585f0308506bc3a8b74b51485c18b35aab0045

Sources/SwiftDocCUtilities/Action/Actions/Convert/ConvertAction.swift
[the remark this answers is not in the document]
sofiaromorales: Implemented here https://github.com/apple/swift-docc/pull/812/commits/55585f0308506bc3a8b74b51485c18b35aab0045

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

sofiaromorales:
@swift-ci please test

PR #818, opened 2024-02-06
build: add an initial CMake based build system for DocC

compnerd:
Using this (along with its dependencies) allows us to significantly reduce the build times for the Swift toolchain (~7% of the overall build time in local testing).


Sources/SwiftDocC/CMakeLists.txt
```diff
@@ -0,0 +1,458 @@
+#[[
+This source file is part of the swift-format open source project
+
+Copyright (c) 2024 Apple Inc. and the DocC project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
+#]]
+
+add_library(SwiftDocC
```
d-ronnqvist, line 10: Does these lists need to be maintained whenever we add a new file to either of these targets?

compnerd: Yes, that is the general pattern for CMake - you manage the source list explicitly so that the build system knows what is part of the build and what is not.

compnerd:
@swift-ci please test

Sources/SwiftDocC/CMakeLists.txt
d-ronnqvist: Is there any way to specify directories instead of files lists? I'm concerned about the maintenance burden of keeping this list up to date.

compnerd: No, not without basically making the build system brittle and unreliable. CI would easily catch any files removed. Any file that the build doesn't catch where the list is not maintained would imply that it can be dropped because there is no reference to the types/functions it defines.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

d-ronnqvist (changes requested):
This is how I understand the implications of these changes:

- Using CMake instead of Swift Package Manager doesn't enable new functionality (for example building on new platforms) but it offers faster builds.
- The CMake files require constant maintenance to work and will break if a new Swift source file is missing from the file list.
- To catch issues of a missing file in the CMake file list, a new CI needs to run the CMake build. 

Based on that understanding I inspected the SwitDocC git history. The numbers below were computed when [3cfeeb64b96709ca284d845637c75f2f8c248f32](https://github.com/swiftlang/swift-docc/commit/3cfeeb64b96709ca284d845637c75f2f8c248f32) was the latest commit on main.

-----

In the last 30 days there was 20 new commit on main. These commits added or removed 18 files (13 added, 5 removed). In those 30 days there were 21 workdays. This means that on average over the last 30 days there was a new commit approximately every 1.05 days and on average a file added or removed on average every 1.17 days. 

This means that if the CMake file list needed to be updated for every added or removed file, the data from the last 30 days indicate that it would need to be updated on average every 1.17 days. 

However, about half of the added or removed files were tests, which AFAICT the CMake configuration doesn't try to build. There were 8 added or removed files in the "Sources" directory (6 added 2 removed) over the last 30 days. This means that there was a file added to or removed from the "Sources" directory on average every 2.63 days over the last 30 days. This data from the last 30 days indicate that the CMake file list would need to be updated on average every 2.63 days (almost twice a week).

If we instead expand this analysis to all the commits since February 6, when this PR was first opened, there were 175 commits, 108 added or removed files, 56 added or removes files in the "Sources" directory, and 205 workdays. 

In this time period there was a new commit on average every 1.17 days and a file was added to or removed from the "Sources" directory on average every 3.66 days. This data since February 6 indicate that the CMake file list would need to be updated on average every 3.66 days (almost 3 times in two weeks).

-----

These are just estimates but the actual number is probably somewhere in this ballpark; the CMake file list would need to be updated about once or twice every week (around 32%-40% of all commits) ongoing for as long as DocC is developed. This IMO is a rather frequent maintenance burden. 

~7% faster toolchain builds is not insignificant but it's a benefit that only applies to people who use CMake and the ongoing maintenance burden of keeping the file list up-to-date applies to everyone, whether or not they use CMake to build DocC.

Unless there's some broader project goal that I'm not aware of here, in my very personal opinion I don't feel that this benefit is worth the associated cost. 

My concerns would be alleviated if the CMake file list was automatically keep up-to-date. 

My concerns would be lessened to a point where the benefits might outweigh the costs if:
- it was possible to easily regenerate the CMake file list based on `find Sources/SwiftDocC -type f -name "*.swift" | sort` or something similar
- there was some automatic validation that the CMake file list was up-to-date (for example a CI that builds using CMake or a new verification step in `bin/tests` that verify the file list contents)
- "CONTRIBUTING.md" contained information about how to update the CMake file list when adding a new file that we could refer new contributors to.

CMakeLists.txt
```diff
@@ -0,0 +1,33 @@
+#[[
+This source file is part of the swift-format open source project
+
+Copyright (c) 2024 Apple Inc. and the DocC project authors
+Licensed under Apache License v2.0 with Runtime Library Exception
+
+See https://swift.org/LICENSE.txt for license information
```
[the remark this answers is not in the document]
compnerd: This was copied from the DocC sources when I wrote the original patch. We can change the text here.

compnerd:
> Using CMake instead of Swift Package Manager doesn't enable new functionality (for example building on new platforms) but it offers faster builds.

This is not entirely correct. It does enable new functionality - CMake has a broad platform support and bringing up Swift on new platforms is made easier by having the CMake based build. It allows bootstrapping without cross-compiling which is not always possible (e.g. I cannot use the macOS SDK on non-Apple hardware, Windows SDK is not available on macOS, etc). We currently are unable to ship DocC on Windows ARM64 due to the lack of CMake support.

> The CMake files require constant maintenance to work and will break if a new Swift source file is missing from the file list.

The "constant maintenance" feels unnecessarily critical. It does require that the source list is maintained (this information is used for ensuring that build targets are properly rebuilt).

> To catch issues of a missing file in the CMake file list, a new CI needs to run the CMake build.

We could change the CI to only build with CMake, and that would change this to "a new CI needs to run the SPM build".

matthewbastien:
I'd also like to add that this is currently blocking the documentation preview support in the VS Code Swift extension. The current approach is to have SourceKit-LSP handle documentation requests from the editor since it is the source of truth for all symbol related requests. However, SourceKit-LSP has a cmake build on Windows and, in order for it to use SwiftDocC as a dependency, requires that SwiftDocC also build with cmake on Windows.

I can take another approach of creating a new executable (e.g. `doccd` or some other appropriate name) that can communicate bi-directionally with SourceKit-LSP using some protocol like JSON-RPC, but that feels like a lot of duplicated code between the two and also doesn't solve the problem that docc isn't available on Windows arm64.

compnerd:
@swift-ci please test

compnerd:
@swift-ci please test

CONTRIBUTING.md
```diff
@@ -350,7 +350,23 @@ by running the test suite in a Docker environment that simulates Swift on Linux.
...
+(you cannot execute the test suite with this build).
+
+CMake requires that the full file list is kept up-to-date. When adding or
+removing files in a given module, the `CMakeLists.txt` list must be updated to
+the file list.
+
+You can use the following 1-line script to enumerate the files in the module:
+
+```bash
+python -c "print('\n'.join((f'{chr(34)}{path}{chr(34)}' if ' ' in path else path) for path in sorted(str(path) for path in __import__('pathlib').Path('.').rglob('*.swift'))))"
```
d-ronnqvist, line 368: I get a syntax error if I try to run this:
```
  File "<string>", line 1
    print('\n'.join((f'{chr(34)}{path}{chr(34)}' if ' ' in path else path) for path in sorted(str(path) for path in __import__('pathlib').Path('.').rglob('*.swift'))))
                                               ^
SyntaxError: invalid syntax
```

d-ronnqvist: How is the python snippet's output different from a simpler `find` command like this
```
find . -type f -name "*.swift" | sort
```

CONTRIBUTING.md
```diff
@@ -350,7 +350,23 @@ by running the test suite in a Docker environment that simulates Swift on Linux.
...
+## Updating Build Rules
+
+In order to build DocC as part of the toolchain distribution, a parallel CMake
+based build exists. Note that this is **not** supported for development purposes
+(you cannot execute the test suite with this build).
+
+CMake requires that the full file list is kept up-to-date. When adding or
+removing files in a given module, the `CMakeLists.txt` list must be updated to
+the file list.
+
```
d-ronnqvist, line 363: Is there anything the developer needs to do if they add a new target or a new dependency?

CONTRIBUTING.md
```diff
@@ -350,7 +350,23 @@ by running the test suite in a Docker environment that simulates Swift on Linux.
     cd swift-docc
     swift run docc
     ```
-    
+
+## Updating Build Rules
+
+In order to build DocC as part of the toolchain distribution, a parallel CMake
```
d-ronnqvist: Is there some information missing in this first sentence? DocC has been part of the Swift toolchain on both Linux and macOS for several years without this CMake build, so it feels incorrect, or at least confusing, to me to say that the CMake build is required for DocC to be part of the toolchain distribution.

CONTRIBUTING.md
```diff
@@ -350,7 +350,23 @@ by running the test suite in a Docker environment that simulates Swift on Linux.
...
+based build exists. Note that this is **not** supported for development purposes
+(you cannot execute the test suite with this build).
+
+CMake requires that the full file list is kept up-to-date. When adding or
+removing files in a given module, the `CMakeLists.txt` list must be updated to
+the file list.
+
+You can use the following 1-line script to enumerate the files in the module:
```
d-ronnqvist: In what directory is the developer expected to run this command and what are they expected to do with the output?

CONTRIBUTING.md
```diff
@@ -350,7 +350,23 @@ by running the test suite in a Docker environment that simulates Swift on Linux.
     cd swift-docc
     swift run docc
     ```
-    
+
+## Updating Build Rules
+
+In order to build DocC as part of the toolchain distribution, a parallel CMake
+based build exists. Note that this is **not** supported for development purposes
```
d-ronnqvist: Just curious: Is there a technical limitation why the CMake build doesn't run the test suite or is it only that these CMake files don't define the test targets and the steps to run them?

CMakeLists.txt
```diff
@@ -0,0 +1,42 @@
+#[[
+This source file is part of the Swift open source project
+
+Copyright © 2014 - 2025 Apple Inc. and the Swift project authors
```
d-ronnqvist, line 4: minor: Is there a particular reason to that this year range starts at 2014?

This repo didn't exist until 2021 so to me it looks a bit odd to me that this year range in this extends before that.

In other files in the repo we start counting from the year that the file was added (which would be 2024) for these files.

compnerd: I can change that. There's some reasoning that I can discuss offline if you like. Either is fine for now: we can use this year or I can change it to the rest of the project's header.

CONTRIBUTING.md
compnerd: Yes, a new target would require a new target to be defined (https://cmake.org/cmake/help/latest/command/add_library.html)

CONTRIBUTING.md
compnerd: I guess we can add explicitly for Windows if you like - if you are bootstrapping the toolchain, you cannot build with SPM on all platforms.

CONTRIBUTING.md
compnerd: The command is meant to be run in the directory that they are changing the file list in, and the output would just be the new list of files that are listed in the CMakeLists.txt for the target.

CONTRIBUTING.md
compnerd: It isn't a technical limitation, merely that I don't feel like we should be supporting that. Long term, I would like to get to the point where we can build with SPM, but that requires new functionality which is not yet possible.

Additionally, having two different mechanisms for development is not the desired state. This is more a tactical thing for creating a toolchain distribution. We have similar things for many other projects - SPM, LSP, driver, etc.

CONTRIBUTING.md
d-ronnqvist: That information and link would be good to have in the contributing documentation but I don't see it as required until we have a CI that verifies the CMake build.

CONTRIBUTING.md
d-ronnqvist: How about this so that the instructions tells the developer what they can use the output from (and vaguely hints that they would run it once per target)?

```suggestion
The 1-line script below lists all the Swift files in the current directory tree. 
You can use the script's output to replace the list of files in the CMakeLists.txt file for each target that you added files to or removed files from.
```

CONTRIBUTING.md
d-ronnqvist: I figured out that this is a Python version issue. I think my machine has updated many times to it still defaults to python 2 and I need to explicitly specify `python3` for this snippet to work.

How do you feel about spelling that out in the snippet? 

```suggestion
python3 -c "print('\n'.join((f'{chr(34)}{path}{chr(34)}' if ' ' in path else path) for path in sorted(str(path) for path in __import__('pathlib').Path('.').rglob('*.swift'))))"
```

Alternatively, a `find` command could easily quote all the paths and might be a more portable 1-liner:

```suggestion
find Sources/NAME_OF_TARGET -type f -name "*.swift" -exec echo "\"{}\"" \; | sort 
```

compnerd: The find doesn't work on all platforms. `python3` is a macOS specific thing, all other platforms use python as the interpreter. I don't think that having two spellings is good as the script can then diverge. How about leaving the interpreter bit off of it and just having the body?

compnerd:
Yeah, merging everything into a single change is not really feasible - there are too many moving pieces. We need to get the basic bits in first.

I have updated the CONTRIBUTING.md with the recommended alteration.

CONTRIBUTING.md
d-ronnqvist: > How about leaving the interpreter bit off of it and just having the body?

This sounds like a good solution to me if it mentions Python 3 somewhere to help developers debug any `SyntaxError: invalid syntax` errors.

jakepetroules (approved):

compnerd:
@swift-ci please test

d-ronnqvist:
> Yeah, merging everything into a single change is not really feasible - there are too many moving pieces. We need to get the basic bits in first.

Yes. It makes sense to land this incrementally.

d-ronnqvist (approved):

jakepetroules:
@swift-ci test

PR #844, opened 2024-02-25
Add visionOS support

yonihemi:
## Summary

Adds support for running DocC on visionOS by extending platform checks.

## Dependencies

Upgrades Swift Crypto to [v2.6.0](https://github.com/apple/swift-crypto/releases/tag/2.6.0), which removed specific os conditions.

## Testing

Run `SwiftDocCTests` and `SwiftDocCUtilitiesTests` with a visionOS simulator target.
No platform-specific tests added.
I don't think there's a public way to run the benchmark for a target other than host.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -41,8 +41,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-crypto.git",
       "state" : {
-        "revision" : "33a20e650c33f6d72d822d558333f2085effa3dc",
-        "version" : "2.5.0"
+        "revision" : "60f13f60c4d093691934dc6cfdf5f508ada1f894",
+        "version" : "2.6.0"
       }
     },
     {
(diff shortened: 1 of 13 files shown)


Kyle-Ye (commented):
LGTM in general.

But I'm not sure if we need to add visionOS support since we do not have support for watchOS/tvOS/etc.

Can you point out some use case on why we need to run swift-docc on iOS / visionOS platform since those platform does not support running executable target directly? 

> Only macOS / Linux / Windows support running cli executable.

cc @franklinsch

snprajwal:
The changes look good, we're happy to merge this. Would you please be able to update the branch and resolve the merge conflicts?

snprajwal:
This PR hasn't been updated in 2 years, and cannot be merged in its current state. If you (or anyone else) would like to continue working on it, we encourage reopening this.

In the meantime, I will close this PR to reduce noise in the list of open PRs.

PR #866, opened 2024-03-25
[6.0] Don't enforce unique filenames for documentation extensions

mayaepps:
- **Explanation:** Excludes documentation extension files from DocC's check that all documents in a catalog must have unique filenames. Previously, one of the doc extensions' content would be dropped (and a diagnostic would be emitted about a redeclared file), which can be confusing when the extensions are for separate symbols with the same name. However, since documentation extensions' filenames have no impact on the URL of those pages, there's no need to enforce unique filenames for them.
- **Scope:** Only impacts documentation extension files with the same filename.
- **Issue:** rdar://117174884
- **Risk:** Low. Excludes documentation extension files (defined as articles starting with an H1 containing link) from DocC's check that all documents in a catalog must have unique filenames.
- **Testing:** Added a unit test ensuring that warnings are no longer emitted and all documentation extension content appears in its correct place. Also tested manually with a small test project and catalog by building a docc executable using this branch, setting DOCC_EXEC to point to it, then running `swift package generate-documentation` and checking that no warnings were emitted and all documentation extension content appears in its correct place.
- **Reviewer:** @d-ronnqvist 
- **Original PR:** https://github.com/apple/swift-docc/pull/863

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -854,7 +854,11 @@ public class DocumentationContext: DocumentationContextDataProviderDelegate {
             let path = NodeURLGenerator.pathForSemantic(analyzed, source: url, bundle: bundle)
             let reference = ResolvedTopicReference(bundleIdentifier: bundle.identifier, path: path, sourceLanguage: .swift)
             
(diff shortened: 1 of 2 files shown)


mayaepps:
@swift-ci Please test

franklinsch (approved):

PR #873, opened 2024-03-30
Fix incorrect key name in ThemeSettings spec JSON

gwynne:
## Summary

The `ThemeSettings.spec.json` file specifies that the key under `theme` for the dictionary of colors is `colors`. However, when used in practice, the correct key is `color`; `colors` does not work. This corrects the spec to match the implementation.

(In truth, it seems like `colors` would be both more accurate and more consistent, but changing the implementation to match the spec would be _considerably_ wider-ranging and have significant compatibility implications.)

## Testing

N/A, documentation-only change

## Checklist

- [ ] ~~Added tests~~ N/A, documentation-only change
- [ ] ~~Ran the `./bin/test` script and it succeeded~~ N/A, documentation-only change


QuietMisdreavus (approved):
Thanks!

QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
Linux CI failure was known issue, fixed by https://github.com/apple/swift-docc/pull/875

QuietMisdreavus:
@swift-ci Please test

PR #886, opened 2024-04-15
[6.0] Auto-Capitalize first word

emilyychenn:
- **Explanation:** When writing a doc comment, it feels intuitive to not capitalise the first word of its description because of the nature of the doc comment. 
    - For example: adding a new parameter would be written as: `/// - Parameter testParam: this parameter is just a test parameter to show what this looks like in lowercase.` 
    - In this example, the sentence in doc comments is natural to write with a lowercase starting word, because the first word (Parameter) is already capitalized.  
    - This PR auto-capitalizes the first word of a new section or aside. 
    - Note that this auto-capitalization only occurs if the first word is all lowercase and contains only characters A-Z, or if the first word contains CharacterSet punctuation characters (e.g. a period, comma, hyphen, semi-colon, colon).

- **Scope**: Developer experience — auto-capitalizes the first word.

- **Issue**: rdar://122167705

- **Risk**: Low, only risk is that developers don't like this change and would prefer taking control of their own capitalization.

- **Testing**: 
    - Added unit tests to check behaviour of the auto-capitalization for a variety of different strings and a variety of languages and characters, spaces. 
    - Added end-to-end integration tests to check behaviour of the auto-capitalization for a variety of different content types (e.g. parameter sections, asides, returns sections etc).
    - Also tested manually with a small test project and catalog by building a docc executable using this branch, then building the documentation within Xcode and checking that the capitalization is applied correctly when applicable and is not applied when it's not relevant.

- **Reviewers**: @patshaughnessy @daniel-grumberg

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent+Capitalization.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent+Capitalization.swift
@@ -1,0 +1,92 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2024 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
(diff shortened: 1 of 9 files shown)


emilyychenn:
@swift-ci please test

PR #931, opened 2024-05-29
Merge automatically curated task groups

anferbui:
Bug/issue #, if applicable: rdar://61899214

## Summary

Currently, when a developer writes a symbol extension markdown file, and creates a topic section which is named the same as one of the titles which are used for automatic curation ("Initializers", "Instance Properties", "Instance Methods", "Type Aliases", etc.), the result is duplicated sections:
3cf36344-f5ab-466f-832f-1ffc8c738ac7

This PR changes this behaviour so that automatically curated section contents are merged with existing sections, if they have the same title:
c0df4dbb-3317-4bf6-a517-ccb8e48dcdff

### Implementation

in `RenderNodeTranslator.visitSymbol(_:)`, after the list of automatically curated task groups has been created, we now check whether there is an existing manually curated section with the same title. 

If it exists, we extend this section in place with the symbol references from the automatically curated task group. Otherwise, we create a new section same as before.

## Dependencies

No dependencies.

## Testing

You can use the following test package to test locally:
[TestPackage.zip](https://github.com/apple/swift-docc/files/15486030/TestPackage.zip)

Steps:
1. Generate rendered documentation for the test package using `docc`
2. View rendered documentation for `ExampleClass` (at http://localhost:8080/documentation/testpackage/exampleclass) and validate that there is only one "Initializers" section under "Topics".

More generally, this can be reproduced by:
1. Creating an extension markdown file for a package's symbol.
2. Creating a section under topics which is titled same as one of the titles which are used for automatic curation ("Initializers", "Instance Properties", "Instance Methods", "Type Aliases", etc.).
3. Render the documentation.
4. View the extended symbol's documentation, and verify that the contents from the manually curated section have been merged with the automatically curated symbols.

## Checklist

- [x] Added tests

### Verification

**Unit tests**

Added test `testAutomaticallyCuratedSymbolTopicsAreMergedWithManuallyCuratedTopics()`, which passes. This test explicitly tests the new merging functionality. I'm relying on the existing tests for testing that previous automatic curation functionality still works as expected.

Ran `./bin/test`:
```
=> Checking for unacceptable language… okay.
=> Checking license headers… okay.
=> Validating scripts in bin subdirectory… okay.
```

**Site rendering**

Verified manually by creating a test package with an extension markdown file which defines a topic section "Initializers". Before this change, the result would be two separate sections, both titled "Initializers". After this change, there is only one section named "Initializer" with the merged contents.

Input markdown:
```md
# ``ExampleClass``


This is a summary.

## Overview

This is an overview.

## Topics

### Other group

- <doc:ExampleArticle>

### Initializers

- <doc:ExampleArticle>

## See Also

- ``TestPackage``
```

Before                                  |  After
:-------------------------:|:-------------------------:
c60e266d-74e0-49a9-8305-e1846d2bc789 | c664cdff-cc81-41bc-bab4-4f8e56ff0cbe

**Documentation**

As far as I can tell, no changes to the documentation are needed.

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1555,6 +1555,8 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 // If the section has been manually curated, merge the references of both the automatic curation and the manual curation into one section (rdar://61899214).
                 if let duplicateSectionIndex = sections.firstIndex(where: { $0.title == group.title }) {
                     let originalSection = sections[duplicateSectionIndex]
+                    // Combining all references here without checking for duplicates should be safe,
+                    // because the automatic curation of topics only returns symbols that haven't already been manually curated.
                     let combinedReferences = originalSection.identifiers + newReferences.map { $0.absoluteString }
                     sections[duplicateSectionIndex] = TaskGroupRenderSection(title: originalSection.title, abstract: originalSection.abstract, discussion: originalSection.discussion, identifiers: combinedReferences)
                 } else {
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1546,16 +1546,21 @@ public struct RenderNodeTranslator: SemanticVisitor {
             
             // Children of the current symbol that have not been curated manually in a task group will all
             // be automatically curated in task groups after their symbol kind: "Properties", "Enumerations", etc.
-            let alreadyCurated = Set(sections.flatMap { $0.identifiers })
             let groups = try! AutomaticCuration.topics(for: documentationNode, withTraits: allowedTraits, context: context)
             
-            sections.append(contentsOf: groups.compactMap { group in
-                let newReferences = group.references.filter { !alreadyCurated.contains($0.absoluteString) }
-                guard !newReferences.isEmpty else { return nil }
-                
+            for group in groups {
+                let newReferences = group.references
                 contentCompiler.collectedTopicReferences.append(contentsOf: newReferences)
-                return TaskGroupRenderSection(taskGroup: (title: group.title, references: newReferences))
-            })
+
+                // If the section has been manually curated, merge the references of both the automatic curation and the manual curation into one section (rdar://61899214).
+                if let duplicateSectionIndex = sections.firstIndex(where: { $0.title == group.title }) {
+                    let originalSection = sections[duplicateSectionIndex]
+                    let combinedReferences = originalSection.identifiers + newReferences.map { $0.absoluteString }
+                    sections[duplicateSectionIndex] = TaskGroupRenderSection(title: originalSection.title, abstract: originalSection.abstract, discussion: originalSection.discussion, identifiers: combinedReferences)
+                } else {
+                    sections.append(TaskGroupRenderSection(taskGroup: (title: group.title, references: newReferences)))
+                }
+            }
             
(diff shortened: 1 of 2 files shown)


anferbui:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1546,16 +1546,21 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
-                guard !newReferences.isEmpty else { return nil }
-                
+            for group in groups {
+                let newReferences = group.references
                 contentCompiler.collectedTopicReferences.append(contentsOf: newReferences)
-                return TaskGroupRenderSection(taskGroup: (title: group.title, references: newReferences))
-            })
+
+                // If the section has been manually curated, merge the references of both the automatic curation and the manual curation into one section (rdar://61899214).
+                if let duplicateSectionIndex = sections.firstIndex(where: { $0.title == group.title }) {
```
patshaughnessy, line 1556: This exact string comparison might be too strict. For example, I tried creating a topic group like this:

```
### Type properties

- ``!=(_:_:)``
```

(lower case "p") ... and my docs ended up with redundant topic sections. Can we use a case insensitive check? And possibly even remove duplicate whitespace? The same thing happened when added two spaces between "Type" and "Properties"... to do that we might need a regex of some kind.

Might be overkill?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1546,16 +1546,21 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+                    let originalSection = sections[duplicateSectionIndex]
+                    let combinedReferences = originalSection.identifiers + newReferences.map { $0.absoluteString }
```
patshaughnessy, line 1558: I'm curious why we don't need to worry about duplicate identifiers after this `+` operation. I suppose that `AutomaticCuration.topics` always returns groups that contain only the missing identifiers. Might be worth a comment if so.

patshaughnessy (approved):
Looks great!

Left an optional suggestion about using a case-insensitive comparison when checking for duplicate topic sections.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
anferbui: Yes, `AutomaticCuration.topics` only returns symbols which haven't been curated already -- will add a comment!

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
anferbui: I thought about this but in the end decided to go with the simplest solution for the moment (no fuzzy matching). Even though I think it would be helpful for the user in some cases, I'm leaning towards no because it's hard to find a level of leniency that would work for all use cases -- e.g. if someone has a topic section "Initialisers" (s instead of z), should we also merge those? And also what you've mentioned about multiple spaces.

WDYT?

anferbui:
@swift-ci Please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
patshaughnessy: I actually had second thoughts about this too - I think going with the most transparent approach makes the most sense. If an author sees a duplicate topic section because of capitalization or spelling, then it should be obvious how to fix it, and also the tool's behavior is easier to understand.

anferbui:
@swift-ci Please test

PR #933, opened 2024-05-30
Merge alternate symbol declarations across symbol graphs

anferbui:
Bug/issue #, if applicable: rdar://128573538

## Summary

Fixes a bug with platform availability of a symbol's alternate declarations. This bug resulted in alternate symbol declarations only ever showing as having one supported platform, regardless of the symbol being declared across different symbol graphs for different platforms:
185a84ce-818c-4786-acf9-41141c18dee8

The expected behaviour is that `async` alternate declarations are supported for the same platforms as their completion handler declaration:
7ed7dad9-9638-435c-a299-06ea67ec201e

With these changes, platform names are now merged across alternate declarations from different symbol graphs.

### Implementation

The issue was caused because alternate declarations were not being merged across different symbol graphs, so that platform availability information was only stored for the first symbol graph that was processed, and further platform availability information was lost.

To fix this issue, `Symbol.mergeDeclaration(...)` now takes an extra parameter which takes in a list of alternate declarations from the symbol being merged in. This was already in place for symbol availability and the symbol's main declaration, but hadn't been added yet for alternate declarations, so information about other platforms was being dropped rather than being merged into the symbol's alternate declarations.

Additionally, we now propagate alternate declarations to the new (copied) symbol that is created when resolving references:
https://github.com/apple/swift-docc/blob/037f0bcca7b1aa75551bd9b46825287c159a86b5/Sources/SwiftDocC/Semantics/ReferenceResolver.swift#L481

## Dependencies

N/A.

## Testing

Steps:
1. Run `docc preview` the test bundle included in this PR:
```
❯ swift run docc preview Tests/SwiftDocCTests/Test\ Bundles/AlternateDeclarations.docc
```
2. Preview http://localhost:8080/documentation/alternatedeclarations/myclass/present(completion:) locally, and the platform availability should be the same for both of the symbol declarations on the page (iOS, macOS).

More generally, this can be reproduced by:
1. Creating a test package which contains an Objective C symbol with a completion handler, e.g.:
```objc
@implementation MyClass
- (void)presentWithCompletion:(void (^)(BOOL success))completion {
    NSLog(@"Hello, World!");
}
@end
```
2. Compiling the package in Swift for multiple platforms to obtain one symbol graph per platform
3. Moving the symbol graphs into a folder called DummyBundle.docc
4. Running `swift run docc preview DummyBundle.docc` to preview the documentation
5.  Verifying that the same supported platforms show for all declarations of `present(completion:)` 

## Checklist

- [x] Added tests
- [x] [n/a] Updated documentation if necessary

### Verification

**Unit tests**
Modified test `testAlternateDeclarations()` to use 2 symbol graphs, one for macOS and one for iOS. The test now expects that the alternate declaration should be available in both platforms.

Ran `/bin/test`:
```
Build complete! (22.28s)
[1682/1682] Testing SwiftDocCUtilitiesTests.DirectoryMonitorTests/testMonitorDoesNotTriggerUpdates
=> Checking for unacceptable language… okay.
=> Checking license headers… okay.
=> Validating scripts in bin subdirectory… okay.
```

**Site rendering**
Verified that the rendered site renders the list of platforms for alternate declarations as expected.

Before                                  |  After
:-------------------------:|:-------------------------:
185a84ce-818c-4786-acf9-41141c18dee8 | 7ed7dad9-9638-435c-a299-06ea67ec201e

**Documentation**
Did not need to update the documentation as this is a bug fix, and this was already the intended behaviour of the feature.


anferbui:
@swift-ci Please test

QuietMisdreavus (approved):
This looks great, thanks!

PR #992, opened 2024-07-23
memoize the common fragments calculation for diffing overloads

QuietMisdreavus:
Bug/issue #, if applicable: None

## Summary

This PR refactors the overload declaration diffing code to memoize the results so that they don't need to be recalculated. The original diffing code recalculates the "common fragments" every time it needs to render a page. However, this wastes work for overloads in the same group, since each symbol's page needs to be calculated individually. By saving those results, we can load the common fragments without having to perform an expensive calculation.

This benchmark was run on a large framework with many overloaded methods:

```
│ Duration for 'bundle-registration'       │ no change¹,²    │ 9.523 sec            │ 9.475 sec            │
│ Duration for 'convert-total-time'        │ -2.796%³        │ 14.724 sec           │ 14.312 sec           │
│ Duration for 'documentation-processing'  │ -7.142%⁴        │ 5.091 sec            │ 4.727 sec            │
│ Duration for 'finalize-navigation-index' │ no change⁵      │ 0.04 sec             │ 0.04 sec             │
│ Peak memory footprint                    │ no change⁶      │ 698.7 MB             │ 704 MB               │
│ Data subdirectory size                   │ no change⁷      │ 129.8 MB             │ 129.8 MB             │
│ Index subdirectory size                  │ no change⁸      │ 1.1 MB               │ 1.1 MB               │
│ Total DocC archive size                  │ no change⁹      │ 232.1 MB             │ 232.1 MB             │
│ Topic Anchor Checksum                    │ no change       │ 53074b40863239b7a00d │ 53074b40863239b7a00d │
│ Topic Graph Checksum                     │ no change       │ f4af046ddef9eab49ba8 │ f4af046ddef9eab49ba8 │
```

## Dependencies

None

## Testing

Ensure that no functionality has changed.

## Checklist

- [ n/a ] Added tests


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
@swift-ci Please test Linux

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
```diff
@@ -11,8 +11,57 @@
...
 
+typealias OverloadDeclaration = (
+    declaration: [SymbolGraph.Symbol.DeclarationFragments.Fragment],
+    reference: ResolvedTopicReference
+)
+
 /// Translates a symbol's declaration into a render node's Declarations section.
 struct DeclarationsSectionTranslator: RenderSectionTranslator {
+    /// A mapping from a symbol reference to the "common fragments" in its overload group.
+    static let commonFragmentsMap: Synchronized<[ResolvedTopicReference: [SymbolGraph.Symbol.DeclarationFragments.Fragment]]> = .init([:])
```
patshaughnessy, line 22: Why define this as static? It will work fine, I think, but would fail someday if we ever compile more than one documentation project in the same DocC process, because we would have values left over from previous runs. As written, we assume the map will only ever be used once. Might be safer to just make this a normal attribute of `DeclarationsSectionTranslator`.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
```diff
@@ -11,8 +11,57 @@
...
+
+    /// Set the common fragments for the given symbol references.
+    func setCommonFragments(
+        references: [ResolvedTopicReference],
+        fragments: [SymbolGraph.Symbol.DeclarationFragments.Fragment])
+    {
+        Self.commonFragmentsMap.sync({ map in
+            for reference in references {
```
patshaughnessy, line 30: Why bother saving the fragments for all of the overload declarations? Reading this, I'm surprised that `setCommonFragments` takes an array of references, but `commonFragments` only queries one.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
```diff
@@ -11,8 +11,57 @@
...
+    ) -> [SymbolGraph.Symbol.DeclarationFragments.Fragment]? {
+        Self.commonFragmentsMap.sync({ $0[reference] })
+    }
+
+    /// Fetch the common fragments for the given references, or compute it if necessary.
+    func commonFragments(
+        for mainDeclaration: OverloadDeclaration,
+        overloadDeclarations: [OverloadDeclaration]
+    ) -> [SymbolGraph.Symbol.DeclarationFragments.Fragment] {
+        if let fragments = commonFragments(for: mainDeclaration.reference) {
```
patshaughnessy, line 48: See the previous comment. If you only ever query the fragments for the main declaration here, then why save the same fragments for each of the overloads in the call to `setCommonFragments` below?

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
```diff
@@ -162,11 +206,12 @@ struct DeclarationsSectionTranslator: RenderSectionTranslator {
                     let processedOverloadDeclarations = overloadDeclarations.map({
                         OverloadDeclaration($0.declaration.flatMap(preProcessFragment(_:)), $0.reference)
                     })
-                    let preProcessedDeclarations = [mainDeclaration] + processedOverloadDeclarations.map(\.declaration)
 
                     // Collect the "common fragments" so we can highlight the ones that are different
                     // in each declaration
-                    let commonFragments = longestCommonSubsequence(preProcessedDeclarations)
+                    let commonFragments = commonFragments(
```
patshaughnessy, line 212: Again, it appears you only ever need the common fragments for the main declaration, and not for the individual overloads.

patshaughnessy (commented):
Looks like a nice speedup 👍   - just some questions before I approve this.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
QuietMisdreavus: The entire trick being relied upon here is that the list of common fragments is going to be the same for a group of overloaded declarations. So the first one in a group that gets rendered can save the calculation for the whole group. So when we check for a cached result, we only need to check for one of the declarations.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
QuietMisdreavus: [`DeclarationsSectionTranslator` instances are not shared between different symbols' renderings.](https://github.com/swiftlang/swift-docc/blob/2e3e73a80a6e0d460e9d6d323020757a1e0dfa12/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift#L1356) If i made this an instance property, the saved results would not be available the next time around.

patshaughnessy: I see, ok makes sense.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
patshaughnessy: So the `mainDeclaration` will change depending on when we call `commonFragments` below? Thanks for the explanation.

patshaughnessy (approved):

QuietMisdreavus:
@swift-ci Please test

Issue #970, opened 2024-06-28
Remove duplicate semantic versioning types


anferbui:
### Description

In the codebase, we have four types which we use to represent a semantic version number:
- [SemanticVersion](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Model/Rendering/SemanticVersion.swift#L14) which looks like a copy of  SymbolKit.SymbolGraph.SemanticVersion
- [Version](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Utility/Version.swift#L12)
- [VersionTriplet](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Infrastructure/External%20Data/ExternalMetadata.swift#L14)
- [Platform.Version](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Indexing/Navigator/AvailabilityIndex%2BExt.swift#L209)

These all seem to be fulfilling a similar purpose. We should converge on using only one type throughout the DocC codebase.

### Checklist

- [X] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

Only one Swift type is used to represent a semantic version number in DocC. All other types are deprecated/removed.

### Actual behavior

There are 4 types used to represent a semantic version number in DocC.

### Swift-DocC Version Information

f019ab8e2e0370d8f95921d4b8f1f7309c4eedbc

anferbui:
Per discussion in https://github.com/apple/swift-docc/pull/959, @d-ronnqvist and me think we should try to move towards converging on [SemanticVersion](https://github.com/apple/swift-docc/blob/main/Sources/SwiftDocC/Model/Rendering/SemanticVersion.swift#L14) because it's the type that is used the most in the repo.

PR #994, opened 2024-07-24
Beta tag heuristic should use >= when comparing platform version information

patshaughnessy:
rdar://132024975

## Summary

Today DocC Render displays a beta badge for symbols that were introduced in the current version of a platform, if the platform's availability was set as beta in the DocC compiler command line invocation.

This change updates this heuristic to consider symbols that were introduced in the current - _or future_ - version of a beta platform. This gives DocC a bit more flexibility to handle symbol data that might describe future versions of a platform, and to handle platforms that might have two versions in beta at the same time.

## Dependencies

None.

## Testing

Steps:

1. Build some API reference documentation containing at least one symbol with availability information. For example, a symbol might have this availability:

```
 "availability": [
    {
      "domain": "myPlatform",
      "introduced": {
        "major": 3,
        "minor": 0,
        "patch": 0
      }
    }
  ]
```

2. Declare this platform ("myPlatform") is beta using a DocC command line argument: `--platform name=myPlatform,version=3.0.0,beta=true`
3. View the symbol’s documentation page, and validate DocC marks it as beta; for example: `myPlatform 3.0+ Beta`
4. Now change the CLI argument and pass in an earlier version of the beta platform, for example: `--platform name=myPlatform,version=1.0.0,beta=true`
5. Recompile the documentation, and validate the symbol is still shown as beta: `myPlatform 3.0+ Beta`, even though the symbol was introduced in a later version of the beta platform.

## Checklist

- [ x] Added tests


patshaughnessy:
@swift-ci Please test

Tests/SwiftDocCTests/Infrastructure/VersionTripletTests.swift
```diff
@@ -12,6 +12,7 @@ import Foundation
 
 import XCTest
 @testable import SwiftDocC
+@testable import SymbolKit
```
QuietMisdreavus, line 15: Does this import need to be `@testable`? Is the `SemanticVersion` initializer not public?

QuietMisdreavus (approved):
This looks great, thanks!

Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
```diff
@@ -230,7 +230,7 @@ public class DocumentationContentRenderer {
             }
 
             // Verify that the current platform is in beta and the version number matches the introduced platform version.
-            guard current.beta && SemanticVersion(introduced).isEqualToVersionTriplet(current.version) else {
+            guard current.beta && VersionTriplet(introduced) >= current.version else {
```
anferbui, line 233: Minor comment: I was trying to work towards only using `SemanticVersion` as the type used internally by DocC for semantic versioning, if we could head in that direction that would be helpful (context: https://github.com/swiftlang/swift-docc/issues/970)

sofiaromorales:
Minor: Specify the beta behaviour in the `platform` CLI help line (https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocCUtilities/ArgumentParsing/Subcommands/Convert.swift#L148), to something like this:

```
...
Optionally, the platform information can include a 'beta={true|false}' component.
If no beta information is provided, the platform is considered not in beta.
If the platform is set to beta, any symbol introduced in a version equal to or greater than the specified semantic version will be marked as beta.
```

Sources/SwiftDocC/Infrastructure/External Data/ExternalMetadata.swift
```diff
@@ -41,6 +42,18 @@ public struct VersionTriplet: Equatable, Comparable {
         self.minor = minor
         self.patch = patch
     }
+
+    public init(_ semanticVersion: SemanticVersion) {
+        self.major = semanticVersion.major
+        self.minor = semanticVersion.minor
+        self.patch = semanticVersion.patch
+    }
```
sofiaromorales, line 50: What's the conceptual difference between a Version Triplet and a SemanticVersion? Do we need both?

Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
sofiaromorales: +1 to this. Can we use Semantic Version here?

sofiaromorales (approved):

Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
patshaughnessy: I changed this code to use `VersionTriplet` since that conforms to `Comparable` making functions such as `isEqualToVersionTriplet` unnecessary. If we're moving towards using `SemanticVersion` then I can add `Comparable` conformance to that and refactor this code.

Tests/SwiftDocCTests/Infrastructure/VersionTripletTests.swift
patshaughnessy: Ah - I misunderstood the meaning of `@testable`. I thought that it was just to indicate we were using the library for testing purposes for informational purposes. But if we don't need that, I'll remove it. thanks!

sofiaromorales:
@swift-ci please test

patshaughnessy:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/SemanticVersion.swift
```diff
@@ -51,6 +53,31 @@ public struct SemanticVersion: Codable, Equatable, CustomStringConvertible {
         self.buildMetadata = try container.decodeIfPresent(String.self, forKey: .buildMetadata)
     }
 
+    /// Create a semantic version from a version triplet
+    /// TODO: https://github.com/swiftlang/swift-docc/issues/970
+    /// Migrate all the code to use semantic versions, and not version triplets.
```
sofiaromorales, line 58:
```suggestion
    // TODO: https://github.com/swiftlang/swift-docc/issues/970
    // Migrate all the code to use semantic versions, and not version triplets.
    /// Create a semantic version from a version triplet.
```

Tests/SwiftDocCTests/Test Bundles/TestBundle.docc/.swp
sofiaromorales, line 1: Do we need this?

sofiaromorales (approved):

patshaughnessy:
@swift-ci please test

PR #995, opened 2024-07-25
Resolve external links in all language variants of parameter sections

patshaughnessy:
rdar://132265198

## Summary

Today if API reference documentation for a symbol contains two different language variants, for example Swift and Objective-C, and these two versions of the symbol have a different list of parameters, DocC will not always find and resolve external links in documentation about these parameters.

This changes the code in `ExternalReferenceWalker` to process all variants of parameters sections, and not just the first parameter section.

## Dependencies

None.

## Testing

Find or write documentation for a symbol that exists in both Swift and Objective-C API, and that has a different list of parameters for each language. In practice, this happens frequently since Objective-C functions often return an error value with an `NSError * *` type, while the Swift function throws an exception.

For example, here's some documentation about a fictional symbol which has different parameters in Swift and Objective-C:

```
# ``SomeSymbol``

This is a fictional symbol.

- Parameters:
  - someParameter: This parameter is the same in the Swift and Objective-C API versions. 
  - someError: This error parameter contains an external link to <doc://com.example/some/other/page> and only exists for the Objective-C version of the symbol, and .

```

Note to reproduce this behavior, you will need to provide two symbol graphs to DocC, one containing the Swift version of this symbol and one containing the Objective-C versions.

Test that DocC produces documentation for Objective-C that display a valid external link to  "some/other/page" for example.

## Checklist

- [x] Added tests


patshaughnessy:
@swift-ci please test

patshaughnessy:
@swift-ci please test

anferbui (approved):
Awesome!

patshaughnessy:
@swift-ci please test

PR #1050, opened 2024-10-09
[WIP] Feature flag to allow default availability items without version. 

sofiaromorales:
Bug/issue #, if applicable: 

## Summary

## Dependencies

## Testing

applicable._

Steps:

## Checklist

- [ ] Added tests


Sources/SwiftDocC/Infrastructure/Workspace/FeatureFlags+Info.swift
```diff
@@ -31,9 +31,18 @@ extension DocumentationBundle.Info {
         ///
         /// This feature flag corresponds to ``FeatureFlags/isExperimentalOverloadedSymbolPresentationEnabled``.
         public var experimentalOverloadedSymbolPresentation: Bool?
+        
+        /// Whether or not experimental support for allowing availability items without version.
+        ///
+        /// This feature flag corresponds to ``FeatureFlags/isExperimentalAvailabilityItemsWithoutVersionEnabled``.
+        public var experimentalAvailabilityItemsWithoutVersion: Bool?
```
anferbui, line 38: There was some other code which was also changed in the original commit and is not under feature flag, how high do you think the risk of potential issues is here?
https://github.com/swiftlang/swift-docc/blob/79bb1d043960c3f4a884e0b9e085fae34777ca31/Sources/SwiftDocC/Infrastructure/Workspace/DefaultAvailability.swift#L104-L112

anferbui (approved):
It's not ideal that the feature flag check is in multiple locations (it'd be nice if the new logic was isolated to a separate area of the code), but I don't think we've got a way around this given the current design for availability.

LGTM!

Sources/SwiftDocC/Infrastructure/Workspace/FeatureFlags+Info.swift
sofiaromorales: Hiding this change under the feature flag would require to do something similar to this:
```swift
let introducedVersion = FeatureFlags.current.isExperimentalAvailabilityItemsWithoutVersionEnabled
            ? try values.decodeIfPresent(String.self, forKey: .platformVersion)
            : try values.decode(String.self, forKey: .platformVersion)
```

This is not feasible because the decoding of the Info plist default availability keys happens all in the same process as the decoding of the Info.plist feature flags. To register the values of `FeatureFlags.current` we need a bundle which is generated in a later step, meaning that by the time we decode the default availability 

```
FeatureFlags.current.isExperimentalAvailabilityItemsWithoutVersionEnabled
```
will be false no matter the actual value defined in the  feature flags Info.plist.

PR #1121, opened 2024-12-06
Remove beta tag from unavailable symbols.

sofiaromorales:
Bug/issue #, if applicable:  140415995

## Summary

Symbols that are unconditionally unavailable in all the platforms should not be tagged as beta.Before this fix the `isBeta` logic was returning true for symbols that were unavailable in all the platforms. Now is filtering out these platforms before assesing if it's beta or not.

## Dependencies

## Testing

See unit tests.

## Checklist

- [X] Added tests


d-ronnqvist (approved):
LGTM

sofiaromorales:
@swift-ci please test

PR #1122, opened 2024-12-09
[6.1] Remove beta tag from unavailable symbols. (#1121)

sofiaromorales:
* **Explanation:**  Symbols that are unconditionally unavailable in all the platforms should not be tagged as beta. Before this fix the `isBeta` logic was returning true for symbols that were unavailable in all the platforms. Now is filtering out these platforms before assessing if it's beta or not.
* **Scope:** Symbol availability information.
* **Issue:** rdar://140415995
* **Risk:** Low
* **Testing:** New tests added to verify that a symbol is not marked in beta if all the platforms are unavailable.
* **Reviewer:** @d-ronnqvist 
* **Original PR:** https://github.com/swiftlang/swift-docc/pull/1121


sofiaromorales:
@swift-ci please test

PR #1145, opened 2025-01-17
Revert "Resolve all current code warnings (#1143)"

tshortli:
The changes from https://github.com/swiftlang/swift-docc/pull/1143 are blocking PRs from being merged in the Swift compiler repo because the NIO dependency is not new enough ([example](https://ci.swift.org/job/swift-PR-Linux-smoke-test/16990/console)). Updating the version of NIO that this project depends on will need to be done in conjunction with an update to update-checkout.sh in the https://github.com/swiftlang/swift repo.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -95,8 +95,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-nio.git",
       "state" : {
-        "revision" : "ba72f31e11275fc5bf060c966cf6c1f36842a291",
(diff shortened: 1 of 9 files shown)


tshortli:
@swift-ci please test

QuietMisdreavus (approved):
Approving this blanket revert so the compiler is unblocked. A targeted un-revert can be made to add the non-NIO warning fixes, or we can go ahead and update the compiler's dependencies first.

PR #1213, opened 2025-05-07
use found disambiguation to calculate diagnostic range if present

QuietMisdreavus:
Bug/issue #, if applicable: rdar://150207195

## Summary

The link resolver emits diagnostics for symbol links that require a disambiguation suffix (or a more precise suffix) to resolve to a single symbol. The diagnostic also includes fix-it solutions suggesting the available suffixes for the conflicting symbols. However, the logic for reporting the ranges of these possible replacements has a flaw when reporting on a curated symbol link: it always uses the full name of the symbol as the base instead of the written symbol or just the suffix.

This PR updates the diagnostic logic to handle the case of a found disambiguation suffix separately when reporting the diagnostic solutions.

## Dependencies

None

## Testing

The functionality is difficult to test outside of a unit test environment, as it requires non-overloadable symbols to have conflicting names. This is primarily only going to happen with errors in symbol graph generation that cause separate sets of symbols to exist. Nonetheless, the unit test captures the behavior.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+Error.swift
+++ b/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+Error.swift
@@ -223,13 +223,27 @@ extension PathHierarchy.Error {
             
         case .lookupCollision(partialResult: let partialResult, remaining: let remaining, collisions: let collisions):
             let nextPathComponent = remaining.first!
-            let (pathPrefix, _, solutions) = makeCollisionSolutions(from: collisions, nextPathComponent: nextPathComponent, partialResultPrefix: partialResult.pathPrefix)
-            
+            let (pathPrefix, foundDisambiguation, solutions) = makeCollisionSolutions(
+                from: collisions,
+                nextPathComponent: nextPathComponent,
+                partialResultPrefix: partialResult.pathPrefix)
+
+            let rangeAdjustment: SourceRange
+            if !foundDisambiguation.isEmpty {
+                rangeAdjustment = .makeRelativeRange(
+                    startColumn: pathPrefix.count - foundDisambiguation.count,
+                    length: foundDisambiguation.count)
+            } else {
+                rangeAdjustment = .makeRelativeRange(
+                    startColumn: pathPrefix.count - nextPathComponent.full.count,
+                    length: nextPathComponent.full.count)
+            }
+
             return TopicReferenceResolutionErrorInfo("""
                 \(nextPathComponent.full.singleQuoted) is ambiguous at \(partialResult.node.pathWithoutDisambiguation().singleQuoted)
                 """,
                 solutions: solutions,
-                rangeAdjustment: .makeRelativeRange(startColumn: pathPrefix.count - nextPathComponent.full.count, length: nextPathComponent.full.count)
+                rangeAdjustment: rangeAdjustment
             )
         }
     }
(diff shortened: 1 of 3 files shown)


QuietMisdreavus:
@swift-ci Please test

Tests/SwiftDocCTests/Infrastructure/DocumentationContext/DocumentationContextTests.swift
```diff
@@ -5266,7 +5266,58 @@ let expected = """
             }
         }
     }
-    
+
+    func testContextDiagnosesInsufficientDisambiguationWithCorrectRange() throws {
+        // This test deliberately does not turn on the overloads feature
+        // to ensure the symbol link below does not accidentally resolve correctly.
+        let symbolKindID = SymbolGraph.Symbol.KindIdentifier.allCases.first { !$0.isOverloadableKind }!
```
d-ronnqvist, line 5273: Is this meant to test more than one kind identifier? If not, could the test explicitly specify a know value so that it's clearer which kind identifier is used in the test.

QuietMisdreavus: It doesn't need to use more than one kind identifier, but i didn't want to depend on a specific symbol kind always being able to be non-overloadable. I can adjust it to use the one that was used when i ran the test (`snippet`) if that would make it clearer, but to me the fact that i pulled an arbitrary non-overloadable symbol kind was the important part.

Tests/SwiftDocCTests/Infrastructure/DocumentationContext/DocumentationContextTests.swift
```diff
@@ -5266,7 +5266,58 @@ let expected = """
...
+        let (_, context) = try loadBundle(catalog: catalog)
+
+        let problems = context.problems.sorted(by: \.diagnostic.summary)
+        XCTAssertEqual(problems.count, 1)
+
+        let problem = try XCTUnwrap(problems.first)
+
+        XCTAssertEqual(problem.diagnostic.summary, "'SymbolName-\(symbolKindID.identifier)' is ambiguous at '/ModuleName'")
+
+        for solution in problem.possibleSolutions {
```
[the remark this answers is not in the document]
QuietMisdreavus: Ah, yeah, that would be prudent to do before iterating over them. 😅 I can add that.

Tests/SwiftDocCTests/Infrastructure/DocumentationContext/DocumentationContextTests.swift
d-ronnqvist: Would it be worth iterating over the kind identifiers (in case there's something odd with one of them)?

QuietMisdreavus: That's possible. The code i copied to set up the symbols was taken from another test that iterated over all of them, so i could easily take that as well.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+Error.swift
```diff
@@ -223,13 +223,27 @@ extension PathHierarchy.Error {
             
         case .lookupCollision(partialResult: let partialResult, remaining: let remaining, collisions: let collisions):
             let nextPathComponent = remaining.first!
-            let (pathPrefix, _, solutions) = makeCollisionSolutions(from: collisions, nextPathComponent: nextPathComponent, partialResultPrefix: partialResult.pathPrefix)
-            
+            let (pathPrefix, foundDisambiguation, solutions) = makeCollisionSolutions(
```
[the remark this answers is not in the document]
d-ronnqvist: Either way works. If it's easy to investigate and fix it might make sense to do it while all of this is fresh in the mind.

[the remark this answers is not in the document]
QuietMisdreavus: I added tests for those cases to see if they would also have problems, and it looks like they're fine. 🤷‍♀️ I'll leave the tests in for posterity, but i didn't need to change anything for the cases i tested.

QuietMisdreavus:
@swift-ci Please test

PR #1221, opened 2025-05-16
Sort equal weighted "Mentioned In" articles alphabetically

mportiz08:
Bug/issue #, if applicable: 150870055

## Summary

This improves the stability of the sorting for the "Mentioned In" list of articles in the case where any pair of articles have mentioned a given symbol the same number of times.

This list is already sorted by weight where an article that mentions a symbol more times will appear closer to the top of the list. However in the scenario where two articles have the same given weight, the order is not deterministic due to the fact that the backing data structure is a Swift Dictionary where the items are not enumerated in a deterministic order. This addresses that by falling back to alphabetical order for articles with the same number of mentions for a given symbol.

## Testing

Steps:
1. Configure a DocC catalog with a single symbol and 2 articles which both have one reference to the symbol.
2. Use this branch to verify that the "Mentioned In" list for the symbol always includes the articles sorted alphabetically.
3. Compare the behavior with the main branch where occasionally the order of the list may differ between compilations.

## Checklist

- [X] Added tests

### DIFF
--- a/Sources/SwiftDocC/Semantics/Article/ArticleSymbolMentions.swift
+++ b/Sources/SwiftDocC/Semantics/Article/ArticleSymbolMentions.swift
@@ -32,7 +32,14 @@ struct ArticleSymbolMentions {
         // Mentions are sorted on demand based on the number of mentions.
         // This could change in the future.
         return mentions[symbol, default: [:]].sorted {
-            $0.value > $1.value
+            // If a pair of articles have the same number of mentions, sort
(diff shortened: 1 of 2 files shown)


bitjammer (approved):
Looks good, thanks!

mportiz08:
@swift-ci test

d-ronnqvist (approved):

mportiz08:
@swift-ci test

PR #1254, opened 2025-07-23
sort platform-specific declarations by their platforms

QuietMisdreavus:
Bug/issue #, if applicable: rdar://155966155

## Summary

When a symbol's declaration is different depending on what platform it's used from, DocC will display all the variations of the declaration on the symbol page. However, the order of these declarations follows a kind of "first come, first served" approach where which platform's declaration appears first can differ depending on which symbol graph was loaded first. This can cause documentation to nondeterministically look different between builds of the same content.

This PR addresses this nondeterminism by sorting the declarations in the same order as the platforms would appear if they were in the same record (i.e. lexicographically by the platform name).

## Dependencies

None

## Testing

As this is a nondeterminism bug, the issue at hand is the way that data may be ordered differently from build to build. Therefore, any test will need to be run many times to ensure that the ordering of declarations is as expected. The unit test added in this PR captures the nondeterminism behavior *mostly* well; i have run into situations where running a test actually passed prior to the code change, but i ran into far more situations where the test failed due to the order being different, especially with a repetition count around 100.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
@@ -187,13 +187,15 @@ struct DeclarationsSectionTranslator: RenderSectionTranslator {
                 return declarations
             }
 
-            func sortPlatformNames(_ platforms: [PlatformName?]) -> [PlatformName?] {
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

patshaughnessy (approved):
Looks good 👍

QuietMisdreavus:
@swift-ci Please test

PR #1274, opened 2025-08-07
Generate concurrency-safe code in test bundle

snprajwal:
## Summary

With the new concurrency features in Swift 6, there are also a set of checks that ensure that the code is safe to run concurrently. Public static mutable variables are considered unsafe unless they are specifically of a type that conforms to `Sendable`. This patch changes the test bundle generation to use `let` instead of `var` in the case of `public static` properties.

## Dependencies

N/A

## Testing

`cd bin/make-test-bundle && swift run make-test-bundle --output . --sizeFactor 5`

## Checklist

- [x] Added tests

## Bug/issue

rdar://157651443


snprajwal:
@swift-ci please test

PR #1304, opened 2025-09-26
Update dependant packages to match 6.2.1 branch

kcieplak:
Update the dependant packages to match 6.2.1, so parent dependencies like sourcekit-lsp will not have version errors.


kcieplak:
@swift-ci test

d-ronnqvist (approved):

PR #1356, opened 2025-11-17
build: adjust for file movement, new targets

compnerd:
The SwiftDocCUtilities directory was renamed to CommandLine, and then renamed again. When The second rename occurred, the file was not restored.

Add new targets, update the build.


compnerd:
@swift-ci please test

PR #1360, opened 2025-11-19
Update RenderIndex.Node's initializer to accept an optional type

mayaepps:
Bug/issue #, if applicable: 

## Summary

If a caller needs to initialize a `RenderIndex.Node` object with an optional `type` value, they cannot do so. Since `RenderIndex.Node`’s `type` property is optional, the public initializer should accept an optional.

## Dependencies

N/A

## Testing

Create a `RenderIndex.Node` object, and confirm that you can provide a `nil` value to the initializer without compilation errors.

## Checklist

- [n/a] Added tests


patshaughnessy:
@swift-ci please test

d-ronnqvist (approved):

PR #1367, opened 2025-12-01
Rendernode spec fixes

heckj:
replaces #1364  - I caused some havoc with my own local branches and had to reset this.

## Summary

Updates the OpenAPI spec declarations to match what DocC actually generates.

## Dependencies

none

## Testing

manually: Using a vendored version of these files, along with Swift OpenAPI generator, to create swift types and decode from two sample DocC repositories to verify the generated types.

- clone https://github.com/heckj/DocCArchive
- copy spec file into `Sources/DocCArchive/Vendored` directory of that project
- regenerate the types using ./Scripts/regenerate-types.sh
- swift test

## Checklist


heckj:
I've got something whacky going on with the branch, I'm resetting this from scratch...

PR #1397, opened 2025-12-15
Read extended module from SymbolGraph.

dylansturg:
## Summary

This is a companion change to the change in the Swift compiler to [emit the "extended" module](https://github.com/swiftlang/swift/pull/86057) as an additional field in the symbol graph. Reading the base and extended modules from the file name is not feasible for module's with long names. A new compiler option was added in [Swift #83782](https://github.com/swiftlang/swift/pull/83782) to allow emitting symbol graph file's without both module names in the file name. The shortened format isn't yet supported by swift-docc though.

This also relies on a corresponding change [to swift-docc-symbolkit](https://github.com/swiftlang/swift-docc-symbolkit/pull/107) where `GraphCollector.moduleNameFor` is updated.

The change here consolidates some `moduleNameFor` functionality where swift-docc was implementing similar behavior as exists in swift-docc-symbolkit, while already depending on symbolkit so it seems prudent to just share the implementation of `moduleNameFor`.

https://github.com/swiftlang/swift-docc/issues/1399

### Backwards Compatibility

swift-docc must continue to function with symbolgraph data generated by previous versions of the swift toolchain (before https://github.com/swiftlang/swift/pull/86057). This change to swift-docc delegates computing the module name for grouping symbols from swift-docc to existing functionality inside swift-docc-symbolkit, which is already a dependency of the former. The `moduleNameFor` [implementation in swift-docc-symbolkit today](https://github.com/swiftlang/swift-docc-symbolkit/blob/6c96b743cb993ad9af4d2cd1d95165fd28475bed/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift#L121-L180) detects the extended module name by extracting it from a `ModuleNameA@ModuleNameB` style file name.

In the related change, to [swift-docc-symbolkit](https://github.com/swiftlang/swift-docc-symbolkit/pull/107), the functionality that is being used from swift-docc-symbolkit is also updated in a backwards compatible way. `GraphCollector.moduleNameFor` continues to use the existing implementation if the newly added `extended` field is not present in the symbolgraph.

## Dependencies

- https://github.com/swiftlang/swift/pull/86057
- https://github.com/swiftlang/swift-docc-symbolkit/pull/107

## Testing

applicable._

Steps:

## Checklist

- [ ] Added tests

No new tests added because all behavior is expected to be identical and existing tests are passing.

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
+++ b/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
@@ -85,8 +85,8 @@ struct PathHierarchy {
             let moduleNode: Node
             
             if !loader.hasPrimaryURL(moduleName: moduleName) {
-                guard let moduleName = SymbolGraphLoader.moduleNameFor(url),
-                      let existingModuleNode = roots[moduleName]
+                let (moduleName, _) = GraphCollector.moduleNameFor(graph, at: url)
+                guard let existingModuleNode = roots[moduleName]
                 else { continue }
                 moduleNode = existingModuleNode
             } else if let existingModuleNode = roots[moduleName] {
--- a/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
+++ b/Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
@@ -314,48 +314,12 @@ struct SymbolGraphLoader {
             symbolGraph.symbols = symbolsWithFilledIntroducedVersions
         }
     }
-    
-    /// Returns the module name, if any, in the file name of a given symbol-graph URL.
-    ///
-    /// Returns "Combine", if it's a main symbol-graph file, such as "Combine.symbols.json".
-    /// Returns "Swift", if it's an extension file such as, "Combine@Swift.symbols.json".
-    /// - parameter url: A URL to a symbol graph file.
-    /// - returns: A module name, or `nil` if the file name cannot be parsed.
-    static func moduleNameFor(_ url: URL) -> String? {
-        let fileName = url.lastPathComponent.components(separatedBy: ".symbols.json")[0]
 
-        let fileNameComponents = fileName.components(separatedBy: "@")
-        if fileNameComponents.count > 2 {
-            // Two "@"s found in the name - it's a cross import symbol graph:
-            // "Framework1@Framework2@_Framework1_Framework2.symbols.json"
-            return fileNameComponents[0]
-        }
-        
-        return fileName.split(separator: "@", maxSplits: 1).last.map({ String($0) })
-    }
-    
     /// Returns the module name of a symbol graph based on the JSON data and file name.
     ///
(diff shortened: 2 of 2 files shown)


d-ronnqvist (changes requested):
If I understand correctly, this _requires_ new symbol graph information (the changes to the Swift compiler and SymbolKit) to identify the extended module (because the current code to do is being deleted). 

If that's correct, my understanding is that DocC wouldn't be able to correctly identify the extended module for a symbol graph created by any _current_ version of the Swift compiler, or any symbol graphs from Clang or any other source of symbol graph files. 
Completely breaking backwards compatibility and breaking compatibility with other sources of symbol graph files would be a deal breaker for this (assuming that I've understood the consequences correctly).

dylansturg:
> If I understand correctly, this _requires_ new symbol graph information (the changes to the Swift compiler and SymbolKit) to identify the extended module (because the current code to do is being deleted).
> 
> If that's correct, my understanding is that DocC wouldn't be able to correctly identify the extended module for a symbol graph created by any _current_ version of the Swift compiler, or any symbol graphs from Clang or any other source of symbol graph files. Completely breaking backwards compatibility and breaking compatibility with other sources of symbol graph files would be a deal breaker for this (assuming that I've understood the consequences correctly).

I agree that backwards compatibility important, and breaking swift-docc for the current compiler version would be untenable. But I don't think this change, nor any of the companion changes, breaks backwards compatibility.

The change to swift-docc is aimed at reducing duplication, since both swift-docc and swift-docc-symbolkit have a version of `moduleNameFor` (for a symbol graph) that does the same thing and is used for similar purposes. So in this PR, I'm suggesting we centralize on 1 implementation of that in swift-docc-symbolkit.

For the companion changes, in swift-docc-symbolkit, I'm proposing a change that is also backwards compatible. When the `extended` field is set, which requires a new compiler version, `GraphCollector.moduleNameFor` uses that field. But it also retains the existing implementation to be used when `extended` is not set. The behavior is unchanged for all existing symbol graph data, and I ran the existing tests in swift-docc-symbolkit to verify that. I still need to add some new tests for the updated symbol graph format of course.

As far as I can tell, each of these changes in isolation and all changes together maintain backwards compatibility. Could you elaborate any on why you believe this is not backwards compatible? Thanks!

dylansturg:
> If I understand correctly, this _requires_ new symbol graph information (the changes to the Swift compiler and SymbolKit) to identify the extended module (because the current code to do is being deleted).

It is true that this PR is deleting code for `moduleNameFor` in swift-docc, but it's delegating the calls to swift-docc-symbolkit. There is an [implementation of moduleNameFor in symbolkit today](https://github.com/swiftlang/swift-docc-symbolkit/blob/6c96b743cb993ad9af4d2cd1d95165fd28475bed/Sources/SymbolKit/UnifiedSymbolGraph/GraphCollector.swift#L121-L180), which appears to be virtually identical to the implementation here in swift-docc.

snprajwal:
@dylansturg the changes in this PR were added into #1286, which has been merged. I'll close this PR since there are no new changes here. If you'd like to make more changes to support the changes in Swift and SymbolKit, please feel free to open a new PR. Thanks!

PR #1408, opened 2025-12-18
Include availability for fallback platforms in articles

snprajwal:
Bug/issue #, if applicable: rdar://158230351

## Summary

When DocC processes the availability information for symbol graphs, it
identifies if fallback platforms are available, and if so, injects
platforms that inherit their availability from the fallback platforms.
E.g. the iPadOS and Mac Catalyst platforms inherit their availability
from iOS.

In the case of articles, the platform availability information is not
used in rendering the page. However, sample code pages are processed as
articles with a different page kind, and these pages do render the
platform availability information. Hence, it is necessary to also
process the fallback platforms when setting the availability for an
article's render node in order to correctly display the list of
supported platforms for sample code pages. This patch updates the logic
in the render node translator to process the availability information
for articles correctly and include fallback platforms.

The `DefaultAvailability.fallbackPlatforms` variable provides a mapping
from platforms to their respective fallbacks. This map was being
deconstructed with incorrect terminology in a few places, with the
platform itself being called "fallback", and the fallback platform being
called "inherited". This has been cleaned up to make the nomenclature
consistent across the codebase.

## Dependencies

N/A

## Testing

The unit tests for availability information have been updated to verify that fallback platforms are included, and that their beta status is correctly inherited.

## Checklist

- [x] Added tests
~- [ ] Updated documentation if necessary~

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -838,16 +838,42 @@ public struct RenderNodeTranslator: SemanticVisitor {
             }
         }
 
-        if let availability = article.metadata?.availability, !availability.isEmpty {
-            let renderAvailability = availability.compactMap({
-                let currentPlatform = PlatformName(metadataPlatform: $0.platform).flatMap { name in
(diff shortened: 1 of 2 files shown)


snprajwal:
@swift-ci please test

PR #1410, opened 2025-12-19
[6.3] Include availability for fallback platforms in articles

snprajwal:
- **Explanation**: When DocC processes the availability information for symbol graphs, it identifies if fallback platforms are available, and if so, injects platforms that inherit their availability from the fallback platforms. E.g. the iPadOS and Mac Catalyst platforms inherit their availability from iOS. In the case of articles, the platform availability information is not used in rendering the page. However, sample code pages are processed as articles with a different page kind, and these pages do render the platform availability information. Hence, it is necessary to also process the fallback platforms when setting the availability for an article's render node in order to correctly display the list of supported platforms for sample code pages. This patch updates the logic in the render node translator to process the availability information for articles correctly and include fallback platforms.
- **Scope**: Render JSON update, UI update
- **Issue**: rdar://158230351
- **Risk**: Low.
- **Testing**: The unit tests for availability information have been updated to verify that fallback platforms are included, and that their beta status is correctly inherited.
- **Reviewer**: @d-ronnqvist
- **Original PR**: #1408


snprajwal:
@swift-ci please test

PR #1414, opened 2026-01-02
Add support for GitHub Actions

shahmishal:


Tests/SwiftDocCTests/Checker/Checkers/NonInclusiveLanguageCheckerTests.swift
heckj, line 1: This checker is explicitly trying to find these invalid words in these cases, so I don't think we want them changed. Not 100% sure how to say "this is acceptable" for the soundness checking in a multi-line Swift string here though

heckj (changes requested):

Tests/signal-test-app/main.swift
```diff
@@ -24,7 +24,7 @@ Signal.on(Signal.all) { _ in
 }
 
 DispatchQueue.global().asyncAfter(deadline: .now() + 1) {
-    kill(getpid(), SIGABRT)
+    kill(getpid(), SIGABRT) # ignore-unacceptable-language
```
d-ronnqvist: This doesn't compile

Tests/DocCCommandLineTests/SignalTests.swift
```diff
@@ -28,7 +28,7 @@ class SignalTests: XCTestCase {
     
     /// Runs `signal-test-app` and confirms that it exits with code 99.
     ///
-    /// `signal-test-app` sends a kill signal to itself and uses ``Signal`` to intercept that signal
+    /// `signal-test-app` sends a terminate signal to itself and uses ``Signal`` to intercept that signal
```
d-ronnqvist: "kill" and "terminate" are terms of art in the case of signals to stop process execution. 

In this case it might be better to spell out the specific signal being sent to avoid any other interpretation
```suggestion
    /// `signal-test-app` sends a `SIGABRT` signal to itself and uses ``Signal`` to intercept that signal
```

.github/scripts/format-check.sh
d-ronnqvist, line 1: These checks feel unrelated to the PRs main purpose and could IMO be added later.

d-ronnqvist:
It's a bit pointless to request additional reviews when the feedback so far hasn't been addressed yet.

d-ronnqvist (changes requested):
There's code that doesn't compile. After that is fixed; DocC's own inclusive language test fail because all the non-inclusive language has been removed from that test content. 

Also, the whitespace checking script seems unnecessary and unrelated to having supporting GitHub actions. Please move that to a separate PR so that it can be discussed on its own merits and not hold up this PR.

.github/workflows/pull_request.yml
```diff
@@ -0,0 +1,27 @@
...
+
+jobs:
+  tests:
+    name: Test
+    uses: swiftlang/github-workflows/.github/workflows/swift_package_test.yml@0.0.11
+    with:
+      linux_host_archs: '["x86_64", "aarch64"]'
+      linux_swift_versions: '["nightly-main", "nightly-6.3"]'
+      windows_swift_versions: '["nightly-main", "nightly-6.3"]'
+      enable_macos_checks: true
```
shahmishal:
```suggestion
```

shahmishal:
Closing this in favor of https://github.com/swiftlang/swift-docc/pull/1537

PR #1417, opened 2026-01-08
Add live reload to preview server

DePasqualeOrg:
Connected browsers now automatically refresh when documentation source files change during preview. This eliminates the need to manually refresh after each edit.

## Changes

- Add Server-Sent Events (SSE) endpoint at `/__docc-live-reload__` for push notifications
- Inject a small JavaScript snippet into served HTML pages that listens for reload events
- Notify all connected browsers when the directory monitor detects changes and conversion completes
- Reduce directory monitor throttle from 1s to 300ms for faster feedback

## Testing

**Manual:** Run `docc preview` on a documentation catalog, open the URL in a browser, edit a markdown file, and observe the browser automatically refreshes.

**Automated:** Added Swift Testing tests for SSE headers, heartbeat, and script injection behavior.

## Checklist

- [x] Added tests
The tests hang before completing, but my new tests pass. I think there may be an existing problem with the tests.

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -1766,7 +1766,13 @@ public class DocumentationContext {
                         ]
                     )
                 )
-                return nil // Don't continue processing this article
+                // In the case of a collision, we drop the article in favour of the symbol.
+                // It also needs to be removed from the list of uncurated articles, so that
+                // it doesn't get pulled in for rendering during link resolution in case
+                // it is referenced by any other page.
+                uncuratedArticles.removeValue(forKey: reference)
+                // Skip processing this article
+                return nil
             } else {
                 documentationCache[reference] = documentation
             }
--- a/Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
+++ b/Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
@@ -1243,6 +1243,8 @@ class PathHierarchyTests: XCTestCase {
         
         let articleNode = try tree.findNode(path: "/MixedLanguageFramework/Bar", onlyFindSymbols: false)
         XCTAssertNotNil(articleNode.symbol, "This should be an article but can't be because of rdar://79745455")
+        // Ensure that the article is dropped from the list of uncurated articles
+        XCTAssertTrue(context.uncuratedArticles.isEmpty)
         // FIXME: Verify that article matches are preferred for general (non-symbol) links once  https://github.com/swiftlang/swift-docc/issues/593 is fixed
 //        XCTAssertNil(articleNode.symbol, "General documentation link find the article")
     }


heckj:
I love the idea, but please detail of the feature you're trying to add in a post to the forums as a pitch or proposal. The [swift-docc topic in the Forums](https://forums.swift.org/c/development/swift-docc/80) would be an ideal place. I'd also tend to thing you might need something in swift-docc-render to work with this as well, and if there's coordinated PRs, it'll be easier to communicate about using the forums as a coordinating point, or at least a head's up to the relevant folks.

heckj:
@swift-ci please test

DePasqualeOrg:
@heckj, I don't believe this feature requires changes to swift-docc-render. The live reload script is injected at runtime by the preview server when serving index.html. The template files themselves are unchanged. The SSE endpoint is also handled entirely by the preview server.

Here's the forum discussion I started, as requested: https://forums.swift.org/t/pr-1417-add-live-reload-to-preview-server/84026

heckj:
I'm sure it doesn't as it stands -  it looks like it'd work beautifully straight off. It's more of a question of if this injection is going to run afoul of anything they're doing. They've been resetting dependencies and updating DocC Render over the past few months, but without specifics, so this is really intended to be a cautionary check against potential non-public plans, not a "does it need it to work" check.

Tests/DocCCommandLineTests/PreviewServer/RequestHandler/LiveReloadTests.swift
```diff
@@ -0,0 +1,86 @@
...
+#if canImport(NIOHTTP1)
+import Foundation
+import Testing
+@testable import DocCCommandLine
+import DocCTestUtilities
+
+import NIO
+import NIOHTTP1
+
+/// Creates a temporary folder with the given content and returns its URL.
```
d-ronnqvist: Is there any way that we can test this without creating real files on the file system? For example, could the `DefaultRequestHandler` be passed a `FileManagerProtocol` or alternatively be passed the `Data` that it's serving?

DePasqualeOrg: The existing `DefaultRequestHandlerTests` and `FileRequestHandlerTests` also create directories, but unlike the tests I added, they don't clean them up. This approach tests the actual code path rather than a mock. Could you help me understand what the concern is?

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
```diff
@@ -20,15 +20,36 @@ import NIOHTTP1
...
 /// content, depending on the requested URL path or query parameters.
 struct DefaultRequestHandler: RequestHandlerFactory {
-    
+
     /// The root of the documentation to serve.
     let rootURL: URL
-    
+
+    /// Whether to inject the live reload script into the response.
+    var injectLiveReload: Bool = false
```
franklinsch: Is there a reason why this property exists if it seems to be set to `true` always, at least in the production code?

Sources/DocCCommandLine/PreviewServer/LiveReloadClients.swift
```diff
@@ -0,0 +1,73 @@
...
+    private let lock = NSLock()
+
+    private init() {}
+
+    /// Registers a channel to receive reload notifications.
+    ///
+    /// The channel is automatically removed when the connection closes.
+    func register(_ channel: any Channel) {
+        let id = ObjectIdentifier(channel)
+        lock.lock()
```
franklinsch: nit: you could use `withLock`, here.

Sources/DocCCommandLine/Action/Actions/PreviewAction.swift
```diff
@@ -209,6 +209,9 @@ extension PreviewAction {
                         throw ErrorsEncountered()
                     }
                     self.print("Done.")
+
+                    // Notify connected browsers to reload
+                    LiveReloadClients.shared.notifyAll()
```
franklinsch, line 214: Do we want to notify if the compilation has errors as well? Not sure what the implications would be.

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
```diff
@@ -20,15 +20,36 @@ import NIOHTTP1
...
+    /// Script injected before `</body>` to enable live reload via SSE.
+    private static let liveReloadScript = Data("""
+        <script>
+        (function() {
+            var es = new EventSource('/__docc-live-reload__');
+            es.addEventListener('reload', function() { location.reload(); });
+        })();
+        </script>
+        </body>
+        """.utf8)
```
franklinsch, line 37: If I understand correctly, this script is injected on all platforms, but `DirectoryMonitor` is only available on macOS. Should we guard this to include include the script on platforms where `DirectoryMonitor` is supported?

Sources/DocCCommandLine/PreviewServer/LiveReloadClients.swift
```diff
@@ -0,0 +1,73 @@
...
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+#if canImport(NIOHTTP1)
```
0xTim, line 11: What happens when this isn't available?

DePasqualeOrg:
> Thanks for this work!
> 
> Would it be possible to add tests for `LiveReloadClients` functionality?

Are you sure that would be valuable?

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
DePasqualeOrg: I've removed it.

Sources/DocCCommandLine/PreviewServer/LiveReloadClients.swift
DePasqualeOrg: Done.

Sources/DocCCommandLine/Action/Actions/PreviewAction.swift
DePasqualeOrg: That would be closer to the behavior of Vite etc. but would be a significant addition. I think I'll leave that for someone else to add if they think it's worthwhile, since this PR is already a major improvement.

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
DePasqualeOrg: On platforms without DirectoryMonitor, the script connects silently and does nothing. If file watching is added on other platforms later, live reload will work without changes.

Sources/DocCCommandLine/PreviewServer/LiveReloadClients.swift
DePasqualeOrg: When it's not available, the live reload feature doesn't exist.

heckj:
@swift-ci please test

Sources/DocCCommandLine/PreviewServer/LiveReloadClients.swift
0xTim: Ah my bad I missed the `canImport` in the place where this is being used

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
franklinsch: I understand the forward-compatibility argument, but I think I'd prefer we only inject the script on platforms where it's functional. A few reasons:

- It avoids opening an SSE connection that will never receive events
- If/when file watching is added to other platforms, adding the injection at that point is straightforward

franklinsch:
> Thanks for this work!
> 
> Would it be possible to add tests for LiveReloadClients functionality?
>> Are you sure that would be valuable?

Looking at the current tests, they cover the script injection behavior in `DefaultRequestHandler`, but `LiveReloadClients` itself has no test coverage, unless I misunderstood. I think tests would be valuable here because:

1. It's the core of the feature: `register`, `notifyAll`, and the automatic cleanup via `closeFuture` are the key behaviors that make live reload work
2. The SSE message format matters: confirming `notifyAll()` sends the correct event format would catch regressions

Specifically, I think these tests would be helpful:
  - Registering a channel adds it to the active set
  - `notifyAll()` writes to registered channels
  - Channels are removed when their `closeFuture` completes
  - Inactive channels are cleaned up during `notifyAll()`

Sources/DocCCommandLine/PreviewServer/RequestHandler/DefaultRequestHandler.swift
DePasqualeOrg: Okay, it's now injected conditionally.

DePasqualeOrg:
I added the requested tests.

Tests/DocCCommandLineTests/PreviewServer/RequestHandler/LiveReloadTests.swift
```diff
@@ -0,0 +1,86 @@
...
+private func makeTempFolder(content: [any File]) throws -> URL {
```
franklinsch: Nit: you could consider re-using `createTempFolder(content:)` here

Tests/DocCCommandLineTests/PreviewServer/LiveReloadClientsTests.swift
franklinsch, line 1: nit: this file should probably be named `LiveReloadClientsTests` to match the type under test.

Tests/DocCCommandLineTests/PreviewServer/LiveReloadClientsTests.swift
```diff
@@ -0,0 +1,111 @@
...
+
+        guard case .body(.byteBuffer(let buffer)) = try channel.readOutbound(as: HTTPServerResponsePart.self) else {
+            Issue.record("Expected body response part")
+            return
+        }
+        #expect(buffer.getString(at: 0, length: buffer.readableBytes) == "event: reload\ndata: {}\n\n")
+    }
+
+    @Test
+    func closedChannelDoesNotReceiveNotification() throws {
```
franklinsch, line 42: Do we need a defer to close the channel in this test?

Tests/DocCCommandLineTests/PreviewServer/LiveReloadClientsTests.swift
```diff
@@ -0,0 +1,111 @@
...
+            Issue.record("Expected body response part for channel2")
+            return
+        }
+        let expectedMessage = "event: reload\ndata: {}\n\n"
+        #expect(buffer1.getString(at: 0, length: buffer1.readableBytes) == expectedMessage)
+        #expect(buffer2.getString(at: 0, length: buffer2.readableBytes) == expectedMessage)
+    }
+
+    @Test
+    func inactiveChannelSkippedDuringNotify() throws {
```
franklinsch, line 90: Should this test also clean up the inactive channel?

DePasqualeOrg:
I've made the requested changes to the tests.

Sources/DocCTestUtilities/FilesAndFolders.swift
```diff
@@ -253,6 +253,23 @@ public struct DataFile: File, DataRepresentable {
...
+/// Creates a ``Folder`` and writes its content to a temporary location on disk.
+///
+/// This is a freestanding function that can be used with both XCTest and Swift Testing.
+/// The caller is responsible for cleaning up the temporary folder when done.
+///
+/// - Parameters:
+///   - content: The files and subfolders to write to a temporary location
+///   - pathPrefix: A prefix to use for the temporary folder name.
+/// - Returns: The temporary location where the temporary folder was written.
+public func createTempFolder(content: [any File], pathPrefix: String = "TempDirectory") throws -> URL {
```
d-ronnqvist: Apologies for the mixed message but I don't feel that we should be adding this function. We should be moving away from doing real disk IO in tests, not make it easier to do so in future tests. Also, this test helper doesn't clean up after itself which is a reason why it's better if its definition is closer to the caller (maybe even inline, and not extracted at all).

d-ronnqvist (changes requested):

DePasqualeOrg:
@d-ronnqvist, @franklinsch, it would be nice to get this wrapped up.

DePasqualeOrg:
@d-ronnqvist, @franklinsch, it doesn't make sense that this PR, which adds important functionality, is now in deadlock over some minor test issue. Please just make whatever change you want to see in the tests so that this can get merged.

franklinsch:
@DePasqualeOrg apologies for the delay, and thank you for your contribution! I am keen to get this merged. Given that other tests in `PreviewHTTPHandlerTests.swift` are also using a temporary directory on disk, it would be great if:

- We could at least add a FIXME to the code to indicate that we should clean this up in the future to not rely on a file system. 
-  We could reduce the duplication of `createTempFolder(content:pathPrefix:)` utility you've created as it's essentially the same as the existing `XCTestCase.createTempFolder(content:)` method below.

DePasqualeOrg:
I've consolidated the two functions into `createTempDirectory`.

As a side note, I notice that there's a lot of usage of "folder" where "directory" would be more appropriate. A folder is a GUI concept, and a directory is a filesystem concept.

franklinsch (approved):
Thanks @DePasqualeOrg!

franklinsch:
@swift-ci please test

franklinsch:
@DePasqualeOrg can you please rebase?

franklinsch:
@swift-ci please test

snprajwal:
@swift-ci please test

snprajwal:
Thank you for working on this!

PR #1420, opened 2026-01-16
Implement a fast custom decoder for symbol graph JSON files

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This implements a custom fast alternative JSON decoder specifically for symbol graph files. 

> Apologies for the size of this PR. I can't think of a meaningful way to make it smaller because it needs both the complete decoder (730 lines of code and 384 lines of comments and documentation) and the decodable conformance for all symbol graph types (1243 lines of code) to be usable. 

The custom decoder implementation is designed around a single linear scan through the bytes of JSON data. Because of this, each decodable type needs to handle the keys and decode the respective values in the order that they appear in the JSON data. The `FastJSONDecodable` documentation comment includes two code examples that shows how a type with properties would implement their decoding and how a single value / raw representable type would implement their decoding.

For future contributors there's also tons of real examples in "SymbolGraph+FastJSONDecodable.swift" which contains the decoding implementations for all decodable types in SymbolKit.

This decoder implementation is makes heave use of `UnsafeRawPointer`. To ensure that the pointer can't exceed the lifetime of the data that it points to, the `FastSymbolGraphJSONDecoder` has a _private_ initializer and is only ever created from the static `FastSymbolGraphJSONDecoder.decode(_:from:)` method where the decoder is both created and destroyed within the scope of `withUnsafeBytes`:

```swift
package static func decode<Decodable: FastJSONDecodable>(_ type: Decodable.Type, from data: consuming Data) throws -> Decodable {
    try data.withUnsafeBytes { buffer in
        var decoder = FastSymbolGraphJSONDecoder(buffer: buffer)
        return try Decodable(using: &decoder)
    }
}
``` 

This decoder implementation allocates space for "only" 64 coding path components. However, in practice the symbol graph JSON structure is rather shallow and doesn't go deeper than 8 components. Having this eightfold headroom should be plenty of space if the symbol graph structure changes in the future and becomes significantly deeper. 

This decoder implementation makes direct byte comparisons of the expected `StaticString` keys because all symbol graph JSON keys are known ASCII strings. Because there aren't any characters that need escaping in any of the keys for the current symbol graph structure, this only relies on the tools that emit symbol graph data to not needlessly encode ASCII characters in the `\uXXXX` format (where `X` is a hex-digit), for example using `\u006D` instead if simply the `m` character. I consider it extremely unlikely that the symbol graph tools in the Swift compiler or in Clang would start doing this, so—even though [the "String Comparison" section of RFC 8259: The JavaScript Object Notation (JSON) Data Interchange Format](https://datatracker.ietf.org/doc/html/rfc8259#section-8.3) specifically calls this out as an interoperability concern—I don't have any actual concerns about this in practice for decoding symbol graph data.

The decoder _does_ correctly handle escaped characters and escaped unicode characters like`\u006D` in the string _values_ because those could contain any text.

The core of the key matching/comparisons is the `UnsafeRawPointer.hasASCIIPrefix(_:byteOffset:)` extension which takes a `StaticString` as an argument and provides optimized loads and comparisons for each length, with a loop if the key is longer than 12 bytes. 

This implementation inlines and optimizes _really_ well because the parameter is a `StaticString` so that the compiler knows its length and all its bytes and compile time. You can see in [this Compiler Explorer comparison](https://godbolt.org/#z:OYLghAFBqd5TKALEBjA9gEwKYFFMCWALugE4A0BIEAZgQDbYB2AhgLbYgDkAjF%2BTXRMiAZVQtGIHgBYBQogFUAztgAKAD24AGfgCsp5eiyaglAdwI0i5FY1RECQ6s0wBhdPQCubJiADMAOzkLgAyBEzYAHLeAEbYpCAArOQADuhKxI5M7l4%2B/kFpGQ5CYRHRbHEJybbY9lkiRCykRDnevoE22HbFTA1NRKVRsfFJNo3NrXkdSuMD4UMVI4kAlDbonqSonFwApFoAgjSeTKgA1EgsSqqk2HTqEAD6p2nhRCOnCkxKLDTYAEosMyqdCveLLU4AWh2flwpwAQugPKcdgE4Xt9qdMacbkQNkxniDhPEAHT0dAsTCfCQEYARTAQLGMpnMzGXEAfACSwgAbNJibYaODoQAREWnLTqPwADgegVl3NliVl0llfllACZZTxTgB6HWnbXq05%2BU7SU6JU7c04BU5S9Es5Hq7k7J0E0GkUnkymseg0um0UjoNhwgCebwA8jQaCoiOzbeRTmzOTy%2BQKhX5RRnxeppDLpAEHtIFdIldIVdI1dJNdJtXrThi4adXKdhadYQAxU4AcVOAAl7SyXc7XS8iR6yRSqb7adh6TRA8Gw9hI9HsLGDbJE0p2QouURefyuoLkRmxRLpLhC8LC65C3DC/tCwBOQt5m11nu904c04AKVOADSpwhKcACypyROiKKigcXCrPQ3CJPwvhcDo5DoNwIgWFYpxKOsmzYI6fh8OQsaoXBqwANYgNIT7Ek%2BPA8IEfjqlKLEVokT6GNw0j8GwNF0U%2BQnCSJInkChaEYVw/DblopHaHB5BwLAKAYDg%2BDEGQlDUHQjCsBw3AkYIwhiBInAyHIwjKGomjkeQ%2BjqoYximNh1g1HUTgQC4ky%2BOqsihPM5SVCAUqpOkmRCD5IB%2BWFRRZIMQUjKF7k9H0EweG0UhySl9SzAlwwJMlsxRTw2V5YFBUhaseEbFs3DokcJznJc1y3AQ9xPKObwJB8Xw/P8gLAu64JQjC8KIvQyKogOmJdSSFxKPsIiuByHKtXcEAuuqPDqn4JbcgEUr7HCrjCrg7Zdr2HK/gBISgZB6rqssUEBDB%2BzoikpAEAAbiwbynNg6hvF8WS9d8vwAkChLdVNaIHDNpyNWcC1LSta03BtnUY%2B17J9A4qANF9JgJjES4rjG7J7ierZaCN0KwgiSIonDGIspcKjNBAn1teoxIEIty2rQmW0i49COMiLiSuFz2PqM9j2nPzTBgFwRCJviqOrbhRBE8AxLfkQKtKImzyBsApDsGw4TAADpCBqQpwkM87OnEwQgQprP4zLrxtO0QSD86cHD%2B1gxLi1iovqltz0HA64eYnWXK%2BhE5xdCk8SI8cHlfI76CnKgNx/YRbCePQDgQqT/2/V42DG%2BsavoDQjtIIRMx/QQZze9bRunDEnhRiS8eZ01E6YOqoZvEoAZBr3/e/D1nzg9gcJz/EQ1jtCri7sIdowgmjeruuVNnnTY3b0QPBWszQ%2BMjieLJhf3IQH3A%2BkDsiQIlGMaOsOjZaO/woY6swdIOAIzZz6X2fqvN%2BH8D7fyHC6RsPAAFCidJvTemI7Sx1AW9OO2DmRI1OKPaQE9a7TzYLPV%2BO4%2Bq/BXq/de3VN7n13rgfeX81yU2ENTcUp9YTnxYrDG%2BWI76kHxPw9UUDX7v0/ofH%2BiDxQoKEaA8Be4WKSPntIuBa45HjwNCguRfhXAYNtEo5kKIVHCDUS/DRsD2FqwQboqOiRAEGKMYYg0zp8EgKZOYh%2BVjoGaLsToxsfh9EIMMRgvyQ9oLRK8YyQho8pSkKnvOGe1j3iL36nQ%2BeDD4hML3Cwthh9OH2KzLTSE9MH68kEXElkIixF7l5OovJtjZEOL/oo2p3ipoWP3NIZpMCZHwLQbo5BzjUHOgie4kxXTvG%2BPPk09JgytH2JGY2JxLjwluObJfUxPiwFVP6UswJbS1nGjCWgqZzYomzJAfMxpRyAmtOGb/U0FzJnbONFHW5ccDkLMeVI552j2nmneeg9x0h/4/OUYcgZJyXnyOdOM1xGDcx7Iln8h5cKgWrNeQEMFVzzSeOAfs3Bg5Zl1mssibkWgLD%2BwUAAFXbEkgJNLcJ5ydhcJgmBGC93QP7U4lFwiYDrk3PGHdCbW0VmOH0is2ApEYBwYQ7chDGy8sQFuDsWBujHKcMgJsMgmF5Z8DuWBW7iCMA7aunhsDLDDrM%2BppsebEjpUgRlzLskZ2ZpQ%2Be0rTF1jhLUFgngVDN0DuEZOtd1aYF7kGkNhF/atx1lKwOlE3ZmHxNqjA8qGAJoIBwfOqrGjCATImwtObGAO0DiwGIvKnboBSA4K2AAvBNLdQ0WvoMbV4ecMjyt5RG%2BYQd0DfUIuEb2nhs56qbsYaVbxgAZ19N1CQSh7Ukrweulk5hiCoCQE6u4xJPBEBoFKdwOATUtHWFw6%2B0L86XEIloEA6LMSOp1ja3U%2Bp9j4mwPKogIZta60VsbCQZgWAhmNhgEGOAbgxu1dzO4Qc/q7ufciW94hQ08CfbexkdYQhemjeWlITRCLasNcAXlldsAocdaPKcfpZzkNIeTDhvcyZ2ITEmZhh56DHhFGKY5H9/7OJQ6Y9DhF1RYc3d43D%2BHjAxuzURm4JsjSUaUNRtc99aM%2Bno3OBcTG7Hsko8x6wW4dx7kvtx3jp4syj3HkuFJC5DPQKAd0nxaH73Gkk65xk7N4g9HbCwBgGxsCbTFmF75UmQEckNlwY2xwIhbCUN8L69B/1O13bUSiiN9Xar8BXJc%2B72qxvEPGxWatsAaozjcAAjp4AgNxjZGlHkofeWqiH4ddSbNglwsv6qUAHHCDb4gqqYPrd%2BRjsNYnbPqggOBhCWBDFK38IhwyREFdgcD0JGwsBSOnblzcE0WwYFKruJhTg4F9FbGGlwDVzyK3hZuf0DuFfUJnegqXEYMC7acTwKQTbSHy1XCQNq13eYjuF6OKG6wMrziwb6IIY2YEJDbIgZhe3YEU0XbE2AKTGzMIRUeppWOTwI91pQWXARgYNNOg7FC4iCCUwppoUq/Yt2NMTqNjdtbt07sms7mA/osFB2Dj9pwGUB19rUJATAO4SA%2B19cnxtC6EDO3EEMQgY1lpiFekVNOdg0oFPruSs81ZxBK6G/HiHyeJlA2ltngMdbavrhCRuFcdfG0owmYg530C12VqrRDRBurC7B4wU39mYfuHlcRmH3DxHYrhEJzZlzPkzMiyLsH9zLESIE3CMZyePkYN2ZNjPML4%2B542RM8F1zpDqdxKI9rk5tMzl00GfTxSOfGY49uPx6pLMTOzFoLQDx2zthH2P0f1N%2BMR/QFHxT2AYcifc6G6QXmRcyYpARpnSntVmlU3XzTXo6Mt8Y2xjvRn2Omd7/3qfNmvQkPs%2BQpzr8XMi9Ex5xIa%2BweJybmW5bq31t/1A4LRVNS0W58RLdxB/c1ZdJJptUjh3tbQOdThrVCJsBR18RLBThLdwhUAvAcBntGAKQpU5NHYjtk4bZTsbYLt81iB4hV0od9QAB1QiXQENNWC4UdXORMVALYRtR7NWf2QOGrOrBrE2GIYgCwUNUg8dN9KdenMgQiSA2dbrSiNtQibffmIQEPbzRgsXWHeHWbH3FnNHXCDHJoLHZXPHAnfDKUDnYDfbMnCnW3dnLnRNOnW4RQwjZnM7VnQiEA%2BzGnNufGADKVAXRoHQ1zaHCXR2KXGXTtf9BXSiJXHHFXG2NXDXZ7bXY4XXLnI3Q3GlEmI9YrYNC3YuHrG3KnMtB3C2PVI9F3GgN3HIj3JcL3NWJHP3FWNWbrIPQeEvIhbRVTSPIMefWPMUf5BPJPKvQlG5dPUvbpLPPpBPfPaYz5FiFDeYiOTFFMBPSvFFCFKFOYzY35XpRZJ5OEUJZFLZVFLBI444sxbYpY3PWvK4lPDBRIYlDPGjI/Zvf0VJRcCMAzTvS/TjB5G/IcQfYfIfB4aE0fcfeEuEyfPjLMIY2fEYmPdAJfO4sTS0L/bzXzZoLIALILG4ULKOCHfo6LHueLWoWuZLBgNLPODLVALLBnE2bkQHQieDdqEmONUNb3CrRNB2EQ%2BrKNJrL0FrPVNrQnTrbVJwqU3CAbBudOC2HoMbKWFDabKtObBwGgRbM7f/NbNQzbPwbbXbFwQg1IqVB4OTJ4R3Y7M7Kg87LoWg7qYDYDF7AjJQO7V7B7f2J7Mtbk17BAj7WAuLP7bVKUTklA4HbASIhYiHR6V/b/fUWPOHBHYw3w0wlQTHf6Kw7AmwzfOw1TUnCoynf9I0NwluDwhnDQ9EwDPw80ewoIxoEIp08IoXPQ8XQON4XdeIuXRI/mZI7HYg1XDbTIrXd3PXA3I8I3Io03PkpQ8o63cs57Gop3eo13bI7lFot4Non3TogPHo4PFDMPew4Y6PG4MYrMCY3PKY/YnZT4%2B4t/R4s4wFPPAlT5WY58l804gFGxOEPY649xdY/o%2B4xYt8gCy4gvavU0Q4n8zPV8/8lpOEF4mCwlNFMC44iC5CwZNC1Y94p8sHb4pvakE/f49vCmYEw%2BbvMzFMcE10CUaE2EifVihE1i2/VsVEufDErE1zHEgIPE1zAk/zQLLwUkyOEWSkmLOLJgBLOk5nD7dLFuFk7LNrAIaMoM3k83MdcrSrYU7AWrUUxrRvEVVrE2GUjVLrCovrJUvVFUkbdUibO4xkLUxWHUhbJbFbI0jbNTU0xMc0/bMte0ig0I/nF0q7DOG7UjH09lAQ57IMt7UMr7cMk2KMyjGMmueMuZRMp6LsgwjMpHEw9HXMwifMy3QnYswI0g%2BU1c7UKsn9WNWs7whsjlNnAIknLnYIjuMK6gwXbKkBaInsuI2Xd7QcxXEctI2NdXIKtnbc3IpufI2cwok3Eo0rS3Wqlw6ooGWo53Lcqcz3MrA8pQaAwPE8/os87i%2BshfPOcYrFO8z84xW4hC3QpC5Yx69xYvFyhCnC3Yj6mvDY7Ct63PaCgikCiLF61zX654/6uCwGzY6G84/Ch800Z6yGuZYGpG2Gj4g/BvLTMiv4vTM/Kii/Giq/f5Biq0Ji4fREti2m9izi88tEy8m6vi7pHEqUIS7pDfWDfbbfEjJA/ffoki70AmhjCi4mljUmmMWiw5cE6zVsRJZJJ/H1MENm6TT9Jgf9J0skEwDOf0/EKqknCIWcX2POOIfOFStQmNcIR2Uw/VNgLwzQvCL4AaiWW9OsNyp8ZuQMTwYAPdHaBMZQ/EQnHgBMRyBUikQgHoCQInVTN2iOZfQiJ8LmjW4CWTPm66tK%2BwsA5gUyrfLO4KtnJ8AVSjBO5kEW4/QmtvSW9caWtcBMHzHvCmtMRmpWx/f45/eeZM0vBxK0UeBkB0CWwE8/Wux0RsUKcmgpG/Rm3PJ8ABdWkBHE0qVOwa/UPDTfUg/m7O0Ag7EOjO%2BTQutnIwGYU4FTezcupkSu348Womkekm2uxurEUE%2Bi1u5ExWr0JJDuxzVW0gHujPPu0yqu2%2Bmu%2B%2BqWse%2BRSel%2Bx%2BeWzMD%2BikOzSeFWpZIpKiqUf%2B8lbEjzRiVeh0Hmgulmne%2BzXO/ezezOwhstE%2BtWPfYh6VPAzwKastKUUugrW27XAVZ21VS%2BnDVMmI1QqNb3P0i4U3flPdMtFQSDGNQnMM57EMuO6q/bJHf3GNE2mNI4ZoTVXuQRpUqVfVCQ7onra2bh4RDTPGn4sW1vAE5cIE%2BukzJuuivpWBsUdupBzu3%2BjBzPEZfu8x6cauqx4zQzCB3RIIKe7PJxu/CkB/Vxn%2BlB6dDvAIDxtzLBjDCTPQje3mw%2BwhyMnOve/Oreo%2BgnS4ah%2Bw4xl9Ux0h0W3xkB/xmxx%2BpkaBg8N%2BhW0yr%2B6JoMLutWrCzEQB/Gqpyxyi8BsB1ZCe2W8RcJ%2BBzAKJshNx2JlZdkdB0xf1Zg8A57J0wOB4XWhdB2A2p4HaHJ4gY2GgIp/WvOfhpK20aM8qqykdeIIwP7A2o6tHUuGNC2p2M5kuMuAgBVVuJAMgGGKwy%2BvUJZ04NywGdgb5hMSDDIaDVZvnG2LaIjf2OfIQZgIgNTBWMwAOXdIDDcS2iw%2Bweg3BrEH/Asya2R%2BrU%2Bo2qNMkxFpAOfaOOdPOFFzK99UgyhopwWwIsk%2BVFF4QdFp6Rl9WURjOVAoOjQ2dZkrLQM2WaVU4Ng0%2Bo0Th12pZ4F2aP6OlkY3ltF1Vg0XafaQ6HV5kHaPaD4g1j2nUYF3sdAfHUdCgRWJuS3HWRkxDNQ00aM/VdUS51Is2xMQwmNG50gO5/ckUsQk0fM0gk0JV%2By4bHoH1yVhK2WIlhOC181rEWl5FiIPlnV41/VtGtO7xHNw17pHaFV819/UNM8jAY4WMPQ/YY2TZrcA7HfJTOwx2pTEsp2ZXWV8R1snqqgnuGYMgFgBdBMQnEOUNKlhg/o36B2FZbheCkXTF3NOJ%2BBdxCAKt69J0CEW0IUaaLp04YATwJoKRnxnTU/IZwJoZ8eldhu0JxxppuBlp5WmZ6BVBljFZcELoUNG9b67xR1I5rtKjfdmJfdud9pMUPN16slFM%2BsKweIUDUgMyy2zLQgjlyd7AqyvCAtANu5lDMDrMDd3FbdyD1zQ949oBm%2B/p2uy96x05RsFZUZsEh95xz%2B59mJ19m99cD9gGQDmpX9q%2B8pxGFdID/jkD0T4WwTt9ETqG16aJWTrxMTsTxSBCLgJCcSBSdCTCVydlWqQiF0FifgMiHQZYVYFuCkEYCAKiJILQYkLQKUKUbkJ0AIS%2BAID4qsbiLgXidTuyKSGSEAOSIziiDz9UPiGiAIWzosGsPwRILQbaLQJ8IICSfgXz%2BSciVYZSBAeACAVSEY3NCgKgdd3LytEAbdKwRzgQBgN06gGIDTiQ1gUgEMQyfgOrpoEMcMGIXQWoMiEibNJVIgcMJgVLDTnAPuYAVwOXbcXgfgHAbrI1bYNCQgG4OoUdSbtCQGWoI9bYEiUEFTtCX0GIC2Brs9ebwzr6fiKb8gW17XFQYUH9ZyCghSVYGgIwYAJQAANQIGwDMHDD2ya8slEAtXMlkGMkUBUA0A0/0DDqMBMBK9ckMAIBiG3EgFWAbVje4AhHDGS9ta%2BhwCR8s86G6CyGcG5RKmCG5XykWASDDsKAimyAyjyGp/Ch6Ap%2BCjDpyiEDSkvVyF8DZ66C69yn6BZ5GDZ%2BKnp557GEF4qkp6kGqnwjqh4HgkQmQg06klOHUAc4hGqVK6IDOG5GJCNAgA0hIAdn04V8M8e9M9SIs6s5i9s4YkSCLD8G5Ac4Oni%2B5A868/4lonolEl96Em88km4D84C8e6UkQGy7QCK/iG0kK4rRGG1/K90iq%2Bflq/CFa7%2B5a4a/a86/sD%2B969RYG6G7shG79vG/e0m5Ihm/u5O/IEW/55W40/W9QE27%2B5240/28O5DGO40%2BTXO5Iiu/SGwFu9m/I2tlD%2Be%2BHfe8%2B%2B%2B9%2B4u5B9MkkAshB%2BsnB7sgcichh%2B1/h8R/gBR8bSyEm4x6x/iBx9rl34J/588m8jF4MACjKEqkZ7ikihv8f9p6F6p4v%2Bzk59J/Z96HKnv%2Bl4i9%2BgP/f/gsFZ6y9dO0URXqp2V4%2BduAavDXlr1ci699epwQ3tHX1T6cno5vNLpb3M4JB8e1Eb3k%2BDs4lgnexEPwCxC4gqcvOSXTTtJBsD%2BdUuxnaASF3ID8REgfeJ8IkGi4OdaIh0EgeqC4h0CUugXEzh5z8CwDA%2BDAsQasFtYZAnA0gIAA%3D) that even for keys that are longer than 12 bytes (the last switch statement) the compiler is able to:
- partially inline only the specific case (based on the length of the `StaticString`)
- pack each group of 8 bytes into a 64-bit constant 
- unroll the loop (because it knows the length of the `StaticString`)
 
so that this implementation produces the same assembly as manually writing the loads and the 64-bit constants.
 
---

The new tests in SwiftDocCTests's "FastSymbolGraphJSONDecoderTests.swift" verifies that both decoder implementations decode all of DocC's test fixture symbol graph files the same. 

The new tests in "FastSymbolGraphJSONDecoderErrorTests.swift" verifies the both decoder implementations raise the same `DecodingError` errors with the same information, so the change of decoder implementations should be entirely invisible, even when encountering invalid JSON. One exception to this is that the custom decoder is strict about trailing commas (which isn't valid JSON) but `JSONDecoder()` is forgiving about trailing commas and is able to parse many invalid JSON files. This shouldn't be an issue in practice because the tools that emit symbol graph data produce valid JSON files.

---

In a synthetic benchmark that decoded a 88 MB symbol graph JSON file on a single thread repeatedly for 30 seconds and reported the number of decodes divided by the time taken, the custom decoder measured 10 times higher than `JSONDecoder` and the canonical `init(from:)` implementations in SymbolKit.

Using DocC's benchmark scripts on a few very large projects (>10k pages) these changes reported around 15(±3)% faster total DocC convert time and around a third less peak memory. The peak memory reduction partly comes from using less memory and party comes from finishing symbol graph decoding much quicker, so that that memory spike can come down before DocC hits its main peak memory usage. 

```
│ Duration for 'convert-total-time'        │ -17,89 %¹       │ 25,438 sec           │ 21,03 sec            │
│ Duration for 'documentation-processing'  │ no change²      │ 3,173 sec            │ 3,212 sec            │
│ Duration for 'finalize-navigation-index' │ no change³      │ 0,221 sec            │ 0,218 sec            │
│ Peak memory footprint                    │ -35,39 %⁴       │ 2,56 GB              │ 1,64 GB              │
│ Data subdirectory size                   │ no change⁵      │ 169,1 MB             │ 169,1 MB             │
│ Index subdirectory size                  │ no change       │ 3,8 MB               │ 3,8 MB               │
│ Total DocC archive size                  │ no change⁶      │ 182,8 MB             │ 182,8 MB             │
│ Topic Anchor Checksum                    │ no change       │ d776a7586d765fb86564 │ d776a7586d765fb86564 │
│ Topic Graph Checksum                     │ no change       │ 00b3e7cfb1b2a017b1e6 │ 00b3e7cfb1b2a017b1e6 │
```

```
│ Duration for 'convert-total-time'        │ -12,2 %¹        │ 6,299 sec            │ 5,53 sec             │
│ Duration for 'documentation-processing'  │ no change²      │ 2,325 sec            │ 2,296 sec            │
│ Duration for 'finalize-navigation-index' │ no change³      │ 0,243 sec            │ 0,236 sec            │
│ Peak memory footprint                    │ -33,56 %⁴       │ 1,2 GB               │ 817,3 MB             │
│ Data subdirectory size                   │ no change⁵      │ 231,1 MB             │ 231,1 MB             │
│ Index subdirectory size                  │ no change       │ 4,6 MB               │ 4,6 MB               │
│ Total DocC archive size                  │ no change⁶      │ 246,8 MB             │ 246,8 MB             │
│ Topic Anchor Checksum                    │ no change       │ 403f3569bc412791e9db │ 403f3569bc412791e9db │
│ Topic Graph Checksum                     │ no change       │ e22a998036ad243f0d85 │ e22a998036ad243f0d85 │
```

## Dependencies

This relies on the 2 new public memberwise initializers in https://github.com/swiftlang/swift-docc-symbolkit/pull/108

## Testing

Nothing in particular. There shouldn't be any noticeable differences from these changes, only faster execution and less memory usage.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
+++ b/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
@@ -1093,44 +1093,46 @@ private struct ByteMatches: ~Copyable {
         //
         //            k        i        n        d        "        :                 "
         //           6B       69       6E       64       22       3A       20       22
-        //     _11_1_11 _11_1__1 _11_111_ _11__1__ __1___1_ __111_1_ __1_____ __1___1_
+        //     _11_1_11 _11_1__1 _11_111_ _11__1__ __1___1_ __111_1_ __1_____ __1___1_   = bytes
         //
         // If the caller is looking for the locations of string delimiter, it will pass a search pattern---
         // like `ByteMatches.quoteSearchPattern`---which repeats the "quote" character (`0x22`) 8 times.
         //
         //            "        "        "        "        "        "        "        "
         //           22       22       22       22       22       22       22       22
-        //     __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_
+        //     __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_   = searchPattern
         
         // The first step to find the matching locations it to XOR the 8 bytes with the search pattern:
-        //     _11_1_11 _11_1__1 _11_111_ _11__1__ __1___1_ __111_1_ __1_____ __1___1_
-        //     __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_
-        //   = _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________
+        //     _11_1_11 _11_1__1 _11_111_ _11__1__ __1___1_ __111_1_ __1_____ __1___1_   (bytes)
+        //     __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_   (searchPattern)
+        //   = _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________   = matches
         //
         // This produces all zeros for the bytes that match the search pattern exactly, and garbage elsewhere.
         let matches = bytes ^ searchPattern
         // Next, we wrapping subtract a bit pattern where each byte only has the lowest bit set:
-        //     _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________
-        //     _______1 _______1 _______1 _______1 _______1 _______1 _______1 _______1
-        //   = _1__1___ _1__1_1_ _1__1_11 _1___1_1 11111111 ___1_11_ _______1 11111111
+        //     _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________   (matches)
+        //     _______1 _______1 _______1 _______1 _______1 _______1 _______1 _______1   (lowBitInEachByte)
+        //   = _1__1___ _1__1_1_ _1__1_11 _1___1_1 11111111 ___1_11_ _______1 11111111   = result
         //
         // This produces all ones for the bytes that match the search pattern exactly, and garbage elsewhere.
         var result = matches &- Self.lowBitInEachByte
         // Next, we bitwise inverse the `matches` ...
-        //     _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________
-        //     1_11_11_ 1_11_1__ 1_11__11 1_111__1 11111111 111__111 111111_1 11111111
+        //     _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________   (matches)
+        //   = 1_11_11_ 1_11_1__ 1_11__11 1_111__1 11111111 111__111 111111_1 11111111   = ~matches
         //
         // This also produces all ones for the exact matching byte location, and different garbage elsewhere.
         // However, these garbage bytes and the previous garbage bytes together have an important attribute:
         // they can't _both_ have the highest bit set.
         //
         // This means that if we bitwise-and this inverse `match` with the result so far ...
-        //     1_11_11_ 1_11_1__ 1_11__11 1_111__1 11111111 111__111 111111_1 11111111
-        //     ________ ________ ______11 _______1 11111111 _____11_ _______1 11111111
+        //     _1__1___ _1__1_1_ _1__1_11 _1___1_1 11111111 ___1_11_ _______1 11111111   (result)
+        //     1_11_11_ 1_11_1__ 1_11__11 1_111__1 11111111 111__111 111111_1 11111111   (~matches)
+        //     ________ ________ ______11 _______1 11111111 _____11_ _______1 11111111   = updated result
         result &= ~matches
         // ... and then bitwise-and with a mask that only has the top bit set:
-        //     1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______
-        //     ________ ________ ________ ________ 1_______ ________ ________ 1_______
+        //     ________ ________ ______11 _______1 11111111 _____11_ _______1 11111111   (result)
+        //     1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______   (highBitInEachByte)
+        //     ________ ________ ________ ________ 1_______ ________ ________ 1_______   = updated result
         //
         // This produces all zeroes for the locations where the byte didn't match the search pattern and `0x80`
         // for the locations that did match the search pattern exactly.


d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -0,0 +1,1296 @@
...
+            
+            var values: [Value] = []
+            _skipWhitespace()
+            
+            while pointer.nextByte != .init(ascii: "]") {
+                values.append(try decode(Value.self))
+                path.incrementCurrentIndex()
+                
+                _skipWhitespace()
+                if pointer.nextByte == .init(ascii: ",") {
```
QuietMisdreavus, line 409: I think your PR description might have been backwards when describing trailing comma semantics - it looks like this decoder is more permissive of trailing commas in arrays than the Foundation decoder :wink:

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -0,0 +1,1296 @@
...
+        }
+        
+        // Check for integer digits
+        while pointer.load(fromByteOffset: length, as: UInt8.self).isAsciiDigit {
+            length &+= 1
+        }
+        
+        // Check for a floating point number
+        let isFloatingPoint = pointer.load(fromByteOffset: length, as: UInt8.self) == .init(ascii: ".")
+        if isNegative {
```
QuietMisdreavus: Looks like a typo here - shouldn’t this be `if isFloatingPoint` instead?

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -0,0 +1,1296 @@
...
+            }
+            // Check for exponent digits
+            while pointer.load(fromByteOffset: length, as: UInt8.self).isAsciiDigit {
+                length &+= 1
+            }
+        }
+        
+        // Verify that the scanned number isn't empty and isn't out of bounds
+        guard 0 < length else {
+            if endOfData <= pointer.advanced(by: length) {
```
QuietMisdreavus: It seems weird to me that the EOF check is happening inside the zero-length branch. Do we want to bounds-check in other situations too?

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -0,0 +1,1296 @@
...
+        //
+        // This produces all zeros for the bytes that match the search pattern exactly, and garbage elsewhere.
+        let matches = bytes ^ searchPattern
+        // Next, we wrapping subtract a bit pattern where each byte only has the lowest bit set:
+        //     _1__1__1 _1__1_11 _1__11__ _1___11_ ________ ___11___ ______1_ ________
+        //     _______1 _______1 _______1 _______1 _______1 _______1 _______1 _______1
+        //   = _1__1___ _1__1_1_ _1__1_11 _1___1_1 11111111 ___1_11_ _______1 11111111
+        //
+        // This produces all ones for the bytes that match the search pattern exactly, and garbage elsewhere.
+        var result = matches &- Self.lowBitInEachByte
```
QuietMisdreavus, line 1118: Tangent: I was momentarily confused about what this step would do if there were multiple instances of the target byte in a row, since the subtraction would give the succeeding bytes `0xFE` instead of `0xFF`. However, this is moot because of the following step, since `0xFF & 0xFE` still has its high-bit set. 😅

d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -0,0 +1,1296 @@
...
+                // An even number of slashes means that it's the _slashes_ that are escaped, not the quotation mark.
+                // For example, consider a string that ends in 4 slashes. The 1st slash escapes the 2nd and the 3rd escapes the 4th, leaving the quotation mark unescaped:
+                //
+                //     "ABC\\\\"
+                //         ├╯├╯╰─╴string delimiter
+                //         │ ╰───╴escaped slash
+                //         ╰─────╴escaped slash
+                break
+            } else {
+                // An off number of slashes means that the last slash escapes the quote, so this byte is part of the string's content.
```
QuietMisdreavus:
```suggestion
                // An odd number of slashes means that the last slash escapes the quote, so this byte is part of the string's content.
```

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: AFAICT both this decoder and `JSONDecoder` are able to to decode `[1, 2, 3, ]` but this decoder fails to decode a semantic version from `{ "major": 1, }` whereas `JSONDecoder` decodes that successfully. According to the JSON spec both of those examples with trailing commas isn't valid JSON so this parser is slightly too permissive with trailing command but not quite as permissive as `JSONDecoder` is.

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: I don't think it's right that there would be `0xFE` anywhere or that the succeeding bytes would interfere with each other in any way. My understanding is that the subtraction produces `0xFF` for all matches, even if they're multiple in a series, because there's a `0x01` in each byte. 


If I repeat the same illustration but the next 8 bytes are `kind\"":`

```
       k        i        n        d        \        "        "        :
      6b       69       6e       64       5c       22       22       3a
_11_1_11 _11_1__1 _11_111_ _11__1__ _1_111__ __1___1_ __1___1_ __111_1_
```

and look for the same `"` character

```
       "        "        "        "        "        "        "        "
      22       22       22       22       22       22       22       22
__1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_
```

then the rest of the illustration looks like this:

```
let matches = bytes ^ searchPattern
    _11_1_11 _11_1__1 _11_111_ _11__1__ _1_111__ __1___1_ __1___1_ __111_1_  (bytes)
    __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_  (searchPattern)
  = _1__1__1 _1__1_11 _1__11__ _1___11_ _111111_ ________ ________ ___11___

var result = matches &- lowBitInEachByte
    _1__1__1 _1__1_11 _1__11__ _1___11_ _111111_ ________ ________ ___11___  (matches)
    _______1 _______1 _______1 _______1 _______1 _______1 _______1 _______1  (lowBitInEachByte)
  = _1__1___ _1__1_1_ _1__1_11 _1___1_1 _11111_1 11111111 1111111_ ___1_11_

~matches
    _1__1__1 _1__1_11 _1__11__ _1___11_ _111111_ ________ ________ ___11___  (matches)
  = 1_11_11_ 1_11_1__ 1_11__11 1_111__1 1______1 11111111 11111111 111__111

result &= ~matches
    _1__1___ _1__1_1_ _1__1_11 _1___1_1 _11111_1 11111111 1111111_ ___1_11_  (result)
    1_11_11_ 1_11_1__ 1_11__11 1_111__1 1______1 11111111 11111111 111__111  (~matches)
  = ________ ________ ______11 _______1 _______1 11111111 1111111_ _____11_

result &= highBitInEachByte
    ________ ________ ______11 _______1 _______1 11111111 1111111_ _____11_  (result)
    1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______  (highBitInEachByte)
  = ________ ________ ________ ________ ________ 1_______ 1_______ ________
```

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: Yes, good catch. There must not have been any floating point `SymbolGraph.AnyScalar` values (or negative `SymbolGraph.AnyScalar` values) anywhere in any of the test fixtures, so this code path was never missed or incorrectly taken.

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: Yes. I think this is left from an earlier iteration where this function also skipped whitespace which technically means that even if it didn't find a number if could have gone out of bounds. However, it doesn't do that now, and is non-mutating so it's strange that it does the bounds check when it has already pushed the responsibility to call `pointer.removeFirst(length)` on the caller. It makes sense that it would also push the bounds checking responsibility on the caller (since these are all private functions, so every caller is known and is fine to know the implementation details of the other private functions its calling). 

I realize in explaining this that I should write that down in the code comments. (added in https://github.com/swiftlang/swift-docc/pull/1420/commits/8b0b33c0421473bd68de045afcccf87a8c1e73e6)

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: fixed in https://github.com/swiftlang/swift-docc/pull/1420/commits/10a0e54f4e25bc2175558d1b4f88a94c2cbd7301

d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: I realized that having the labels next to each series if bits was useful so I added that to the code comments in https://github.com/swiftlang/swift-docc/pull/1420/commits/be2c5b189320ca572c55f9d80adfae47f22074c0

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: Additionally, this specific redundant looking bounds check is so that this custom decoder emits the same error as JSONDecoder when the JSON data ends unexpectedly and there's only whitespace left while trying to decode a value (specifically a number in this case). This is also described in the new code comments.

d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
QuietMisdreavus: Ah, right, trailing commas are also a problem in objects. That makes sense, thanks!

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
QuietMisdreavus: I wrote some helper code in a REPL to show off what i was thinking of. Here's the results with your samples, with what the REPL output:

```
// `pretty(_:)` is a function to render a UInt64 into a string like in your comments
 15> let lowBitsInEachByte: UInt64 = 0x01_01_01_01__01_01_01_01
lowBitsInEachByte: UInt64 = 72340172838076673
 16> pretty(lowBitsInEachByte)
$R0: String = "_______1 _______1 _______1 _______1 _______1 _______1 _______1 _______1"
 17> let bytes: UInt64 = 0x6B_69_6E_64_5C_22_22_3A
bytes: UInt64 = 7739838811934433850
// pretty(bytes) = "_11_1_11 _11_1__1 _11_111_ _11__1__ _1_111__ __1___1_ __1___1_ __111_1_"
 18> let searchPattern: UInt64 = 0x22_22_22_22__22_22_22_22
searchPattern: UInt64 = 2459565876494606882
// pretty(searchPattern) = "__1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_ __1___1_"
 19> let matches = bytes ^ searchPattern
matches: UInt64 = 5281398853667389464
 20> pretty(matches)
$R1: String = "_1__1__1 _1__1_11 _1__11__ _1___11_ _111111_ ________ ________ ___11___"
 21> var result = matches &- lowBitsInEachByte 
result: UInt64 = 5209058680829312791
 22> pretty(result)
$R2: String = "_1__1___ _1__1_1_ _1__1_11 _1___1_1 _11111__ 1111111_ 11111111 ___1_111"
 23> String(result, radix: 16, uppercase: true)
$R3: String = "484A4B457CFEFF17"
```

The `0xFE` i was concerned about is the third-least-significant byte in `result` there. However, if you continue with the algorithm...

```
 26> ~matches
$R6: UInt64 = 13165345220042162151
 27> pretty(~matches)
$R7: String = "1_11_11_ 1_11_1__ 1_11__11 1_111__1 1______1 11111111 11111111 111__111"
 28> result &= ~matches
 29> pretty(result)
$R8: String = "________ ________ ______11 _______1 ________ 1111111_ 11111111 _____111"
 30> let highBitInEachByte: UInt64 = 0x80_80_80_80__80_80_80_80
highBitInEachByte: UInt64 = 9259542123273814144
 31> pretty(highBitInEachByte)
$R9: String = "1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______ 1_______"
 32> result &= highBitInEachByte
 33> pretty(result)
$R10: String = "________ ________ ________ ________ ________ 1_______ 1_______ ________"
```

...you still get the matches in the appropriate places, which is why i immediately corrected myself that the concern was moot.

<details>
<summary>aside: the implementation of `pretty(_:)` and its helpers</summary>

Here's the beginning of that REPL session, just for completeness:

```
  1> import Foundation
  2> func padTo64(_ input: String) -> String { String(repeating: "_", count: max(64 - input.count, 0)) + input } 
  3> func render(_ input: UInt64) -> String { padTo64(String(input, radix: 2).replacingOccurrences(of: "0", with: "_")) }
  4> func spaced(_ input: String) -> String { 
  5.   var input = input 
  6.   var output: [Substring] = [] 
  7.   while input.count >= 8 { 
  8.     output.append(input.prefix(8)) 
  9.     input.removeFirst(8) 
 10.   } 
 11.   if !input.isEmpty { output.append(Substring(input)) } 
 12.   return output.joined(separator: " ") 
 13. } 
 14> func pretty(_ input: UInt64) -> String { spaced(render(input)) }
```

</details>

d-ronnqvist: Now I see what you are saying! Like you pointed out; in the end it only matters that the top bit of each matching byte is set so the code still returns the correct values after all the steps.

Examples like [the string in this JSON](https://github.com/swiftlang/swift-docc/pull/1420/changes#diff-7769f27e679875a146ce1c7c48e95e51885dc271f1fc46fcc8eba1e22c6e400eR87) helps verify that the code does the right thing in exactly this situation.

```json
{
  "text" : "multi-statement closures, producing conditional content for the \"then\""
}
```

PR #1421, opened 2026-01-22
build: add missing dependency in CMake build

speednoisemovement:
Adds SymbolKit as a dep to DocCCommon.

(Also needs https://github.com/swiftlang/swift-docc/pull/1422 to fully un-break)

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
@@ -336,20 +336,17 @@ public struct ConvertAction: AsyncAction {
             }
         }
         
-        let analysisProblems: [Problem]
-        let conversionProblems: [Problem]
         do {
-            conversionProblems = try signposter.withIntervalSignpost("Process") {
-                try ConvertActionConverter.convert(
-                    context: context,
-                    outputConsumer: outputConsumer,
-                    htmlContentConsumer: htmlConsumer,
-                    sourceRepository: sourceRepository,
-                    emitDigest: emitDigest,
-                    documentationCoverageOptions: documentationCoverageOptions
-                )
-            }
-            analysisProblems = context.problems
+            let processInterval = signposter.beginInterval("Process", id: signposter.makeSignpostID())
+            try await ConvertActionConverter.convert(
+                context: context,
+                outputConsumer: outputConsumer,
+                htmlContentConsumer: htmlConsumer,
+                sourceRepository: sourceRepository,
+                emitDigest: emitDigest,
+                documentationCoverageOptions: documentationCoverageOptions
+            )
+            signposter.endInterval("Process", processInterval)
         } catch {
             if emitDigest {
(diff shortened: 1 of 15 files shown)


d-ronnqvist:
> * CMake unconditionally uses the default settings for language mode and upcoming features

I'm not sure I understand what this means. What language mode and settings is it using? 

> * Some SPM targets seem to accidentally(?) drop much of `swiftSettings`
> 
> We can probably refactor the CMake to match, but I was curious if the `swiftSettings` discrepancies are intentional first.

The Swift 6 language mode enabled a number of other settings by default. I thought that "InternalImportsByDefault" was one of them, but apparently it's not for [compatibility reasons](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0409-access-level-on-imports.md#default-import-access-level).

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
The changes themselves are not controversial but they fail to compile.

speednoisemovement:
> > * CMake unconditionally uses the default settings for language mode and upcoming features
> 
> I'm not sure I understand what this means. What language mode and settings is it using?
> 
Language mode 5 and the other settings from `swiftSettings`
https://github.com/swiftlang/swift-docc/blob/main/CMakeLists.txt#L24
https://github.com/swiftlang/swift-docc/blob/main/Package.swift#L15

>  The changes themselves are not controversial but they fail to compile.

Sorry, I just tested CMake, didn't realize adding the `package` specifications would break SPM! Would you be open to explicitly adding `InternalImportsByDefault`?

d-ronnqvist:
> Sorry, I just tested CMake, didn't realize adding the package specifications would break SPM! Would you be open to explicitly adding InternalImportsByDefault?

Yes, I opened this PR that applies the compiler feature flags consistently and fixed the import warnings/errors: https://github.com/swiftlang/swift-docc/pull/1422

d-ronnqvist:
> > > * CMake unconditionally uses the default settings for language mode and upcoming features
> > 
> > 
> > I'm not sure I understand what this means. What language mode and settings is it using?
> 
> Language mode 5 and the other settings from `swiftSettings` 
> https://github.com/swiftlang/swift-docc/blob/main/CMakeLists.txt#L24 
> https://github.com/swiftlang/swift-docc/blob/main/Package.swift#L15
> 

That language mode configuration might cause an issue in the future because the Swift 6 language mode enables certain compiler features by default. Most of that is about stricter concurrency checking but not everything and it's possible that a future code change will compile fine with SPM but not CMake because that is configuring the wrong language mode.

This is not an issue related to this PR, just something worth noting.

speednoisemovement:
> Yes, I opened this PR that applies the compiler feature flags consistently and fixed the import warnings/errors: https://github.com/swiftlang/swift-docc/pull/1422

Thanks! I pared this back to just the CMake change, and will follow-up with a refactor to apply settings per-target in CMake

compnerd:
@swift-ci please test

speednoisemovement:
> As far as I understand it doesn't fully fix the CMake build but it fixes the current issue in one of the CMakeLists files.

Changed the title, thanks!

compnerd:
@swift-ci please test

PR #1423, opened 2026-01-27
Flatten group marker node children in navigator

snprajwal:
Bug/issue #, if applicable: rdar://165630645

## Summary

The navigator index constructs group marker nodes and nests references as child nodes under the group if the render relationship specifies that references are nested. While this information is relevant to the render JSON itself, the rendering spec for the navigator expects that group marker nodes do not contain any child nodes [1]. This means that with the current navigator index, the child nodes of a group marker nodes are elided when rendering. In practice, this situation is encountered only with non-generated default implementation groups, where the render JSON includes the references as children of the group. This patch updates the navigator index to insert references at the same level as the group marker, allowing the frontend to render these nodes correctly.

[1]: https://github.com/swiftlang/swift-docc-render/blob/0d779421a47163e0c891fd20212e51670f07a003/src/components/Navigator/NavigatorCardItem.vue#L196

## Dependencies

N/A

## Testing

The existing tests have been updated to validate this behaviour correctly. It can also be visually verified:

- Run `docc preview` with the `LegacyBundle_DoNotUseInNewTests.docc` test bundle
- Visit http://localhost:8080/documentation/sidekit/sideprotocol/func()

| `main` | Patched |
|--------|---------|
|Screenshot 2026-01-27 at 12 12 16|Screenshot 2026-01-27 at 12 13 05|

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -797,8 +797,6 @@ extension NavigatorIndex {
             // Process the children
             var children = [Identifier]()
             for (index, child) in renderNode.navigatorChildren(for: traits).enumerated() {
-                let groupIdentifier: Identifier?
-                
                 if let title = child.name {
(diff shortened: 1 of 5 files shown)


snprajwal:
@swift-ci please test

PR #1428, opened 2026-02-03
Add warnings for multiple root pages

abhijeetsharma200:
Bug/issue #1170

  ## Summary

  Adds warnings when DocC detects multiple root pages in documentation inputs. This helps developers who call `docc convert` directly identify unexpected configurations.

  **Three warning scenarios:**
  - Multiple main modules from symbol graphs (`org.swift.docc.MultipleMainModules`)
  - `@TechnologyRoot` directive when symbols already provide a root (`org.swift.docc.TechnologyRootWithSymbols`)
  - Multiple `@TechnologyRoot` directives in article-only catalogs (`org.swift.docc.MultipleTechnologyRoots`)

  Warnings include source locations and fix-it suggestions where applicable. The logic uses mutually exclusive checks to avoid duplicate warnings for the same issue.

  ## Dependencies

  None.

  ## Testing

  Steps:
  1. Run `swift test --filter DocumentationContext_RootPageTests`
  2. All 11 tests should pass, including 6 new tests covering the warning scenarios

  New test cases:
  - `testWarnsAboutMultipleTechnologyRootDirectives` - Multiple `@TechnologyRoot` without symbols
  - `testWarnsAboutTechnologyRootWithSymbols` - `@TechnologyRoot` with symbol graphs
  - `testWarnsAboutMultipleTechnologyRootsWithSymbols` - Verifies mutually exclusive warning logic
  - `testWarnsAboutMultipleMainModules` - Multiple modules from symbol graphs
  - `testNoWarningForSingleModule` / `testNoWarningForSingleTechnologyRoot` - No false positives

  ## Checklist

  - [x] Added tests
  - [x] Updated documentation if necessary (N/A - no doc changes needed)

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -9,8 +9,13 @@
 */
 
 public import Foundation
-import Markdown
+private import Markdown
+public import DocCCommon
 import SymbolKit
+
+#if canImport(os)
+private import os
+#endif
 
 /// The documentation context manages the in-memory model for the built documentation.
 ///
@@ -970,7 +975,9 @@ public class DocumentationContext {
                 ))
             }
         }
-        results.forEach { addPreparedSymbolToContext($0) }
+        for result in results {
+            addPreparedSymbolToContext(result)
+        }
     }
 
     /// Adds a prepared symbol data including a topic graph node and documentation node to the context.
@@ -1418,7 +1425,7 @@ public class DocumentationContext {
         }
         
         // Merge in all the dictionary keys for each target into their section variants.
-        keysByTarget.forEach { targetIdentifier, keys in
+        for (targetIdentifier, keys) in keysByTarget {
             let target = documentationCache[targetIdentifier]
             if let semantic = target?.semantic as? Symbol {
                 let keys = keys.sorted(by: \.name)
@@ -1431,7 +1438,7 @@ public class DocumentationContext {
         }
         
         // Merge in all the parameters for each target into their section variants.
-        parametersByTarget.forEach { targetIdentifier, parameters in
+        for (targetIdentifier, parameters) in parametersByTarget {
             let target = documentationCache[targetIdentifier]
             if let semantic = target?.semantic as? Symbol {
                 let parameters = parameters.sorted(by: \.name)
@@ -1444,7 +1451,7 @@ public class DocumentationContext {
         }
         
         // Merge in the body for each target into their section variants.
-        bodyByTarget.forEach { targetIdentifier, body in
+        for (targetIdentifier, body) in bodyByTarget {
             let target = documentationCache[targetIdentifier]
             if let semantic = target?.semantic as? Symbol {
                 // Add any body parameters to existing body record
@@ -1461,7 +1468,7 @@ public class DocumentationContext {
         }
         
         // Merge in all the responses for each target into their section variants.
-        responsesByTarget.forEach { targetIdentifier, responses in
+        for (targetIdentifier, responses) in responsesByTarget {
             let target = documentationCache[targetIdentifier]
             if let semantic = target?.semantic as? Symbol {
                 let responses = responses.sorted(by: \.statusCode)
@@ -1579,7 +1586,9 @@ public class DocumentationContext {
         }
 
         // Add any new symbols to the documentation cache.
-        results.forEach { addPreparedSymbolToContext($0) }
+        for result in results {
+            addPreparedSymbolToContext(result)
+        }
     }
     
     private static let supportedImageExtensions: Set<String> = ["png", "jpg", "jpeg", "svg", "gif"]
@@ -1712,7 +1721,20 @@ public class DocumentationContext {
     ///   - articles: Articles to register with the documentation cache.
     /// - Returns: The articles that were registered, with their topic graph node updated to what's been added to the topic graph.
     private func registerArticles(_ articles: DocumentationContext.Articles) -> DocumentationContext.Articles {
-        articles.map { article in
+        // Because of rdar://79745455, articles can collide with top-level symbols.
+        // If the file system is case insensitive, this collision also happens for articles that case _in_sensitive match a top-level symbol's reference.
+        // To help detect this; pre-build a lookup of the lowercase paths of all top-level symbols so that we can match the articles against that.
+        let topLevelSymbolsThatCanCollideWithArticles: [String /* lowercased path */: ResolvedTopicReference] = {
+            let topLevelSymbols = linkResolver.localResolver.topLevelSymbols()
+            var result = [String: ResolvedTopicReference]()
+            result.reserveCapacity(topLevelSymbols.count)
+            for reference in topLevelSymbols {
+                result[reference.path.lowercased()] = reference
+            }
+            return result
+        }()
+                    
+        return articles.compactMap { article in
             guard let (documentation, title) = Self.documentationNodeAndTitle(
                 for: article,
                 // By default, articles are available in the languages the module that's being documented
@@ -1726,7 +1748,9 @@ public class DocumentationContext {
             }
             let reference = documentation.reference
             
-            if let existing = documentationCache[reference], existing.kind.isSymbol {
+            if let existingSymbolReference = topLevelSymbolsThatCanCollideWithArticles[reference.path.lowercased()],
+               let existing = documentationCache[existingSymbolReference]
+            {
                 // By the time we get here it's already to late to fix the collision. All we can do is make the author aware of it and handle the collision deterministically.
                 // rdar://79745455 and https://github.com/swiftlang/swift-docc/issues/593 tracks fixing the root cause of this issue, avoiding the collision and allowing the article and symbol to both exist.
                 diagnosticEngine.emit(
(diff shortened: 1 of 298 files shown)


Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2539,7 +2541,79 @@ public class DocumentationContext {
...
+
+        if mainModuleNames.count > 1 {
+            let sortedModuleNames = mainModuleNames.sorted()
+            let diagnostic = Diagnostic(
+                source: mainSymbolGraphURLs.first,
+                severity: .warning,
+                range: nil,
+                identifier: "org.swift.docc.MultipleMainModules",
+                summary: "The documentation input contains more than one main module: \(sortedModuleNames.map { $0.singleQuoted }.joined(separator: ", "))",
+                explanation: "DocC doesn't support building combined documentation for multiple modules. Each module should be documented separately."
```
heckj:
```suggestion
                explanation: "DocC doesn't support building combined documentation for multiple modules in a single catalog. Each module should be documented separately."
```

Improves the explanation just a smidge - it's not about combined documentation so much as combined in a single catalog.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2539,7 +2541,79 @@ public class DocumentationContext {
...
+            }
+
+            let problems = rootPageArticles.map { article -> Problem in
+                let diagnostic = Diagnostic(
+                    source: article.source,
+                    severity: .warning,
+                    range: article.value.metadata?.technologyRoot?.originalMarkup.range,
+                    identifier: "org.swift.docc.TechnologyRootWithSymbols",
+                    summary: "The '\(TechnologyRoot.directiveName)' directive creates an additional root page, but the documentation already has a root from its symbols",
+                    explanation: "When documentation contains symbols (from symbol graph files), those symbols provide a module root. The '\(TechnologyRoot.directiveName)' directive creates an additional root page which results in unexpected documentation structure.",
```
heckj:
```suggestion
                    explanation: "When documentation contains symbols from source files or libraries, those symbols provide a module root. The '\(TechnologyRoot.directiveName)' directive creates an additional root page which results in an unexpected documentation structure.",
```

I'm not sure the end users will be all that aware of symbol graph files, so switching it up to where they're sourced from by DocC "behind the scenes".

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2539,7 +2541,79 @@ public class DocumentationContext {
...
+
+        if mainModuleNames.isEmpty && rootPageArticles.count > 1 {
+            let problems = rootPageArticles.map { article -> Problem in
+                let diagnostic = Diagnostic(
+                    source: article.source,
+                    severity: .warning,
+                    range: article.value.metadata?.technologyRoot?.originalMarkup.range,
+                    identifier: "org.swift.docc.MultipleTechnologyRoots",
+                    summary: "The documentation has multiple articles with the '\(TechnologyRoot.directiveName)' directive",
+                    explanation: "Only one article should use the '\(TechnologyRoot.directiveName)' directive to define the documentation's root page. Having multiple root pages results in unexpected documentation structure."
```
heckj:
```suggestion
                    explanation: "Only one article should use the '\(TechnologyRoot.directiveName)' directive to define the documentation's root page. Having multiple root pages results in an unexpected documentation structure."
```

abhijeetsharma200:
Thanks for the review. Just a quick clarification. I was going through the coding guidelines and saw that swift testing is to be used. I am in the middle of refactoring my code to match that guideline. Is that necessary, or is the code fine as it is?

Other than that, I will make the changes as suggested :)

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
abhijeetsharma200: Done!

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
abhijeetsharma200: Done!

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
abhijeetsharma200: Done!

abhijeetsharma200:
Done and switched to swift testing

abhijeetsharma200:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2539,7 +2541,79 @@ public class DocumentationContext {
             diagnosticEngine.emit(Problem(diagnostic: Diagnostic(source: articleResult.source, severity: .information, range: nil, identifier: "org.swift.docc.ArticleUncurated", summary: "You haven't curated \(articleResult.topicGraphNode.reference.description.singleQuoted)"), possibleSolutions: []))
         }
     }
-    
+
+    /// Emits warnings when the documentation contains multiple root pages.
+    private func emitWarningsForMultipleRootPages(rootPageArticles: [SemanticResult<Article>]) {
+        let mainSymbolGraphURLs = inputs.symbolGraphURLs.filter { !$0.lastPathComponent.contains("@") }
+        let mainModuleNames = Set(mainSymbolGraphURLs.compactMap { SymbolGraphLoader.moduleNameFor($0) })
```
d-ronnqvist: This isn't a reliable way to find the unique module names. 

It produces false positives both when there are multiple platform specific symbol graphs for a single module (for example `testFunctionWithDifferentSignaturesOnDifferentPlatforms`) and when there are multiple source language specific symbol graphs for a single module (for example `testLanguageSpecificTopicSectionDoesNotAppearInAutomaticSeeAlso`, `testResolvesAlternateDeclarations`, and `testParametersWithExternalLink`).

---

I haven't run this but I think that you might be able to get all the module names like this:
```suggestion
        let moduleNames = linkResolver.localResolver.modules().map { reference in
            documentationCache[reference]?.name.plainText
        }
```

If that doesn't work then this might be an alternative. Note that `rootModules` contains both symbol roots and authored technology roots (which is why the `compactMap` does a `isSymbol` check.

```suggestion
        let mainModuleNames: Set<String> = Set(rootModules.compactMap { reference in
            guard let node = documentationCache[reference], node.kind.isSymbol else { return nil }
            return node.name.plainText
        })
```

d-ronnqvist: Note; it's _possible_ that neither of those suggestions work correctly when there's snippets involved, so it's possible that we need to identify and exclude snippets specifically.

d-ronnqvist (changes requested):
This generally looks good but there's 22 tests that fail with these changes. 

---

A handful of those failures are because of false positives in identifying the unique main module names.

The various failures in SnippetResolverTests verifies behavior when there are both a main module symbol graph file and a symbol graph file of snippets. This is supported and shouldn't raise any warnings.

Both `testModuleUnderTechnologyRoot` and `testModuleUnderAncestorOfTechnologyRoot` should probably be updated to expect these warnings.

Any test involving "LegacyBundle_DoNotUseInNewTests" will probably have one or more of these errors. They might be hard to deal with, so you could do two things:
- update the test to expect the new warnings
- add a `// FIXME:` comment right above the test expectations for those new warning, saying that we should update the test to not use "LegacyBundle_DoNotUseInNewTests" anymore.

If there are any other tests that you don't know what to do with; don't hesitate to @-mention me in this PR and ask.

abhijeetsharma200:
Thanks! You’re right, filtering `inputs.symbolGraphURLs` and extracting module names from filenames isn’t robust. It produces false positives for multi-platform/multi-language symbol graphs and when snippet symbol graphs are present.

I’ve updated the logic to compute the set of “real” module roots from `rootModules` by:

- only counting references that are top-level module roots (`/documentation/<ModuleName>`)
- requiring a real module node in the graph/cache (`docNode.kind == .module`)
- excluding virtual nodes (`!node.isVirtual`) so snippets don’t trigger warnings

This should align with your suggestion to use resolved roots instead of symbol graph URLs, while avoiding snippet-related false positives. I also set `source: nil` for `MultipleMainModules` since it isn’t attributable to a single file.

If you notice cases where module roots aren’t always exactly `/documentation/<ModuleName>` (or if there’s a better snippet/root discriminator than `isVirtual`), let me know and I’ll adjust.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -2539,7 +2541,91 @@ public class DocumentationContext {
...
+    private func emitWarningsForMultipleRootPages(rootPageArticles: [SemanticResult<Article>]) {
+        // Get module names from symbol sources (not from @TechnologyRoot articles).
+        // Filter out virtual modules (like snippets) which shouldn't count as separate roots.
+        let symbolModuleNames: Set<String> = Set(rootModules.compactMap { reference in
+            // Top-level modules have paths like /documentation/ModuleName (2 components).
+            // The path check is needed because `rootModules` may include nested symbols that
+            // happen to have kind `.module` (e.g., when testing all symbol kinds). Without this
+            // check, those nested symbols would incorrectly be counted as separate modules.
+            let pathComponents = reference.path.split(separator: "/")
+            guard pathComponents.count == 2, pathComponents.first == "documentation" else { return nil }
```
d-ronnqvist: It's not good for the actual code to make workaround to accommodate a test. It would be better to update the test to not do the strange behavior in the first place.

However, in this case I think that it'd be better to use a different source for this data. AFAICT the modules from the link resolver data has already filtered out the snippets and isn't sensitive to the SGFs containing symbols that themselves have a `.module` kind.

The main thing to note is that modules from the link resolver already combines both the symbol and non-symbol roots into a single list. You could either filter out the non-symbols, and only user this data as the list of symbol roots, or you could use this source for both kinds of roots and not pass a `rootPageArticles`.

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

abhijeetsharma200:
Thanks for the detailed reviews! Learned a lot. Thanks for your time!

PR #1426, opened 2026-02-02
[DNM] Empty PR to see if CI is failing on main

d-ronnqvist:
This empty PR tests if the Linux CI fails on main

PR #1429, opened 2026-02-03
Unblock DocC Linux CI

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This unblocks the DocC in a bit of an unusual way; by replacing each use of `forEach` with a regular for-in loop, and by additionally updating a few tests to use parametrized tests in Swift testing rather than looping over inputs in the test body.

There's already a PR open for the compiler fix, so if we just wait a few days there's likely to be a new development toolchain with the fix. 

That said; both myself and others I spoke to felt that the for-in code was easier to read. And if it additionally unblocks out CI a couple days sooner, one of those is a nice added bonus to the other (not sure which is more important). 

## Dependencies

None

## Testing

Verify that the Linux CI passes.
Otherwise, nothing in particular. This isn't a user-facing change.

## Checklist

- [x] ~Added~ Updated tests

### DIFF
--- a/Sources/SwiftDocC/Model/DocumentationMarkup.swift
+++ b/Sources/SwiftDocC/Model/DocumentationMarkup.swift
@@ -127,7 +127,7 @@ struct DocumentationMarkup {
             }
             let isLastChild = index == (markup.childCount - 1)
             
-            // Already parsed all expected content, return.
+            // Already parsed all expected content.
             guard currentSection != .end else {
                 continue
             }
--- a/Tests/SwiftDocCTests/Infrastructure/SymbolGraph/ExtendedTypesFormatTransformationTests.swift
+++ b/Tests/SwiftDocCTests/Infrastructure/SymbolGraph/ExtendedTypesFormatTransformationTests.swift
@@ -44,7 +44,7 @@ class ExtendedTypesFormatTransformationTests: XCTestCase {
         XCTAssertEqual(addedMemberSymbolsTypeB.count, 2)
         
         // check the symbols are connected as expected
-        for relationShip in [
+        for relationship in [
             SymbolGraph.Relationship(source: addedMemberSymbolsTypeA[0].identifier.precise, target: extendedTypeA.identifier.precise, kind: .memberOf, targetFallback: nil),
             SymbolGraph.Relationship(source: addedMemberSymbolsTypeA[1].identifier.precise, target: extendedTypeA.identifier.precise, kind: .memberOf, targetFallback: nil),
             SymbolGraph.Relationship(source: addedMemberSymbolsTypeATwo[0].identifier.precise, target: extendedTypeATwo.identifier.precise, kind: .memberOf, targetFallback: nil),
@@ -57,7 +57,7 @@ class ExtendedTypesFormatTransformationTests: XCTestCase {
             SymbolGraph.Relationship(source: extendedTypeB.identifier.precise, target: extendedModuleA.identifier.precise, kind: .declaredIn, targetFallback: nil),
         ] {
             XCTAssert(graph.relationships.contains(where: { sample in
-                sample.source == relationShip.source && sample.target == relationShip.target && sample.kind == relationShip.kind
+                sample.source == relationship.source && sample.target == relationship.target && sample.kind == relationship.kind
             }))
         }
         
@@ -66,8 +66,7 @@ class ExtendedTypesFormatTransformationTests: XCTestCase {
         XCTAssertEqual(graph.relationships.count, 3 /* .declaredIn */ + 6 /* .memberOf */)
         
         // check correct module name was prepended to pathComponents
-        for symbol in ([extendedModuleA, extendedTypeA, extendedTypeATwo, extendedTypeB] + addedMemberSymbolsTypeA + addedMemberSymbolsTypeATwo)
-        {
+        for symbol in ([extendedModuleA, extendedTypeA, extendedTypeATwo, extendedTypeB] + addedMemberSymbolsTypeA + addedMemberSymbolsTypeATwo) {
             XCTAssertEqual(symbol.pathComponents.first, "A")
         }
     }


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
I reverted https://github.com/swiftlang/swift-docc/pull/1426 back to be an empty commit on top of the latest main branch and you can see that [the CI using swift-DEVELOPMENT-SNAPSHOT-2026-02-02-a-ubuntu22.04 failed](https://ci.swift.org/job/pr-swift-docc-linux/2237/consoleText) but for this branch the CI using the same development snapshot passes.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
```diff
@@ -255,9 +252,8 @@ struct SymbolGraphLoader {
                         }
                     }
                     // Add the missing default platform availability.
-                    missingAvailabilities.forEach { missingAvailability in
-                        if !symbolAvailability.contains(missingAvailability.platformName) {
-                            guard let defaultAvailability = AvailabilityItem(missingAvailability) else { return }
+                    for missingAvailability in missingAvailabilities where !symbolAvailability.contains(missingAvailability.platformName) {
```
patshaughnessy, line 255: I wonder if behind the scenes the Swift compiler implements `where` clauses like this using an if statement like what you replaced here?

Sources/SwiftDocC/Model/DocumentationMarkup.swift
```diff
@@ -120,15 +120,17 @@ struct DocumentationMarkup {
...
         
         // Index all headings as a lookup during parsing the content
-        markup.children.enumerated().forEach({ pair in
+        for (index, child) in markup.children.enumerated() {
             // If we've parsed the last section we're interested in, skip through the rest
-            guard currentSection <= parseUpToSection || currentSection == .end else { return }
-            
-            let (index, child) = pair
+            guard currentSection <= parseUpToSection || currentSection == .end else {
+                continue
```
patshaughnessy: Should this be `return` ?

Sources/SwiftDocC/Model/DocumentationMarkup.swift
```diff
@@ -120,15 +120,17 @@ struct DocumentationMarkup {
...
+            }
             let isLastChild = index == (markup.childCount - 1)
             
             // Already parsed all expected content, return.
-            guard currentSection != .end else { return }
+            guard currentSection != .end else {
+                continue
```
patshaughnessy, line 132: Let's update the preceding comment also:

> Already parsed all expected content, continue.

Sources/SwiftDocC/Semantics/ExternalLinks/ExternalReferenceWalker.swift
```diff
@@ -66,12 +59,16 @@ struct ExternalReferenceWalker: SemanticVisitor {
         
     mutating func visitTutorialSection(_ tutorialSection: TutorialSection) {
         visitMarkupLayouts(tutorialSection.introduction)
-        tutorialSection.stepsContent.unwrap { visitSteps($0) }
```
patshaughnessy, line 69: Is this related to removing uses of `forEach` ? Or do we need this for some other reason?

...and a few more times following...

Tests/SwiftDocCTests/Infrastructure/SymbolGraph/ExtendedTypesFormatTransformationTests.swift
```diff
@@ -44,7 +44,7 @@ class ExtendedTypesFormatTransformationTests: XCTestCase {
         XCTAssertEqual(addedMemberSymbolsTypeB.count, 2)
         
         // check the symbols are connected as expected
-        [
+        for relationShip in [
```
patshaughnessy: Typo? `relationship`

Tests/SwiftDocCTests/Model/SemaToRenderNodeTests.swift
```diff
@@ -651,8 +651,8 @@ class SemaToRenderNodeTests: XCTestCase {
         XCTAssertFalse(firstTutorialReference.abstract.isEmpty)
         XCTAssertEqual(firstTutorialReference.estimatedTime, "20min")
         
-        renderNode.references.compactMap { $0.value as? TopicRenderReference } .forEach {
-            XCTAssertFalse($0.abstract.isEmpty)
+        for case let renderReference as TopicRenderReference in renderNode.references.values {
```
patshaughnessy, line 654: Nice simplification

Tests/SwiftDocCTests/Utility/ValidatedURLTests.swift
```diff
@@ -8,85 +8,65 @@
...
-            let validated = ValidatedURL(parsingExact: url.absoluteString)
-            XCTAssertEqual(url.absoluteString, validated?.absoluteString)
-        }
-
-        // Test ValidatedURL.init(URL)
-        validURLs.forEach { url in
-            let validated = ValidatedURL(url)
-            XCTAssertEqual(url.absoluteString, validated?.absoluteString)
-        }
+    @Test(arguments: [
```
patshaughnessy, line 17: Much cleaner with Swift testing 👍

Sources/SwiftDocC/Semantics/ExternalLinks/ExternalReferenceWalker.swift
d-ronnqvist: It's unrelated. However, when I looked at the code with only the for loops updated I felt that the `unwrap` helper functions stuck out and felt that the code would be easier to read if it _consistently_ used standard language constructs like if-statements and for-loops, so I updated this as for code readability.

Sources/SwiftDocC/Infrastructure/Symbol Graph/SymbolGraphLoader.swift
d-ronnqvist: Yes. A for loop with a where clause is like having a `guard condition else { continue }` at the start of the loop body. 

This specific original loop had two nested conditions (one if-statement and one guard-statement) so I moved one of them to a where clause to avoid the nested conditions. I could have also changed the two conditions to either both use if statements or both use guard statements so that they could be expressed as two conditions in the same if or guard statement.

d-ronnqvist:
@swift-ci please test

Tests/SwiftDocCTests/Infrastructure/SymbolGraph/ExtendedTypesFormatTransformationTests.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1429/commits/b6f47b077275a7eb8617b9a3a7a53a1fa4525b5d

Sources/SwiftDocC/Model/DocumentationMarkup.swift
d-ronnqvist: Updated in https://github.com/swiftlang/swift-docc/pull/1429/commits/b6f47b077275a7eb8617b9a3a7a53a1fa4525b5d

PR #1430, opened 2026-02-04
Improve article symbol collision detection

d-ronnqvist:
Bug/issue #, if applicable: rdar://169608711

## Summary

This improves the detection of articles that would override a symbol page to also consider case _in_sensitive reference matches.

## Dependencies

None

## Testing

In any project with symbol documentation:
- Create an article with the same file name as one of the top-level symbols but with different capitalization.
- Build documentation
  - DocC should warn about the article having a name that would override the symbol.
  - DocC should drop the article in favor of the symbol, to produce a deterministic output.

## Checklist

- [x] ~Added~ Updated tests

### DIFF
--- a/Tests/SwiftDocCTests/Model/SemaToRenderNodeMultiLanguageTests.swift
+++ b/Tests/SwiftDocCTests/Model/SemaToRenderNodeMultiLanguageTests.swift
@@ -508,7 +508,7 @@ class SemaToRenderNodeMixedLanguageTests: XCTestCase {
             
             An article in a mixed-language framework. This symbol link should display the correct title depending on \
             the language we're browsing this article in: ``MixedLanguageFramework/Bar/myStringFunction(_:)``.
-            """.write(to: url.appendingPathComponent("bar.md"), atomically: true, encoding: .utf8)
+            """.write(to: url.appendingPathComponent("my-article.md"), atomically: true, encoding: .utf8)
         }
         
         let articleRenderNode = try outputConsumer.renderNode(withTitle: "MyArticle")


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -1716,7 +1716,20 @@ public class DocumentationContext {
...
+        // To help detect this; pre-build a lookup of the lowercase paths of all top-level symbols so that we can match the articles against that.
+        let topLevelSymbolsThatCanCollideWithArticles: [String /* lowercased path */: ResolvedTopicReference] = {
+            let topLevelSymbols = linkResolver.localResolver.topLevelSymbols()
+            var result = [String: ResolvedTopicReference]()
+            result.reserveCapacity(topLevelSymbols.count)
+            for reference in topLevelSymbols {
+                result[reference.path.lowercased()] = reference
+            }
+            return result
+        }()
```
snprajwal, line 1730: Since the references to top level symbols are unique, we can also directly use [`init(uniqueKeysWithValues:)`](https://developer.apple.com/documentation/swift/dictionary/init(uniquekeyswithvalues:))
```suggestion
        let topLevelSymbolsThatCanCollideWithArticles = 
            Dictionary(
                uniqueKeysWithValues: linkResolver.localResolver.topLevelSymbols()
                .map { ($0.path.lowercased(), $0) }
            )
```

d-ronnqvist: Symbols with multiple language representations can appear more than once in `topLevelSymbols()`. This is a side effect of how each language representation has its own node in the path hierarchy but they both share a common topic reference. 

This might be an unexpected sharp edge of that API, so it might be worth looking into deduping those references in a different future PR.

snprajwal: Makes sense, thanks for explaining.

PR #1432, opened 2026-02-04
CMake: apply language settings per-target

speednoisemovement:
Currently, CMake applies Swift language mode and upcoming feature settings globally for the whole build, but the SPM build applies this per-target. This can cause language mode skew between the two.

This adds a per-target setting to mirror the SPM setup to the CMake build.


speednoisemovement:
> I think that it might be nicer to move the `Swift_LANGUAGE_VERSION` setting to global scope as `CMAKE_Swift_LANGUAGE_VERSION`. But, I _can_ see the argument that it is being explicitly set on each target and the package doesn't want a default. If the latter is desired, perhaps setting the default to an invalid value to ensure that is the best choice.

Done in https://github.com/swiftlang/swift-docc/pull/1432/changes/3adddccaad63adf6e01c15036d48894dafaf3ca2

compnerd:
@swift-ci please test

compnerd:
Merging this for now - this should have no impact on any other platform.

PR #1433, opened 2026-02-05
Add CLI flag to provide custom host for previews

snprajwal:
## Summary

`docc preview` allows providing a custom port via the `--port` flag. The default host used is `localhost`. However, it is also desirable to allow using a custom host in order to serve the preview outside the local machine, e.g. by binding to `0.0.0.0` instead. This patch adds a `--host` flag that accepts a string parameter which is used as the host on the local machine for the HTTP server.

Supersedes #713 with changes to prevent breaking the public API.

## Dependencies

N/A

## Testing

Unit tests and integration tests have been added in to verify this functionality.

## Checklist

- [x] Added tests

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -390,8 +390,11 @@ Starting Local Preview Server
 Monitoring ~/Developer/swift-docc/Sources/SwiftDocC/SwiftDocC.docc for changes...
 ```
 
-And if you navigate to <http://localhost:8080/documentation/swiftdocc> you'll see
-the rendered documentation for `SwiftDocC`.
+And if you navigate to <http://localhost:8080/documentation/swiftdocc> you'll
+see the rendered documentation for `SwiftDocC`. If your browser isn't running
+on the same computer as Swift-DocC, you may need to provide the server's public
+host name or IP address (or just `0.0.0.0`) with the `--host` option to allow
+the browser to connect.
 
 ### Using Docker to Test Swift-DocC for Linux
 
@@ -628,4 +631,4 @@ project's technical documentation:
   with support for building and viewing documentation for your framework and
   its dependencies.
 
-<!-- Copyright (c) 2021-2025 Apple Inc and the Swift Project authors. All Rights Reserved. -->
+<!-- Copyright (c) 2021-2026 Apple Inc and the Swift Project authors. All Rights Reserved. -->
--- a/Sources/DocCCommandLine/ArgumentParsing/Options/PreviewOptions.swift
+++ b/Sources/DocCCommandLine/ArgumentParsing/Options/PreviewOptions.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -17,6 +17,15 @@ import Foundation
 public struct PreviewOptions: ParsableArguments {
     public init() { }
 
+    /// The host name to use for the preview web server.
+    ///
+    /// Defaults to `localhost`.
+    @Option(
+        name: .long,
+        help: ArgumentHelp(
+            "Host name to use for the preview web server.",
+            valueName: "host-name"))
+    public var host: String = "localhost"
     /// The port number to use for the preview web server.
     ///
     /// Defaults to `8080`.
--- a/Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
+++ b/Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -45,16 +45,16 @@ final class PreviewServer {
         /// The server did not find the content directory.
         case pathNotFound(String)
         /// Cannot bind the server to the given address
-        case cannotStartServer(port: Int)
+        case cannotStartServer(host: String, port: Int)
         /// The given port is not available
-        case portNotAvailable(port: Int)
+        case portNotAvailable(host: String, port: Int)
         
         var errorDescription: String {
             switch self {
                 case .failedToStart: return "Failed to start preview server"
                 case .pathNotFound(let path): return "The preview content path '\(path)' is not found"
-                case .cannotStartServer(let port): return "Can't start the preview server on port \(port)"
-                case .portNotAvailable(let port): return "Port \(port) is not available at the moment, "
+                case .cannotStartServer(let host, let port): return "Can't start the preview server on port \(port) on host \(host)"
+                case .portNotAvailable(let host, let port): return "Port \(port) on host \(host) is not available at the moment, "
                     + "try a different port number by adding the option '--port XXXX' "
                     + "to your command invocation where XXXX is the desired (free) port."
             }
@@ -72,15 +72,15 @@ final class PreviewServer {
     /// A list of server-bind destinations.
     public enum Bind: CustomStringConvertible {
         /// A port on the local machine.
-        case localhost(port: Int)
+        case localhost(host: String, port: Int)
         
         /// A file socket on disk.
         case socket(path: String)
         
         var description: String {
             switch self {
-            case .localhost(port: let port):
-                return "localhost:\(port)"
+            case .localhost(host: let host, port: let port):
+                return "\(host):\(port)"
             case .socket(path: let path):
                 return path
             }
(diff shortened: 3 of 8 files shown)


snprajwal:
@swift-ci please test

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
```diff
@@ -72,15 +72,15 @@ final class PreviewServer {
     /// A list of server-bind destinations.
     public enum Bind: CustomStringConvertible {
         /// A port on the local machine.
-        case localhost(port: Int)
+        case localhost(host: String, port: Int)
```
d-ronnqvist, line 75: minor: Is it correct for this case to be named "localhost" when it may represent other host names?

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
```diff
@@ -45,16 +45,16 @@ final class PreviewServer {
...
-        case portNotAvailable(port: Int)
+        case portNotAvailable(host: String, port: Int)
         
         var errorDescription: String {
             switch self {
                 case .failedToStart: return "Failed to start preview server"
                 case .pathNotFound(let path): return "The preview content path '\(path)' is not found"
-                case .cannotStartServer(let port): return "Can't start the preview server on port \(port)"
-                case .portNotAvailable(let port): return "Port \(port) is not available at the moment, "
+                case .cannotStartServer(let host, let port): return "Can't start the preview server on port \(port) on host \(host)"
```
d-ronnqvist, line 56: minor: Is is necessary to say "host localhost" or would it be sufficient to only use the hostname?

Alternatively; would it make sense to combine the hostname and the port number into `\(host):\(port)`?

Sources/DocCCommandLine/ArgumentParsing/Options/PreviewOptions.swift
```diff
@@ -17,6 +17,15 @@ import Foundation
...
 
+    /// The host name to use for the preview web server.
+    ///
+    /// Defaults to `localhost`.
+    @Option(
+        name: .long,
+        help: ArgumentHelp(
+            "Host name to use for the preview web server.",
+            valueName: "host-name"))
+    public var host: String = "localhost"
```
d-ronnqvist, line 28: What about the user-provided hostname do we need to validate? 

Should we check something like `host.unicodeScalars.allSatisfy { CharacterSet.urlHostAllowed.contains($0) }` to raise errors about invalid configuration earlier (and not when the preview server fails to start)?

Tests/DocCCommandLineTests/ArgumentParsing/PreviewSubcommandTests.swift
```diff
@@ -97,6 +97,25 @@ class PreviewSubcommandTests: XCTestCase {
                 testBundleURL.path,
             ]))
         }
+
+        // Test default host is localhost
+        do {
+            setenv(TemplateOption.environmentVariableKey, templateDir.path, 1)
```
d-ronnqvist, line 103: I don't see this test cleaning up the environment variables that it sets.

Also, this test uses "LegacyBundle_DoNotUseInNewTests". Since we're modifying the test, can we figure out what inputs the test actually needs and switch to a minimal test input?

Tests/DocCCommandLineTests/PreviewActionIntegrationTests.swift
```diff
@@ -316,6 +316,79 @@ class PreviewActionIntegrationTests: XCTestCase {
...
+        defer {
+            try? preview.stop()
+        }
+
+        preview.bindServerToSocketPath = try createTemporaryTestSocketPath()
+
+        let logStorage = LogHandle.LogStorage()
+
+        // Start the preview server and verify the custom host is used in the log output
+        do {
```
d-ronnqvist, line 358: nit: this `do` statement doesn't catch any errors and there's no scope after it ends that it needs to scope variable out of.

Tests/DocCCommandLineTests/PreviewActionIntegrationTests.swift
```diff
@@ -316,6 +316,79 @@ class PreviewActionIntegrationTests: XCTestCase {
         #endif
     }
     
+    func testCustomHostParameter() async throws {
```
d-ronnqvist, line 319: question: should we try to write this test using Swift Testing?

CONTRIBUTING.md
```diff
@@ -390,8 +390,11 @@ Starting Local Preview Server
 Monitoring ~/Developer/swift-docc/Sources/SwiftDocC/SwiftDocC.docc for changes...
 ```
 
-And if you navigate to <http://localhost:8080/documentation/swiftdocc> you'll see
-the rendered documentation for `SwiftDocC`.
+And if you navigate to <http://localhost:8080/documentation/swiftdocc> you'll
+see the rendered documentation for `SwiftDocC`. If your browser isn't running
+on the same computer as Swift-DocC, you may need to provide the server's public
+host name or IP address (or just `0.0.0.0`) with the `--host` option to allow
+the browser to connect.
```
d-ronnqvist, line 397: I would argue that text shouldn't be wrapped at any arbitrary character count but rather be wrapped at semantically meaningful boundaries (mainly punctuation).

By wrapping the text like this it's unnecessarily causing a diff in the first sentence which doesn't have any changes (except that for some reason you moved the word "see" from the first line to the second line).

For this I would almost say that the newly added documentation is such an edge case that it could be explained in its own paragraph or maybe an aside.

Tests/DocCCommandLineTests/PreviewActionIntegrationTests.swift
```diff
@@ -316,6 +316,79 @@ class PreviewActionIntegrationTests: XCTestCase {
...
+        let createConvertAction = {
+            try ConvertAction(
+                documentationBundleURL: sourceURL,
+                outOfProcessResolver: nil,
+                analyze: false,
+                targetDirectory: outputURL,
+                htmlTemplateDirectory: templateURL,
+                emitDigest: false,
+                currentPlatforms: nil,
+                fileManager: FileManager.default,
```
d-ronnqvist, line 338: question: I haven't read this code for a quite a while but is there a reason why it needs to read and write real files?

Tests/DocCCommandLineTests/PreviewActionIntegrationTests.swift
```diff
@@ -316,6 +316,79 @@ class PreviewActionIntegrationTests: XCTestCase {
...
+                temporaryDirectory: convertActionTempDirectory)
+        }
+
+        // Create a preview action with a custom host
+        let customHost = "0.0.0.0"
+        let preview = try PreviewAction(
+            host: customHost,
+            port: 8080, // We ignore this value when we set the `bindServerToSocketPath` property below.
```
d-ronnqvist, line 346: question: I haven't read this code for a quite a while but it seems strange to me that we would create a preview action with one configuration and then override it with something else and then still assert that the original configuration is written to the log output. 

Design-wise that feels like an oversight that should be addressed. If we still need some test-only mechanism to fiddle with the internal state then we should make that very explicit in the naming of that API that _that's_ what it's for.

d-ronnqvist:
I'll also echo Kyle's feedback from https://github.com/swiftlang/swift-docc/pull/713 which matches my overall opinion on DocC's preview command:

>I'm generally -0.5 for this PR.
>
>IMO, Preview is actually a little out of scope of Swift-DocC. So we only provide basic support for it(Just my opinion).
>If people want advanced feature for the preview(eg. check the preview on other network device), they can start a http server manually. It's very easy for almost every OS and in all language. We should not add complex here on DocC side.

CONTRIBUTING.md
d-ronnqvist: To illustrate how silly it is to wrap text like this; this is what the PR description looks like because you've wrapped the text 
Screenshot 2026-02-05 at 16 27 29


The issue is even worse on narrow viewports  where the wrapping actively harms readability
 Screenshot 2026-02-05 at 16 28 41

snprajwal:
Regarding the reason for this feature: I agree that we should not add unnecessary complexity into DocC's preview command, but IMO it is mostly trivial to be able to provide a custom host when we already have a flag to provide a custom port. It's a quality-of-life change for folks who tend to work on the repository via SSH or similar cases, where they can just port-forward from the remote machine and preview on the host machine's browser.

I mostly just moved the commits from #713 to this PR, with a minor change to prevent breaking the public API. Thanks for the detailed review, DocC has evolved significantly since the original PR. We can use this opportunity to modernise the test cases for this component, I'll rework the PR to bring stuff up to speed.

CONTRIBUTING.md
snprajwal: GitHub usually provides an unwrapped version of the commit message in the PR description, but for some reason did not insert it for this one, so I copied the commit message verbatim, which is wrapped for readability as the default git pager does not wrap lines. Makes sense to unwrap it in the web UI, I'll keep this in mind!

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
snprajwal: It makes more sense to rename both variants `localhost` and `socket` to be more explicit as `ip` and `unix` or something similar, so that it's obvious that one of them is a network bind and the other is a local file socket bind. What do you think?

d-ronnqvist:
> It's a quality-of-life change for folks who tend to work on the repository via SSH or similar cases, where they can just port-forward from the remote machine and preview on the host machine's browser.

Is that something that you do or is that a _hypothetical_ use-case? 

If someone needs preview functionality that's not offered by DocC's preview command it is very easy to start a local static HTTP on almost every OS in a wide range of languages. See for example [this list of one or two line commands for a variety of languages](https://gist.github.com/willurd/5720255).

d-ronnqvist:
> DocC has evolved significantly since the original PR. We can use this opportunity to modernise the test cases for this component, I'll rework the PR to bring stuff up to speed.

There's an opportunity cost to updating this code. In other words; the time spent both updating this code and reviewing those changes could be spent on other fixing other bugs or making other quality-of-life improvements that people have actually reported issues about.

snprajwal:
> Is that something that you do or is that a hypothetical use-case?

I do run DocC via SSH, but I see your point about alternative servers also being trivial to set up and use.

> There's an opportunity cost to updating this code. In other words; the time spent both updating this code and reviewing those changes could be spent on other fixing other bugs or making other quality-of-life improvements that people have actually reported issues about.

Fair enough, I agree. Closing this, thanks!

PR #1434, opened 2026-02-09
Include availability for fallback platforms when using @Available in symbols

franklinsch:
Bug/issue #, if applicable: rdar://169441896

## Summary

As a follow up to https://github.com/swiftlang/swift-docc/pull/1408, include availability for fallback platforms for symbols that use the `@Available` directive to override or complement the availability provided by the symbol graph.

The first commit extracts the logic we implemented for articles into a helper function, and the second commit uses it for symbols as well.

## Dependencies

None.

## Testing

Steps:

1. Build documentation for a symbol that defines iOS availability using `@Available`
2. Verify that the rendered documentation shows availability for iOS, iPadOS, and Mac Catalyst, as these should be inferred.

Alternatively, use the `AvailabilityBundle.docc` catalog to test.

## Checklist

- [x] Added tests


franklinsch:
@swift-ci please test

PR #1440, opened 2026-02-18
Treat API collections as articles in navigator

snprajwal:
Bug/issue #, if applicable: rdar://170471605

## Summary

The navigator uses a local enum for page types which is independent of the page types stored in the render JSON, and performs a mapping between the two when indexing a node for the navigator. The curation for the navigator restricts how nodes can be curated across languages:

- Symbols cannot be curated in languages that they are not available in, e.g. Swift symbols cannot be curated in an Objective-C-only page.
- Internal non-symbol pages can be curated across languages by making them available in the target language via the `@SupportedLanguage` directive.
- External non-symbol pages can be indiscriminately curated in a page of any language since a non-symbol page is not semantically attached to a specific language unless explicitly specified via the `@SupportedLanguage` directive.

According to these rules, API collections (which are just articles) can be curated in a page that is of a different language. However, while this works correctly in the topics section, the navigator does not do this due to API collections being treated as a symbol. This patch remaps the API collection to be treated as an article.

## Dependencies

N/A

## Testing

The existing tests have been updated to verify this behaviour. I also ran a documentation build with a test archive 100 times and verified that the output was consistent each time.

## Checklist

- [x] Added tests

### DIFF
--- a/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
+++ b/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
@@ -503,11 +503,6 @@ Root
                 return node.bundleIdentifier == testBundleIdentifier
             }))
             
-            let allNodes = navigatorIndex.navigatorTree.numericIdentifierToNode.values
-            let symbolPages = allNodes.filter { NavigatorIndex.PageType(rawValue: $0.item.pageType)! == .symbol }
-            // Pages with type `symbol` should be 6 (collectionGroup type of pages) as all the others should have a proper type.
-            XCTAssertEqual(symbolPages.count, 6)
-            
             assertUniqueIDs(node: navigatorIndex.navigatorTree.root)
             results.insert(navigatorIndex.navigatorTree.root.dumpTree())
             try FileManager.default.removeItem(at: targetURL)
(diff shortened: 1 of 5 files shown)


snprajwal:
@swift-ci please test

Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift
```diff
@@ -503,11 +503,6 @@ Root
                 return node.bundleIdentifier == testBundleIdentifier
             }))
             
-            let allNodes = navigatorIndex.navigatorTree.numericIdentifierToNode.values
-            let symbolPages = allNodes.filter { NavigatorIndex.PageType(rawValue: $0.item.pageType)! == .symbol }
```
d-ronnqvist, line 507: Just curious: are there other articles in this test or could we update this check to filter where `== .article` instead ad keep the assertion that there are 6 API collections?

snprajwal: These tests all use the legacy test bundle, and it does contain multiple other articles. I didn't want to change the test any more than I had to, since the right way to go about it would be to update the test to not use that bundle anymore. That's probably better suited for a separate PR.

snprajwal:
@swift-ci please test

PR #1424, opened 2026-01-28
Skip failing testSymbolLinksInDeclarationsAndRelationships test

franklinsch:
Bug/issue #, if applicable: rdar://169083752

## Summary

Skip testSymbolLinksInDeclarationsAndRelationships test that is causing CI failures. https://ci.swift.org/view/Apple%20Silicon%20(main)/job/oss-swift-pr-test-macos-arm64/1561/console

## Dependencies

N/A

## Testing

Existing tests should pass.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- ~[ ] Added tests~ N/A
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1447, opened 2026-02-20
Refine errors about missing HTML templates and missing files in the templates

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This refines the warnings and errors about missing either the full HTML template or missing one of the required files in the HTML template, based on the guidelines in https://github.com/swiftlang/swift-docc/pull/1442.

## Dependencies

None.

## Testing

- Build documentation where `DOCC_HTML_DIR` is set to a location that doesn't exist. Note the warning.
- Build documentation where `DOCC_HTML_DIR` is set to a non-directory. Note the warning.
- Build documentation where `DOCC_HTML_DIR` is set to a directory without an "index.html" file. Note the warning.
- _Preview_ documentation where `DOCC_HTML_DIR` is set to a location that doesn't exist, isn't a directory, or is missing the "index.html" file. Note the warning.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] ~Added~ Updated tests
- [ ] Ran the `./bin/test` script and it succeeded
- [ ] Updated documentation if necessary

PR #1442, opened 2026-02-20
Add new contributor documentation about adding diagnostics to DocC

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This adds a new article to DocC's contributor documentation with guidelines and recommendation about how to write user-facing actionable diagnostics (warnings and errors).

These commits were originally part of https://github.com/swiftlang/swift-docc/pull/1347 but I separated them into their own PR to make it easier to merge changes incrementally.

## Dependencies

None.

## Testing

Nothing. This is a documentation-only change.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- ~[ ] Added tests~
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1450, opened 2026-02-20
Finish loading external resolvers before registering local symbols

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This moves the `await` call for the external resolvers before the `registerSymbols(...)`call inside the context's private `register()` implementation because `registerSymbols(...)` internally calls `resolveExternalSymbols(...)` in its implementation.

Without these changes there is a risk that the external resolvers either:
- hadn't finished loading the external resolvers, which would result in links to external symbols not resolving. I think this was the cause of https://github.com/swiftlang/swift-docc/pull/1424
- was in the middle of assigning a value to `LinkResolver.externalResolvers` at the same time as `resolveExternalSymbols(...)` was accessing it, which would _likely_ result in a crash. The command with `--sanitize=thread` below find this case somewhat frequently (roughly 1 in 2 or 1 in 3 times).

## Dependencies

None

## Testing

- Run `for i in {1..10000}; do echo "\n$i"; swift test --sanitize=thread --parallel --filter SwiftDocCTests.DocumentationContextTests/testResolveExternalLinkFromTechnologyRoot; done`
- It shouldn't report any data races 

## Checklist

- [ ] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -2123,6 +2123,9 @@ public class DocumentationContext {
         
         registerRootPages(from: rootPageArticles)
         
+        // `registerSymbols(...)`, just below, calls `resolveExternalSymbols()` internally so we need to have finished loading the external symbol data before calling it.
+        _ = try await loadExternalResolvers
+        
         try registerSymbols(symbolGraphLoader: symbolGraphLoader, documentationExtensions: documentationExtensions)
         // We don't need to keep the loader in memory after we've registered all symbols.
         _ = consume symbolGraphLoader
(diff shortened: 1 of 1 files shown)


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
I'm separately running `for i in {1..10000}; do echo "\n$i"; swift test --parallel --quiet; done` locally to see if I can still reproduce the crash which seems to have impacted the macOS CI in https://github.com/swiftlang/swift-docc/pull/1442 and https://github.com/swiftlang/swift-docc/pull/1447. Without this change I could roughly reproduce that crash about 1 in 50 runs.

d-ronnqvist:
I've run the tests locally repeatedly for over an hour without any crashes. I'm sufficiently confident that this solve that issue.

PR #1454, opened 2026-02-20
Refine diagnostic about multiple files producing the same output path

d-ronnqvist:
Bug/issue #, if applicable:

## Summary

This refines the diagnostic about different articles colliding or different tutorials colliding based on the guidelines in https://github.com/swiftlang/swift-docc/pull/1442.

Specifically it displays the resulting output path that collides, lists the catalog-relative paths for the collisions, and explain why this issue happens.

## Dependencies

None.

## Testing

Build documentation where more than one article in different subdirectories have the same file name (case insensitive). Note the warning.

## Checklist

- [x] ~Added~ Updated tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -774,7 +774,7 @@ public class DocumentationContext {
             
             // Since documentation extensions' filenames have no impact on the URL of pages, there is no need to enforce unique filenames for them.
             // At this point we consider all articles with an H1 containing link a "documentation extension."
-            let isDocumentationExtension = (analyzed as? Article)?.title?.child(at: 0) is (any AnyLink)
+            let isDocumentationExtension = (analyzed as? Article)?.title?.startsWithAnyLink == true
             
             if let firstFoundAtURL = fileByOutputPath[path.lowercased()], !isDocumentationExtension {
                 let thisRelativePath: String
@@ -1254,7 +1254,7 @@ public class DocumentationContext {
                     }
                     
                     // Present a diagnostic specific to documentation extension files but get the solutions and notes from the general unresolved link problem.
-                    let unresolvedLinkProblem = unresolvedReferenceProblem(source: documentationExtension.source, range: link.range, severity: .warning, uncuratedArticleMatch: nil, errorInfo: errorInfo, fromSymbolLink: link is SymbolLink)
+                    let unresolvedLinkProblem = unresolvedReferenceProblem(source: documentationExtension.source, range: link.range, severity: .warning, errorInfo: errorInfo, fromSymbolLink: link is SymbolLink)
                     
                     diagnosticEngine.emit(
                         Problem(
@@ -1793,7 +1793,7 @@ public class DocumentationContext {
                 // rdar://79745455 and https://github.com/swiftlang/swift-docc/issues/593 tracks fixing the root cause of this issue, avoiding the collision and allowing the article and symbol to both exist.
                 diagnosticEngine.emit(
                     Problem(
-                        diagnostic: Diagnostic(source: article.source, severity: .warning, identifier: "org.swift.docc.articleCollisionProblem", summary: """
+                        diagnostic: Diagnostic(source: article.source, severity: .warning, identifier: "ArticleCollideWithSymbol", summary: """
                             Article '\(article.source.lastPathComponent)' (\(title)) would override \(existing.kind.name.lowercased()) '\(existing.name.description)'.
                             """, explanation: """
                             DocC computes unique URLs for symbols, even if they have the same name, but doesn't account for article filenames that collide with symbols because of a bug. 
@@ -2254,7 +2254,7 @@ public class DocumentationContext {
         }
 
         // Emit warnings for any remaining uncurated files.
-        emitWarningsForUncuratedTopics()
+        warnAboutArticlesOutsideTheDocumentationHierarchy()
         
         linkResolver.localResolver.addAnchorForSymbols(localCache: documentationCache)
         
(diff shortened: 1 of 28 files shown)


Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +768,57 @@ public class DocumentationContext {
...
-                        summary: """
-                        Redeclaration of '\(firstFoundAtURL.lastPathComponent)'; this file will be skipped
-                        """,
+                        identifier: "OutputPathCollision",
+                        summary: "Multiple \(fileDescription)s with output path '\(path.lowercased())'; this \(fileDescription) will be skipped",
                         explanation: """
-                        This content was already declared at '\(firstFoundAtURL)'
-                        """
+                        The relative path of \(isArticle ? "an article" : "a tutorial") in the rendered documentation is the name of its markup file, without the '.\(url.pathExtension)' extension, \
+                        replacing consecutive sequences of whitespace and punctuation with a hyphen, in this case '\(webURLPathComponent))'.
```
franklinsch:
```suggestion
                        replacing consecutive sequences of whitespace and punctuation with a hyphen, in this case '\(webURLPathComponent)'.
```

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +768,57 @@ public class DocumentationContext {
...
+                if let indexOfCatalog = url.standardizedFileURL.pathComponents.firstIndex(where: { $0.hasSuffix(".docc") }) {
+                    thisRelativePath  = url.standardizedFileURL.pathComponents.dropFirst(indexOfCatalog + 1).joined(separator: "/")
+                    otherRelativePath = firstFoundAtURL.standardizedFileURL.pathComponents.dropFirst(indexOfCatalog + 1).joined(separator: "/")
+                } else {
+                    let this  = url.standardizedFileURL.path
+                    let other = firstFoundAtURL.standardizedFileURL.path
+                    
+                    let commonPrefixLength = this.commonPrefix(with: other).count
+                    thisRelativePath  = String(this.dropFirst(commonPrefixLength))
+                    otherRelativePath = String(other.dropFirst(commonPrefixLength))
```
franklinsch: Hmm I think by using `dropFirst` on the String, it might produce odd behaviour when computing the relative paths. For example if the paths are `/Foo/AB/C` and `/Foo/AD/E`, the relative path would based off of the common `/Foo/A` rather than `/Foo`

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +768,57 @@ public class DocumentationContext {
...
+                }
+                
+                let isArticle = url.pathExtension == "md"
+                let fileDescription = isArticle ? "article" : "tutorial"
+                
+                let webURLPathComponent = urlReadablePath(firstFoundAtURL.deletingPathExtension().lastPathComponent).lowercased()
```
franklinsch: Should this use `url` instead, i.e., the file being skipped?

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +768,57 @@ public class DocumentationContext {
...
-                        This content was already declared at '\(firstFoundAtURL)'
-                        """
+                        The relative path of \(isArticle ? "an article" : "a tutorial") in the rendered documentation is the name of its markup file, without the '.\(url.pathExtension)' extension, \
+                        replacing consecutive sequences of whitespace and punctuation with a hyphen, in this case '\(webURLPathComponent))'.
+                        Because the pages for '\(thisRelativePath)' and '\(otherRelativePath)' would have the same web URL, DocC can only create a web page for one of them; deterministically keeping '\(otherRelativePath)' and dropping '\(thisRelativePath)'.
+                        """,
+                        notes: [
+                            DiagnosticNote(
+                                source: firstFoundAtURL,
+                                range: SourceLocation(line: 1, column: 1, source: nil) ..<  SourceLocation(line: 1, column: 1, source: nil),
```
franklinsch, line 815: I'm not quite sure how consumers of this would behave for an empty range. Have you tested this in IDEs for example? If it does work, it might make sense to create an `emptyRange` static property or something like that.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +768,57 @@ public class DocumentationContext {
...
+                                message: "Other \(fileDescription) with same filename here"
```
franklinsch: I'm guessing in this particular instance it doesn't matter, but technically, this could say "with same output path", or "with same web URL" or something like that.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: Good catch. In an earlier version there was an opening parenthesis before "in this case" but I replaced that with a comma when I added the extra details about consecutive sequences of whitespace and punctuation.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: I think I left this from an earlier iteration of this phrasing.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: An empty range simply represents a location without a length. We already have a handful of places that emit diagnostic with that range. 

We would still want to encourage people to specify exact ranges when possible so I think I'd want to name it something long and complicated rather than just `emptyRange`. 🤔

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: This used to just be the file URL but that got a bit confusing once I expanded the tests to cover filenames that aren't an exact match.

Because the explanation contextualizes this as having replaced whitespace and punctuation; it would look confusing if the value didn't match the description.

d-ronnqvist: Oh. Now I see what you're asking. It shouldn't be any difference which value we base this on singe they produce the same value.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1454/commits/788a0472133368f50a225ba20a1623ae821ced49. I had to use my own commit to also update the tests

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1454/commits/0bd1a0069013f00abb1e63f8f7f66a99746b98b2. 

I also updated the sorting logic to prefer shallower paths (which seems like the behavior that I would expect as a developer/writer)

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: Added an assert that the paths are the same in https://github.com/swiftlang/swift-docc/pull/1454/commits/1c00041904819b71856017f1a9c8134fcb9a065d

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: I added a in `makeEmptyStartOfFileRangeWhenSpecificInformationIsUnavailable(source:)` helper in https://github.com/swiftlang/swift-docc/pull/1454/commits/7d81578b464eab98169c2ed51559343598bfc91c.

Hopefully that's a long and scary enough name that people only use it as a last resort (preferring to determine a more specific range)

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1454/commits/ab7b1c52911cceccd6b87cbfb06ea926be2a1961

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -768,28 +776,58 @@ public class DocumentationContext {
...
+                let thisRelativePath: String
+                let otherRelativePath: String
+                if let indexOfCatalog = url.standardizedFileURL.pathComponents.firstIndex(where: { $0.hasSuffix(".docc") }) {
+                    thisRelativePath  = url.standardizedFileURL.pathComponents.dropFirst(indexOfCatalog + 1).joined(separator: "/")
+                    otherRelativePath = firstFoundAtURL.standardizedFileURL.pathComponents.dropFirst(indexOfCatalog + 1).joined(separator: "/")
+                } else {
+                    let this  = url.standardizedFileURL.pathComponents
+                    let other = firstFoundAtURL.standardizedFileURL.pathComponents
+                    
+                    let commonPrefixLength = zip(this, other).prefix(while: { $0 == $1 }).count
```
franklinsch, line 789: Nice!

d-ronnqvist:
@swift-ci please test

PR #1455, opened 2026-02-20
Remove some repetitive boilerplate in the command line actions

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This removes some repetitive boilerplate in DocC's command line actions:

- Single-use initializers that transform a "command" into an "action". Note that the `ConvertAction` initializer is called from both the `Convert` command and the `Preview` command, so that reduces code duplication and is _not_ removed.
- Properties that only forward its getter and setter to a property (often with the same name) in an option group.

The goal with this change is that the next time someone—either a frequent contributor or a new contributor—needs to add a command line flag, they won't repeat this type of boilerplate out of an implicit convention. 

> [!NOTE]  
> It doesn't matter that this removes "public" properties because the `DocCCommandLine` target isn't a _product_; so it's not accessible outside of the DocC repository. As such, there's no concern about breaking changes in _that_ target.

## Dependencies

None.

## Testing

## Checklist

- [ ] Added tests

### DIFF
--- a/Tests/SwiftDocCTests/Infrastructure/ExternalPathHierarchyResolverTests.swift
+++ b/Tests/SwiftDocCTests/Infrastructure/ExternalPathHierarchyResolverTests.swift
@@ -687,7 +687,6 @@ class ExternalPathHierarchyResolverTests: XCTestCase {
     }
     
     func testSymbolLinksInDeclarationsAndRelationships() async throws {
-        throw XCTSkip("Skipping due to CI failures - rdar://169083752")
         // Build documentation for the dependency first
         let symbols = [("First", .class), ("Second", .protocol), ("Third", .struct), ("Fourth", .enum)].map { (name: String, kind: SymbolGraph.Symbol.KindIdentifier) in
             return SymbolGraph.Symbol(


Sources/DocCCommandLine/ArgumentParsing/Subcommands/TransformForStaticHosting.swift
```diff
@@ -46,24 +46,21 @@ extension Docc.ProcessArchive {
...
-            } else {
-                throw TemplateOption.missingHTMLTemplateError(
-                    path: templateOption.defaultTemplateURL.path
-                )
-            }
-        }
-        
         func run() async throws {
-            let action = try TransformForStaticHostingAction(fromCommand: self)
+            // If we did this validation in `validate()` we'd still have to force unwrap the `templateURL` here.
```
anferbui: I found this sentence hard to follow, WDYT about changing the wording to focus on the approach we've gone with instead of the one we haven't:
```suggestion
            // We perform validation here instead of in  `validate()` to avoid having to force unwrap the `templateURL` here.
```

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
```diff
@@ -241,6 +209,7 @@ extension Docc {
             )
             var diagnosticsOutputPath: URL?
             
+            /// The diagnostic severity level to filter
```
anferbui:
```suggestion
            /// The diagnostic severity level to filter.
```

d-ronnqvist:
@swift-ci please test

hjyamauchi:
Hi @d-ronnqvist

It seems that we started getting this build error about a file that this PR removed:

```
CMake Error at Sources/DocCCommandLine/CMakeLists.txt:10 (add_library):
  Cannot find source file:
    ArgumentParsing/ActionExtensions/EmitGeneratedCurationAction+CommandInitialization.swift
CMake Error at Sources/DocCCommandLine/CMakeLists.txt:10 (add_library):
  No SOURCES given to target: DocCCommandLine
CMake Generate step failed.  Build files cannot be regenerated correctly.
```

The file is still referenced here:

https://github.com/swiftlang/swift-docc/blob/eb56ed06352eb7d2a1f88ea748c5386c79cbd809/Sources/DocCCommandLine/CMakeLists.txt#L32

Do we need to update the CMakeLists.txt?

hjyamauchi:
Created https://github.com/swiftlang/swift-docc/pull/1457

PR #1460, opened 2026-02-27
Prevent anchors from colliding with pages in links

d-ronnqvist:
Bug/issue #, if applicable: rdar://126662992

## Summary

This prevents anchors/headings/task groups/landmarks from colliding with pages in links by both return them separately from the other page components in the path parser and by storing them separately from the nodes that represent pages in the path hierarchy.

With these changes, developer have to explicitly use `#` to refer to an anchor in a link.
If the link is _only_ the anchor it's still permitted to omit the `#`, but if there's a page match with the same name, that is found first.

This also adds a new link warning specifically about anchors not existing on the node representing a specific page.

I have some further code cleanup in mind but I think I'll save that for its own PR.

## Dependencies

None.

## Testing

- Create a catalog with 2 articles: 
  - "First.md"
    ```
    # Some article

    The heading below has the same name as the second article.

    ## Second

    These links, with and without a '#' prefix, resolve to different things:
    - <doc:#Second>
    - <doc:Second>
    ```
  - "Second.md"
    ```
    # Second article

    This article has the same name as the heading in the first article.
    ```
- Build documentation
- Inspect the "First" page. 
  - The first link on that page should resolve to the heading on the same page
  - The second link on that page should resolve to the other page.

## Checklist

- [x] Added tests


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
Looks like there's some `#expect()` differences between 6.0 and 6.1 (and later):
```
macro expansion #expect:3:8: error: call can throw, but it is executed in a non-throwing autoclosure
```

I'll fix this on Monday. I might also add another test or two.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -481,10 +489,33 @@ struct PathHierarchy {
...
     
+    /// Adds an anchor to an existing element in the path hierarchy.
+    /// - Parameters:
+    ///   - parent: The unique identifier of the existing element to add the new child element to.
+    ///   - name: The name of the new anchor.
+    /// - Returns: The new unique identifier that represent this element.
+    @inlinable
+    mutating func addAnchor(parent: ResolvedIdentifier, name: String) -> ResolvedIdentifier {
+        let parent = lookup[parent]!
+        // Its rather easy to author the same heading more than once. Only add the anchor once.
```
franklinsch:
```suggestion
        // It's rather easy to author the same heading more than once. Only add the anchor once.
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -503,56 +534,78 @@ extension PathHierarchy {
...
         /// The descendants of this node in the hierarchy.
-        /// Each name maps to a disambiguation tree that handles
+        ///
+        /// Each name maps to a disambiguation tree that handles disambiguating matches and identifying collisions
+        @usableFromInline
         private(set) var children: [String: DisambiguationContainer]
         
+        /// The anchors of this node.
+        ///
+        /// An an anchor represents a heading or on-page landmark and is always a leaf node in the hierarchy.
```
franklinsch:
```suggestion
        /// An anchor represents a heading or on-page landmark and is always a leaf node in the hierarchy.
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+Error.swift
```diff
@@ -64,6 +64,14 @@ extension PathHierarchy {
         /// - A list of the names for the children of the partial result.
         case unknownName(partialResult: PartialResult, remaining: [PathComponent], availableChildren: Set<String>)
         
+        /// Encountered an unknown anchor at the end of the the path.
```
franklinsch:
```suggestion
        /// Encountered an unknown anchor at the end of the path.
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
```diff
@@ -78,18 +78,20 @@ extension PathHierarchy.PathParser {
     ///
     /// - Parameters:
     ///   - path: The documentation link string, containing a path and an optional fragment.
-    /// - Returns: A pair of the parsed path components and a flag that indicate if the documentation link is absolute or not.
-    static func parse(path: String) -> (components: [PathComponent], isAbsolute: Bool) {
-        guard !path.isEmpty else { return ([], true) }
+    /// - Returns: The the parsed components, a flag that indicate if the documentation link is absolute or not, and the parsed trailing anchor (if any).
```
franklinsch:
```suggestion
    /// - Returns: The parsed components, a flag that indicate if the documentation link is absolute or not, and the parsed trailing anchor (if any).
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -503,56 +534,78 @@ extension PathHierarchy {
...
         /// A set of non-standard behaviors that apply to this node.
+        @usableFromInline
         fileprivate(set) var specialBehaviors: SpecialBehaviors
         
         /// Options that specify non-standard behaviors of a node.
         struct SpecialBehaviors: OptionSet {
+            @usableFromInline
             let rawValue: Int
             
             /// This node is disfavored in the the case of a link collision.
```
franklinsch: nit: 

```suggestion
            /// This node is disfavored in the case of a link collision.
```

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -2360,13 +2455,17 @@ class PathHierarchyTests: XCTestCase {
         
         let articleID = try tree.find(path: "/CatalogName/Some-Article", onlyFindSymbols: false)
         
-        XCTAssertEqual(try tree.findNode(path: "SomeClass", onlyFindSymbols: false, parent: articleID).symbol?.identifier.precise, nil,
-                       "A general documentation link will find the heading because its closer than the symbol.")
+        XCTAssertEqual(try tree.findNode(path: "SomeClass", onlyFindSymbols: false, parent: articleID).symbol?.identifier.precise, "some-class-id",
+                       "A general documentation link will pages rather than headings.")
```
franklinsch:
```suggestion
                       "A general documentation link will find pages rather than headings.")
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -503,56 +534,78 @@ extension PathHierarchy {
...
+        @usableFromInline
         private(set) var children: [String: DisambiguationContainer]
         
+        /// The anchors of this node.
+        ///
+        /// An an anchor represents a heading or on-page landmark and is always a leaf node in the hierarchy.
+        @usableFromInline
+        fileprivate(set) var anchors: [String: Node] // ???: Should this be just the ID?
+        
+        @usableFromInline
```
franklinsch: Do we need all these `@usableFromInline`/`@inlinable`? It seems unnecessary because the call sites are within the same module, but maybe I'm missing something.

d-ronnqvist: My primary reason for adding those attributes is to signal my intent to the compiler and have the compiler tell me if I'm wrong. 

The compiler may or may not decide to inline these calls with or without these annotations. However, by adding these attributes the compiler will raise an error if the implementation does something that prevents inlining. That way, I can fix the implementation so that I know that the code _is_ inlinable—if the compiler should choose to do so—rather than _think_ that it's inlinable when there's really something that prevents it from being inlined, so it never does get inlined.

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
franklinsch: That makes sense for public APIs, but since `PathHierarchy` and `Node` are internal/private, I believe the compiler already has full visibility of their implementations and can inline them freely without the attribute.

d-ronnqvist: I think this tiny example better illustrates what I'm trying to prevent here. 

As far as I understand internal API can inline just fine when called from other internal API. However, that same API can't be used—even as an implementation detail—in an inlinable `package` or `public` API:
Screenshot 2026-03-06 at 17 51 27

If I mark all of those `@inlineable`/`@usableFromInline` I get new errors about what's preventing that API from being inlinable or usable from inline:
Screenshot 2026-03-06 at 17 58 47

It's only by fixing the errors all the way down that the `package` or `public` API can even use this internal API in its implementation.

Screenshot 2026-03-06 at 18 02 27

That said; I _think_ that the compiler may still be able to inline a call to that package function—even if it wasn't marked `@inlinable`—as long as that call was from another internal or package function. 

A little experimentation in Godbolt / Compiler Explorer seems to confirm that but I don't _know_ if that's always true.

----

Specifically for the PathHierarchy code; it's really only being called from two places:
-  the internal `PathHierarchyBasedLinkResolver` which wraps it to encapsulate the mapping to and from topic references. 
- the public `LinkCompletionTools` which doesn't mark any of its API as `@inlinable`

so I don't think this has an any impact _in practice_; at least not yet. 

If we ever extract the link resolution code into its own target/module we would need to make its API `package`-level access so that the other targets can call it and in that case I would likely want to add all of these annotations to ensure that the code remain inlinable everywhere. 

Likewise, if we ever wanted to make some of the API `LinkCompletionTools` inlinable we would need to add all these annotations to fix the compiler errors (as seen in the screenshots above).

----

If you feel that these annotations make the code too noisy I can remove them for now; and only add them back when they need to be called from something else that's inlinable or if we find a case where internal code isn't automatically inlinable when we think it should be.

d-ronnqvist: Actually. Let me just remove them now to avoid an extra round of reviews.

(done in https://github.com/swiftlang/swift-docc/pull/1460/commits/94f2d5d1849580f2674b35dc3df0a8134ecd645e)

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
```diff
@@ -213,6 +209,7 @@ extension PathHierarchy.PathParser {
 private struct PathComponentScanner: ~Copyable {
     private var remaining: Substring
     
+    @usableFromInline
```
franklinsch: There is still a `@usableFromInline` here, not sure if intentional.

franklinsch (approved):

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
d-ronnqvist: My regular expression that removed the rest must have kissed it.

d-ronnqvist:
@swift-ci please test

PR #1462, opened 2026-03-02
Deterministically sort loaded symbol graphs

snprajwal:
Bug/issue #, if applicable: rdar://170241366

## Summary

The symbol graph loader reads symbol graphs in parallel into a dictionary that maps from `URL` to `SymbolGraph`. The ordering of this dictionary is unstable, requiring a sort prior to processing the symbol graphs in order to ensure that the symbols are always registered in the same order of platforms and languages. However, the comparator to sort symbol graphs only performs a partitioned sort to place main symbol graphs ahead of extensions. It does not sort within these partitions to order the platforms and languages, leading to non-deterministic behaviour when building documentation for targets with symbol graphs for more than one platform and language. This patch updates the comparator to perform a cascading sort based on the following keys in order:

- Whether it is a main symbol graph or an extension
- The platform that the symbol graph is for
- The language of the symbol graph

This ensures that irrespective of the initial ordering of symbol graphs in the dictionary, the order of registering symbols in these symbol graphs is guaranteed to be the same in every documentation build.

## Dependencies

N/A

## Testing

The unit tests for platform and language sorting have been updated. A new unit test has been added to verify the ordering of symbol graphs.

## Checklist

- [x] Added tests


snprajwal:
@swift-ci please test

d-ronnqvist (commented):
I don't know if I feel that this is the right solution to this problem. 

Rather than enforcing an order I would prefer if the result were the same regardless of order (separating main and extension files into two phases).

Sources/DocCCommon/SourceLanguage.swift
```diff
@@ -358,20 +358,18 @@ public extension SourceLanguage {
...
-            false
-        } else {
-            // Otherwise, sort by ID (a string) for a stable order.
-            lhs.id < rhs.id
+        // If either of the languages is known, it is sorted before the other.
+        if lhsIsKnown != rhsIsKnown {
+            return lhs._isKnownLanguage
         }
+        // Otherwise, they are sorted by the string identifier for a stable order.
+        return lhs.id < rhs.id
```
d-ronnqvist, line 372: Why this change in behavior?

d-ronnqvist (changes requested):

snprajwal:
The main issue that this patch addresses has also been identified to possibly be an issue with the symbol graphs themselves, moving this to a draft for now.

PR #1463, opened 2026-03-03
Make DisplayName directives override both the page title and the navigator title

d-ronnqvist:
Bug/issue #, if applicable: rdar://170218493

## Summary

This updates the `RenderMetadata` transformation to also use custom `DisplayName` values for the `navigatorTitle`. This behavior matches what's already described in the [DisplayName directive documentation](https://www.swift.org/documentation/docc/displayname#Overview) (my added emphasis):

> **A custom display name appears** in place of the symbol’s name **in the following locations**:
> - The main header on the symbol page
> - **The navigator hierarchy**
> - Breadcrumbs for the symbol page and its child symbol pages
> - Link text anywhere that includes a link to the symbol page

## Dependencies

None.

## Testing

> [!NOTE]
>  because of https://github.com/swiftlang/swift-docc/issues/176 you might get false positives (it working as described without these changes) if you try to verify this behavior with a Swift symbol.

- In some non-Swift project, specify a custom display name using the `DisplayName` directive for one of the symbols and build documentation.
  - On that page, the symbol should display the custom display name instead of its symbol name (this worked before)
  - In the navigator hierarchy, the symbol should _also_ display the custom display name instead of its symbol name (this is new)

## Checklist

- [x] Added tests
- ~[ ] Updated documentation if necessary~ The [relevant documentation](https://www.swift.org/documentation/docc/displayname#Overview) already describes this behavior.

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -321,7 +321,7 @@ public class NavigatorIndex {
             case "enumsub", "structsub", "instsub", "intfsub", "subscript": self = .subscript
             case "enumcm", "structcm", "clm", "intfcm", "type.method": self = .typeMethod
             case "httpget", "httpput", "httppost", "httppatch", "httpdelete": self = .httpRequest
-            case "dict": self = .dictionarySymbol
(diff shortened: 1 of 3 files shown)


d-ronnqvist:
@swift-ci please test

PR #1464, opened 2026-03-03
Avoid reading out-of-bounds even when safe to avoid false-positive AddressSanitizer issues

d-ronnqvist:
Bug/issue #, if applicable: rdar://171251109

## Summary

This updates the implementation of the faster custom symbol graph JSON decoder to avoid doing safe out-of-bounds reads because it prevented us from running AddressSanitizer issues.

The very short version is that instead of only doing 8-bytes-at-a-time check when scanning for closing string delimiters. The implementation now falls back to a 1-byte-at-a-time check for the last 7 bytes of the JSON data.
If you're curious to known more; the added code comments explain the changes in more detail than what's worth repeating in this PR description. 

## Dependencies

None.

## Testing

Run `swift test --sanitize=address`. There shouldn't be any issues reported.

## Checklist

- ~[ ] Added tests~

### DIFF
--- a/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
+++ b/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
@@ -689,7 +689,7 @@ package struct FastSymbolGraphJSONDecoder: ~Copyable {
         try _skipExpectedByte(.init(ascii: "\""))
         let startOfString = pointer
 
-        // Unless the bytes contains a slash (`0x5C`), they can trivially be decoded as a string.
+        // Unless the bytes contains a backslash (`0x5C`), they can trivially be decoded as a string.
         var isTrivialStringConvertible = true
         
         var length = 0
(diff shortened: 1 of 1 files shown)


d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -695,7 +695,76 @@ package struct FastSymbolGraphJSONDecoder: ~Copyable {
...
+            } else {
+                // An odd number of slashes means that the last slash escapes the quote, so this byte is part of the string's content.
+                // For example, consider a string that ends in 3 slashes. The 1st slash escapes the 2nd leaving 3rd slash to escape the quotation mark:
+                //
+                //     "ABC\\\"
+                //         ├╯├╯
+                //         │ ╰───╴escaped quote
+                //         ╰─────╴escaped slash
+                return true
+            }
```
d-ronnqvist, line 755: FYI: this code hasn't changed. I just extracted it to be able to call it from both the 8-bytes-at-a-time loop and the 1-byte-at-a-time loop.

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -695,7 +695,76 @@ package struct FastSymbolGraphJSONDecoder: ~Copyable {
...
+        }
+        
+        // This inner function exist so that both the 8-bytes-at-a-time loop and the 1-byte-at-a-time loop do the same bounds check before returning the buffer.
+        func boundsCheckAndMakeReturnValue() throws(ScanningError) -> (buffer: UnsafeRawBufferPointer, isTrivialStringConvertible: Bool) {
+            pointer.removeFirst(length + 1)
+            try _boundsCheck()
+            
+            return (.init(start: startOfString, count: length), isTrivialStringConvertible)
```
d-ronnqvist, line 763: FYI: this code also hasn't changed. I just extracted it to be able to call it from both the 8-bytes-at-a-time loop and the 1-byte-at-a-time loop. See the comment at the 8-bytes-at-a-time call site for why both loops need to call this.

Sources/DocCCommon/JSONDecoding/UnsafeRawPointer+HasASCIIPrefix.swift
```diff
@@ -67,60 +67,62 @@ extension UnsafeRawPointer {
...
-                // This technically risks reading beyond the bounds of `self`, but because we mask away the extra out-of-bounds byte, it doesn't matter.
-                let bytesToCompareTo = UInt64(buffer[0])
-                                     | UInt64(buffer[1]) &<<  8
-                                     | UInt64(buffer[2]) &<< 16
-                                     | UInt64(buffer[3]) &<< 24
-                                     | UInt64(buffer[4]) &<< 32
-                return loadUnaligned(fromByteOffset: byteOffset, as: UInt64.self) & 0x00_00_00_FF_FF_FF_FF_FF == bytesToCompareTo
+                // It would be safe and avoid a 2nd read if we loaded 8 bytes and masked away the last 3 bytes before comparing to the 5 bytes of static string data.
+                // However, to an address sanitizer this understandably looks like an unsafe mistake and we want DocC to build without address sanitizer issues to help find real memory issues.
+                let bytesToCompareTo: UInt32 = UInt32(buffer[0])
```
d-ronnqvist, line 84: FYI: The extra type annotations in this file is just to avoid slow type checking. I've been noticing in the CI build logs that these bit shifting expressions exceed the 1 second type checking warning and even locally these bit shifting expressions take ~350ms each to type check. That entirely goes away by explicitly specifying the type on the value (and not just on the individual shifted bits that are being OR'd together.)

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -695,7 +695,76 @@ package struct FastSymbolGraphJSONDecoder: ~Copyable {
...
+        
+        // This inner function exist so that both the 8-bytes-at-a-time loop and the 1-byte-at-a-time loop can verify that they've found the closing string delimiter.
+        func checkIfQuoteAtCurrentLocationIsEscaped() -> Bool {
+            assert(
+                pointer.load(fromByteOffset: length, as: UInt8.self) == .init(ascii: "\""),
+                "The current character isn't a quote; is '\(Character(UnicodeScalar(pointer.load(fromByteOffset: length, as: UInt8.self))))' instead."
+            )
+            
+            if isTrivialStringConvertible {
+                // If we haven't encountered any slashes since the start of this string, the quote cannot be escaped.
```
snprajwal:
```suggestion
                // If we haven't encountered any backslashes since the start of this string, the quote cannot be escaped.
```

d-ronnqvist:
@swift-ci please test

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
```diff
@@ -695,7 +695,76 @@ package struct FastSymbolGraphJSONDecoder: ~Copyable {
...
+                return false
+            }
+            
+            // To determine if the quote at current location is escaped or not, we counting the number of slashes before it (because slashes also escape other slashes).
```
snprajwal: Ditto for this section, we check for backslashes, slash typically refers to `/`.

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: There were a lot more places to update so I made those changes in https://github.com/swiftlang/swift-docc/pull/1464/commits/5363f70123637b4116e2d0fca71963826ee2fb5f

Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
d-ronnqvist: There were a lot more places to update so I made those changes in https://github.com/swiftlang/swift-docc/pull/1464/commits/5363f70123637b4116e2d0fca71963826ee2fb5f

d-ronnqvist:
@swift-ci please test

PR #1456, opened 2026-02-23
Remove workaround for unnamed C symbols (#1100)

snprajwal:
Bug/issue #, if applicable: rdar://171002757

## Summary

This reverts commit 093049d1c34869a703eee05785af91eebbaacb65.

The issue has been fixed upstream in https://github.com/llvm/llvm-project/pull/120801, which means that we can safely remove this workaround.

## Dependencies

N/A

## Testing

Tested with the original framework that was failing to build in rdar://139305015.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1467, opened 2026-03-05
Revert "Remove workaround for unnamed C symbols (#1100) (#1456)"

snprajwal:
This reverts commit 275fae1be027c1c6aa7bd503951e717d3f26fb34.

It was identified that there were still some edge cases where Clang incorrectly generates symbol graph entries with empty titles and path components. Removing this workaround causes the indexing records to fail to generate, breaking builds.

rdar://171796871


snprajwal:
@swift-ci please test

PR #1470, opened 2026-03-11
Consider "unavailable" fallback platforms for availability provided using directives

d-ronnqvist:
Bug/issue #, if applicable: rdar://171710882

## Summary

The recent number of closed and open fixes in the availability logic has highlighted inconsistencies in how availability is computed depending on both type of page and depending on where the availability information is coming from (Info.plist, in-source attributes, or Available directives).

This PR aims to fix some of those inconsistencies while leaving explicit FIXME comment about others. I drew the line of what to fix and what to defer based on what I thought could be fixed _before_ extracting and refactoring the availability logic. 

## Dependencies

None.

## Testing

For some iOS project
- Configure either "iPadOS" or "Mac Catalyst" to be "unavailable" into default availability in the Info.plist:
  ```
  <key>CDAppleDefaultAvailability</key>
  <dict>
    <key>YourModuleName</key>
    <array>
      <dict>
        <key>name</key>
        <string>iPadOS</string>
        <key>unavailable</key>
        <true/>
      </dict>
    </array>
  </dict>
  ```
- Specify "iOS" availability for one symbol using each of the following methods
  - "Default" availability in the Info.plist
  - an in-source `@available` attribute
  - a `@Available` metadata directive in the symbol's documentation comment
  
- Build documentation for the symbol and inspect its rendered availability
  - Regardless of how the availability information was provided, the page should display the "iOS" availability and _either_ "iPadOS" or "Mac Catalyst" depending on which of the two _wasn't_ marked "unavailable" in the Info.plist
  
> Note: if the "default" iOS availability from the Info.plist doesn't specify an introduced version; DocC won't add the "fallback" availability. This is an inconsistency but there is one test that verifies this behavior—although I don't know for sure if it's intentional behavior—so I added a code comment rather than changing this behavior so that we can revisit that inconsistency in a follow up PR.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -83,10 +83,10 @@ public struct RenderNodeTranslator: SemanticVisitor {
         result.reserveCapacity(availabilities.count)
         
         for availability in availabilities {
-            let current = currentPlatforms.flatMap { platforms in
-                PlatformName(metadataPlatform: availability.platform).flatMap { name in
-                    platforms[name.displayName]
-                }
+            let current: PlatformVersion? = if let currentPlatforms, let name = PlatformName(metadataPlatform: availability.platform) {
+                currentPlatforms[name.displayName]
+            } else {
+                nil
             }
             let renderItem = AvailabilityRenderItem(availability, current: current)
             if let name = renderItem.name {


d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/ExternalPathHierarchyResolver.swift
```diff
@@ -202,7 +202,7 @@ extension LinkDestinationSummary {
             estimatedTime: nil,
             conformance: nil,
             isBeta: isBeta,
-            isDeprecated: platforms?.contains(where: { $0.unconditionallyDeprecated == true }) ?? false,
+            isDeprecated: platforms?.contains(where: { $0.unconditionallyDeprecated == true || $0.deprecated != nil }) ?? false,
```
patshaughnessy, line 205: Can `$0.deprecated` be a blank/empty string? Should we check for that here and consider that equivalent to nil? Or possibly add a computed property to `AvailabilityRenderItem` ?

```
    /// The version of the platform SDK deprecating the symbol.
    public var deprecated: String?
```

Sources/SwiftDocC/Model/Rendering/Symbol/AvailabilityRenderMetadataItem.swift
```diff
@@ -152,14 +152,25 @@ public struct AvailabilityRenderItem: Codable, Hashable, Equatable {
         return introduced >= SemanticVersion(versionTriplet: current.version)
     }
     
-    /// Creates a new item with the given platform name and version string.
-    /// - Parameters:
-    ///   - name: A platform name.
-    ///   - introduced: A version string.
-    ///   - isBeta: If `true`, the symbol is introduced in a beta version of the platform.
-    init(name: String, introduced: String?, isBeta: Bool) {
+    init(
```
patshaughnessy, line 155: What's the reason for adding these default values explicitly? Doesn't Swift automatically initialize optionals to nil?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -74,43 +74,27 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
-            // or if the fallback platform is not available.
-            guard !platforms.contains(platform),
-                  let fallbackIndex = platforms.firstIndex(of: fallback) else {
-                return nil
+    ) -> [String: AvailabilityRenderItem] {
+        var result = [String: AvailabilityRenderItem]()
+        result.reserveCapacity(availabilities.count)
+        
+        for availability in availabilities {
+            let current = currentPlatforms.flatMap { platforms in
```
patshaughnessy: Is there a simpler way to write this passage, using `?` instead of calling `flatMap` twice?

It's a little hard to follow this logic as written.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -878,13 +862,37 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
 
         if let availabilities = article.metadata?.availability, !availabilities.isEmpty {
-            let allRenderAvailabilities = renderAvailabilities(
-                from: availabilities,
-                currentPlatforms: context.configuration.externalMetadata.currentPlatforms
-            )
-
-            if !allRenderAvailabilities.isEmpty {
-                node.metadata.platformsVariants = .init(defaultValue: allRenderAvailabilities)
+            // FIXME: Move this logic out of the rendering code (rdar://172280267)
```
patshaughnessy, line 865: Yes please - this one function visitArticle is now 297 lines long!

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -878,13 +862,37 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+            let currentPlatforms = context.configuration.externalMetadata.currentPlatforms
+            // These are the same for all platforms, so we only need to compute them once.
+            var directiveAvailabilityByPlatform = documentationNode.metadata.map {
+                renderAvailabilities(from: $0.availability, currentPlatforms: currentPlatforms)
+            } ?? .init()
+            
+            if let iOSAvailability = directiveAvailabilityByPlatform[PlatformName.iOS.displayName] {
```
patshaughnessy, line 872: This new logic doesn't use the `DefaultAvailability.fallbackPlatforms` - are you removing that? Why is this version better?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1278,45 +1286,101 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
-                .compactMap { availability -> AvailabilityRenderItem? in
-                    // Allow availability items without introduced and/or deprecated version,
-                    // but filter out items that are obsoleted.
-                    if availability.obsoletedVersion != nil {
-                        return nil
-                    }
-                    // Filter out this availability item if it has a missing or invalid domain.
-                    guard let name = availability.domain.map({ PlatformName(operatingSystemName: $0.rawValue) }) else {
-                        return nil
+            // FIXME: Combine the in-source attributes with the Available directives as a per-platform override (rdar://171807245)
```
patshaughnessy, line 1321: Yes please! This code looks similar to or the same as what I just read above.

Sources/SwiftDocC/Infrastructure/Link Resolution/ExternalPathHierarchyResolver.swift
d-ronnqvist: I'm planning on changing this (specifically for external entities) in a follow up PR. The reason is that render nodes include the presence of a `DeprecationSummary` when deciding if the page is deprecated but that information isn't present in the `platforms` so an external entity can't reflect that logic. This results in some pages being considered deprecated within the module itself but not when referenced externally.

For now I'm satisfied with knowing that the `AvailabilityRenderItem` initializers don't create empty `deprecated` strings. Long term I feel like this property should be computed based on model values rather than render values.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
d-ronnqvist: What this is trying to express is that both `currentPlatforms` and `PlatformName(...)` have to be non-`nil` to look up the version information (which could also be missing)

It would be _simpler_ to check `PlatformName(...)` first and use `?` in the subscript
```swift
let current = PlatformName(metadataPlatform: availability.platform).flatMap { name in
    currentPlatforms?[name.displayName]
}
```

However, the cost of that simplification is that it initializes a `PlatformName` (which does a dictionary lookup and/or string processing) for every platform for every symbol even when `currentPlatforms` is `nil`. The compiler _may_ realize this and reorder the code, but it also may not.

Because most documentation builds don't pass any current platform information on the command line, this simplification would end up doing unnecessary work for every symbol in the common case. "Rendering" is already the slowest part of DocC, so I switched the order around to avoid that unnecessary work, resulting in one `flatMap` instead of a `?` to compute the same value.

----

The other alternative implementation that I could think of is to use regular `if-let` expressions for both values. That way, the code can check the `currentPlatforms` first and skip the `PlatformName` creating in the common case.

This gets 1 line longer, but is perhaps more readable, using an if-expression:

```swift
let current: PlatformVersion? = if let currentPlatforms, let name = PlatformName(metadataPlatform: availability.platform) {
    currentPlatforms[name.displayName]
} else {
    nil
}
``` 

I made this change in https://github.com/swiftlang/swift-docc/pull/1470/commits/e2db4fc2ac1255ff1dc7867fbe329e12783e6f78

Alternatively, it would be be 2 lines longer without an if-expression where both branches of the if-else-statement explicitly spell out `current = ` to assign the value.

```swift
let current: PlatformVersion?
if let currentPlatforms, let name = PlatformName(metadataPlatform: availability.platform) {
    current = currentPlatforms[name.displayName]
} else {
    current = nil
}
```

Sources/SwiftDocC/Model/Rendering/Symbol/AvailabilityRenderMetadataItem.swift
d-ronnqvist: Because the previous initializer didn't have any parameters for these values, it wasn't possible to optionally pass them. They could _only_ be initialized to `nil` using this initializer.

Because the type had an explicit initializer, Swift didn't synthesize a memberwise initializer so there was no way to pass these values during initialization.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
d-ronnqvist: Yes, in the future I am planning on removing `DefaultAvailability.fallbackPlatforms`. The problem with doing it now is that there are unit test that verify an intermediate value instead of the rendered value meaning that if I removed the fallback platforms property now, those tests would fail even if the rendered page would look the same. I think the time to remove `DefaultAvailability.fallbackPlatforms` is when we move all the availability logic into one place. Then it won't make sense to use it anymore.

As to why I think this version is better; the way that `DefaultAvailability.fallbackPlatforms` is structured is with keys for the fallbacks and values for the platform that each fallback derives its availability from. However, in practice both fallbacks derive their availability from the same platform. This means that instead of doing one lookup per symbol to see if there's iOS availability, there's two lookups per symbol of the same value. Further, because `iPadOS` is never present in symbol graph files, the generalized structure of the existing code means that it's always doing a redundant "iPadOS" lookup in the symbol information.

Note that during rendering that lookup is no longer redundant because it's possible that the iPadOS availability came from the Info.plist. 

By leveraging this "domain knowledge" of platforms and what their fallbacks we can switch this relationship around. This not only saves some work but IMO it makes it more clear that when there's iOS availability we infer other platforms from that. 

In other words; I find `DefaultAvailability.fallbackPlatforms` to be a premature generalization that's preparing for a future case that's unlikely to ever happen at the cost of current speed and complexity.

d-ronnqvist:
> I understand why you would like to make these changes on step at a time, but this logic is so complicated, and it's repeated in multiple different places, maybe it would make sense to jump right to the finished solution. What we really need is a single place where the availability for every symbol is calculated in the same way. And a series of unit tests that describe what the algorithm is.

We agree on what's needed here; a single place for the logic and a series of tests that make it clear what logic is.

My reason for not doing those changes in this PR is that I wanted to make this bug fix smaller and easier to review. I also have other changes in mind in some of this modified code, and I wanted this PR to be merged sooner so that I could start working on those. 

I think that the PR that moves all the availability logic into one place might be fairly big and it would help to ensure that it's solely a refactoring with no change in behavior. (so that changes in behavior can more easily be made in that one place later).

PR #1472, opened 2026-03-11
Fix non-Swift symbols not marked as deprecated in render references

timdream:
## Summary

`isDeprecatedVariants` was initialized with `defaultVariantValue: false`, which caused `firstValue` to return `false` for non-Swift languages (e.g., `interfaceLanguage: "data"`). The `defaultVariantValue` made `self[.swift]` return `Optional.some(false)` instead of `nil`, preventing the `??` fallback in `firstValue` from reaching the actual language variant's `true` value.

When the availability item also had a `message`, the bug was masked because `deprecatedSummary` became non-nil via a secondary `|| symbol.deprecatedSummary != nil` check. Without a message, both conditions were false and the symbol was silently rendered as non-deprecated.

The fix:

- ~~Remove `defaultVariantValue: false` from `isDeprecatedVariants` so `self[.swift]` returns `nil` for non-Swift symbols, allowing `firstValue` to fall through to the real variant value.~~
- ~~Change `isDeprecatedVariants.firstValue!` to `firstValue ?? false` for safe default when no variant is set.~~
- Update `DocumentationDataVariants.firstValue` to use `hasVariant` to check if the Swift variant exists, and returns the first value if not, instead of calling to `subscript()` which returns the default value.
- Added two tests in `DeprecationSummaryTests` using `interfaceLanguage: "data"` to verify deprecated symbols are correctly marked in topic render references both with and without a deprecation message.

## Dependencies

None. There is going to be a separate swift-docc-render PR to allow the renderer to consume the `deprecated` flag without deprecation message.

## Testing

Steps:
1. Run `xcrun docc preview` with `--additional-symbol-graph-dir` using the symbol graph attached: `xcrun docc convert Temp.docc --additional-symbol-graph-dir symbols --output-path Temp.doccarchive`
2. Inspect `Temp.doccarchive/data/documentation/testmodule/testclass/brokenprop.json` and verify that the brokenProp reference contains `"deprecated": true`
3. Inspect `Temp.doccarchive/data/documentation/testmodule/testclass.json` and verify both child references contain `"deprecated": true`

## Checklist

- [x] Added tests
- [ ] Updated documentation if necessary (N/A — minor bug fixes)


timdream:
[TestModule.symbols.json](https://github.com/user-attachments/files/25920706/TestModule.symbols.json)

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -706,7 +706,7 @@ extension Symbol {
     
     /// Whether the first variant of the symbol is deprecated.
     public var isDeprecated: Bool {
-        get { isDeprecatedVariants.firstValue! }
+        get { isDeprecatedVariants.firstValue ?? false }
```
patshaughnessy: Thanks for the fix here!

This looks good but I actually think you've been a bit too timid :) The problem you've found seems like a bug in the underlying `firstValue` function. Could you make the fix there instead? That way we avoid the same problem in other call sites that use `firstValue` with a default value.

Here's the `firstValue` implementation now:

```
    /// Convenience API to access the first variant, or the default value if there are no registered variants.
    ///
    /// > Important:
    /// > Do not use this property in new code.
    /// > It exists to transition existing code from only working with Swift symbols to working with multi-language symbols.
    /// > This property should be considered deprecated but isn't formally deprecated to avoid the ~50 warnings that would make it
    /// > harder to spot new warnings. (rdar://86580516)
    var firstValue: Variant? {
        // A Dictionary's order isn't stable across program executions so accessing the `first` value would
        // result in non-deterministic behavior and also flaky tests.
        //
        // Since this convenience accessor exist to transition existing code from only working with Swift symbols,
        // it accesses the Swift value first, if it exist, and otherwise accesses the real non-deterministic first value.
        // This assumes that variant only represents one non-Swift language.
        get { self[.swift] ?? self.values.first?.value }
        set { self[.swift] = newValue }
    }
```

The idea is that we return:
- The Swift variant if there is one, or
- The "first" variant (regardless of the dictionary ordering), or
- nil if there are no values.

But as you say in this PR's description, the `[.swift]` subscript call will return a value when a default value is specified, even if there is no Swift variant: `Optional.some(false)` instead of `nil`.

Maybe instead we should first check whether there is a Swift key using `hasVariant` just a few lines up in this file, and return the value if so. If not we continue with the other logic. I think the intent of `firstValue` was to return the first variant, even if it's not for Swift. (We also may want to return the default value, if there is not, instead of nil. This was happening now because of the way the `[]` operator works.)

Let me know what you think, or if you have any questions...

patshaughnessy (changes requested):

patshaughnessy (approved):
Looks good - thanks for the additional fix!

patshaughnessy:
@swift-ci please test

PR #1479, opened 2026-03-18
Small performance improvements for various link resolutions tasks 

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This implements few different smaller performance improvements for the `PathHierarchy` resulting in:
- 50% faster initialization
- 80–90% faster computation of `caseInsensitiveDisambiguatedPaths(...)` (used to determine the path of every symbol's `ResolvedTopicReference`)

Both these results were measured using symbol graph information for the Swift standard library.

These differences are hard to measure in a full `docc convert` call but show clear improvements in isolated benchmarks. 

## Dependencies

None.

## Testing

## Checklist

- [x] Added tests


d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
```diff
@@ -318,51 +302,57 @@ extension PathHierarchy.DisambiguationContainer {
...
-        for (id, disambiguations) in duplicateSymbols {
-            let primaryDisambiguation = disambiguated.first(where: { $0.value.symbol?.identifier.precise == id })!.disambiguation
             
-            for element in disambiguations {
+            // Disambiguate the elements with the unique symbols collapsed (duplicate symbols with the same USR are not in `new`).
+            var disambiguated = new.disambiguatedValues(includeLanguage: includeLanguage, allowAdvancedDisambiguation: allowAdvancedDisambiguation)
+            
+            // Add values for the duplicate symbols (without updating the _amount_ of necessary disambiguation)
+            for index in 0 ..< duplicatesCount {
+                let element = duplicatesBuffer[index]
```
franklinsch, line 359: I don't know if this increments the reference count. If it does, I think it may cause a leak because this is in `withUnsafeTemporaryAllocation`.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
```diff
@@ -486,3 +476,76 @@ private extension Collection {
...
+            0x61...0x7A,
+            // ~
+            0x7E: true
+        default:  false
+        }
+    }
+    
+    /// A Boolean value indicating if the UTF8 code unit represents a single-byte character or if it is the start of a variable-length unicode scalar.
+    var isSingleByteCharacter: Bool {
+        self < 0b1000_000
```
franklinsch: I think maybe there's a missing zero here, but I'm not sure. In any case, I think this would hinder performance ever so slightly but not correctness. Or I'm misunderstanding what this property is representing.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
franklinsch: Thinking more, yes I think this might be a leak, but I haven't tested with ASan

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
d-ronnqvist: Right. That's a typo. Thanks for catching that.
I had it spelled out as `128` but changed it to a binary literal after running the tests.

franklinsch: Actually yes, why not write `128`?

d-ronnqvist: The UTF-8 spec describes that the first bit indicates if the variable-length encoding uses one or multiple bytes to represent a unicode scalar. I felt that it was more explicit to represent that using a binary literal. 

The pattern gets more obvious when you consider the byte patterns that indicate if a unicode scalars is represented using 2, 3, or 4 bytes (for example `0b1110_0000`) which I was looking at for another possible future PR.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
d-ronnqvist: I don't think that this leaks anything for two reasons:
- `Element` is a structure, rather than a class, so it isn't reference counted but instead copied to the new location. 

- `withUnsafeTemporaryAllocation` provides access to a temporary buffer that's only available during the scope of that closure. Elements are both inserted (initialized) and read inside the scope of that function. The buffer doesn't outlive the closure. 


The only reason I could see for this possibly to leak anything would be if the individual string properties of each Element is retained an extra time by the copy _and_ the buffers memory isn't cleaned up when it is deallocated (meaning that unallocated garbage memory continue to have the same old bits of those element copies).

Even in this potential case (which depends on the assumption that the memory isn't cleaned up) I think that the original elements would live to the end of the program execution, meaning that the string properties wouldn't be fully released regardless. 

Still, it's good to deinitialize the buffers memory like its documentation says. See ([this comment][1] below. That should ensure that there's no stale bits unintentionally retaining any of the string properties. 



[1]: https://github.com/swiftlang/swift-docc/pull/1479#discussion_r2967384923

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
d-ronnqvist: I fixed the bit literal in https://github.com/swiftlang/swift-docc/pull/1479/commits/88dd46e5cca25a9ae464b8a4a9dc78db91fa6065 and added a code comment about what it represents.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
d-ronnqvist: I noticed while reading the documentation that the closure us responsible for both initializing and deinitializing (but not deallocating) the buffer so I added the missing `deinitialize()` call with a code comment in https://github.com/swiftlang/swift-docc/pull/1479/commits/88dd46e5cca25a9ae464b8a4a9dc78db91fa6065

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+DisambiguatedPaths.swift
d-ronnqvist: I ran `swift test --sanitize=address --parallel --filter SwiftDocCTests` (and the same with `--sanitize=thread`) to verify that neither sanitized reports any issues.

d-ronnqvist:
Franklin and I talked about this code offline and after describing some of the implementation to him I felt that it could be beneficial to have the same descriptions in the code so I added more code comments in https://github.com/swiftlang/swift-docc/pull/1479/commits/c3b433bd39aab3c62aa49c055b68ccc52003a350 that reiterate what we talked about.

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@franklinsch do you have any other questions about these changes?

franklinsch (approved):
Thanks David!

d-ronnqvist:
@swift-ci please test

PR #1480, opened 2026-03-19
Don't include HTML in markdown output

jrturton:
Bug/issue #, if applicable: 171381879

## Summary

HTML inside DocC documentation should not be included in the markdown output, unless it is included in a code block. HTML is not included in the render JSON output and is not visible in documentation previews. Markdown output should reflect that.

## Dependencies

None

## Testing

New unit tests have been added to exercise the code. 

## Checklist

- [x] Added tests
- [ ] Updated documentation if necessary - N/A for bugfix


Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -352,6 +352,22 @@ extension MarkdownOutputMarkupWalker {
         }
         
     }
+    
+    mutating func visitHTMLBlock(_ html: HTMLBlock) -> () {
```
franklinsch: I had a look at what we do for render JSON, and it seems like we're stripped HTML blocks and inline HTML entirely. It would seem beneficial to try and keep those two rendering paths as close to each other as possible. What do you think?

jrturton: Makes perfect sense, as long as that's a feature and not a bug.

jrturton: Re-worked the extract and retitled / redescribed the PR to reflect this.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -580,6 +580,64 @@ final class MarkdownOutputTests: XCTestCase {
...
+        
+        let (node, _) = try await markdownOutput(catalog: catalog, path: "Comments")
+        let markdown = node.markdown
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 1"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 2"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 4"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 5"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 6"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 7"))
+        XCTAssertFalse(markdown.contains("COMMENT CONTENT 8"))
```
franklinsch: This is vacuously true since the string doesn't appear in the source markdown, so can probably be removed

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -352,6 +352,17 @@ extension MarkdownOutputMarkupWalker {
...
         
     }
+    
+    // HTML is not included in render JSON output, so is omitted here
+    mutating func visitHTMLBlock(_ html: HTMLBlock) -> () {
+        return
+    }
+    
+    // HTML is not included in render JSON output, so is omitted here
+    mutating func visitInlineHTML(_ inlineHTML: InlineHTML) -> () {
```
franklinsch: I think it would be good to add tests for this as well.

franklinsch: Also FYI, you can omit `-> ()` in both these declarations.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -580,6 +580,64 @@ final class MarkdownOutputTests: XCTestCase {
...
+                <!-- COMMENT CONTENT 3 -->
+                ```
+                
+                Raw HTML in the body should not be kept
+                                
+                <h1>More Complex example</h1>
+
+                  <!-- COMMENT CONTENT 4 -->
+
+                  <p>This paragraph is invisible.</p>
```
franklinsch: Did you want to assert that this doesn't appear in the resulting output?

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
d-ronnqvist: @franklinsch FWIW I think this is a limitation because raw HTML can't be represented in the Render JSON. 

I still think that it makes sense to align with the RenderJSON here but in the bigger picture I consider this to be a bug caused by flaws in the RenderJSON.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -352,6 +352,17 @@ extension MarkdownOutputMarkupWalker {
         }
         
     }
+    
+    // HTML is not included in render JSON output, so is omitted here
+    mutating func visitHTMLBlock(_ html: HTMLBlock) {
+        return
+    }
+    
+    // Inline HTML is stripped of any tags, but included in render JSON output, so the same path is followed here
```
franklinsch: Nit, for clarity:

```suggestion
    // Inline HTML is stripped of any tags, but its content is included in render JSON output, so the same path is followed here
```

franklinsch (approved):
Thanks!

jrturton:
@swift-ci please test

PR #1481, opened 2026-03-23
Synchronize PreviewServer state to fix `start` and `stop` race condition

d-ronnqvist:
Bug/issue #, if applicable: rdar://173135444

## Summary

This fixes a few data races in the PreviewServer reported by the ThreadSanitizer.

## Dependencies

None.

## Testing

Run `swift test --sanitize=thread --quiet --parallel --filter DocCCommandLineTests`. 
It shouldn't report any data races.

## Checklist

- ~[ ] Added tests~

### DIFF
--- a/Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
+++ b/Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
@@ -61,11 +61,22 @@ final class PreviewServer {
         }
     }
     
-    private let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
-    private let threadPool = NIOThreadPool(numberOfThreads: System.coreCount)
-
-    private var bootstrap: ServerBootstrap!
-    internal var channel: (any Channel)!
+    private struct State {
+        let group = MultiThreadedEventLoopGroup(numberOfThreads: System.coreCount)
+        let threadPool = NIOThreadPool(numberOfThreads: System.coreCount)
+        
+        var bootstrap: ServerBootstrap!
+        var channel: (any Channel)!
+    }
+    
+    private var state = Synchronized(State())
+    
+    @_spi(Testing)
+    public var _boundPort: Int? {
+        state.sync {
+            $0.channel.localAddress?.port
+        }
+    }
     
     private let contentURL: URL
     private let fileManager: any FileManagerProtocol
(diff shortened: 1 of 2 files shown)


d-ronnqvist:
@swift-ci please test

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
```diff
@@ -120,76 +131,76 @@ final class PreviewServer {
...
+            $0.threadPool.start()
             
-            // Enable TCP_NODELAY for the accepted Channels
-            .childChannelOption(ChannelOptions.socket(IPPROTO_TCP, TCP_NODELAY), value: 1)
-            .childChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
-            .childChannelOption(ChannelOptions.maxMessagesPerRead, value: 16)
-            .childChannelOption(ChannelOptions.allowRemoteHalfClosure, value: true)
-        
-        // Start the server
-        threadPool.start()
```
d-ronnqvist, line 146: FYI: this diff is from reindenting the code (because it's now called within a `state.sync { }` scope and prefixing `bootstrap`, `channel`, `group`, and `threadPool` with `$0`.

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
```diff
@@ -120,76 +131,76 @@ final class PreviewServer {
...
-            // The given port is not available.
-            switch bindTo {
-                case .localhost(let port): throw Error.portNotAvailable(port: port)
-                default: throw error
-            }
-        } catch {
-            // Cannot bind the given address/port.
-            switch bindTo {
-                case .localhost(let port): throw Error.cannotStartServer(port: port)
-                default: throw error
```
d-ronnqvist, line 166: FYI: I restructured the code here but it behaves the same. 
By doing only one `switch bindTo` check and by catching the errors inside the `case .localhost(let port):` scope I felt that the pattern became clearer that we customize the errors when binding to a port but not the errors when binding to a socket.

Sources/DocCCommandLine/PreviewServer/PreviewServer.swift
```diff
@@ -120,76 +131,76 @@ final class PreviewServer {
...
-            .serverChannelOption(ChannelOptions.socket(SocketOptionLevel(SOL_SOCKET), SO_REUSEADDR), value: 1)
-
-            // Configure the channel handler - it handles plain HTTP requests
-            .childChannelInitializer { [contentURL, fileManager] channel in
-                // HTTP pipeline
-                return channel.pipeline.configureHTTPServerPipeline(withErrorHandling: true).flatMap {
-                    channel.pipeline.addHandler(PreviewHTTPHandler(rootURL: contentURL, fileManager: fileManager))
+        let closeFuture = try state.sync {
+            // Create a server bootstrap
+            $0.bootstrap = ServerBootstrap(group: $0.group)
```
franklinsch, line 136: Nit: I think readability could be improved by naming the closure argument here—it becomes a bit confusing seeing `$0` the deeper down you go, although I can see why an exception could be made because this is a `Synchronized` value.

d-ronnqvist: I don't feel strongly about it. 

I had kept it `$0` because I like that it made it explicit which state was synchronized and what wasn't (because I found it easier to skim for `$0` in that code block than another word/variable name like `state`).

PR #1482, opened 2026-03-23
Stop sharing a global mutable FeatureFlags value

d-ronnqvist:
Bug/issue #, if applicable: rdar://173137641

## Summary

This changes `FeatureFlags` from being a globally shared mutable value to being an immutable per-context value that's passed explicitly to any place that has conditional behavior based on a feature flag.

I separated these changes into two commits to make it easier to review; one that updates everything except the directives (https://github.com/swiftlang/swift-docc/commit/e166ccc78fd370e10d93b525aac85c12fea10067) and one that only updates the directives (https://github.com/swiftlang/swift-docc/commit/f3b8a21333b7bdd97e8ee35bea2a74f8f0af643d). In both commits about two thirds of the changes are in tests, with the directive commit being about twice as big as the everything-else commit:

commit      | where     | files changes | insertions | deletions
-----------|---------|-------------:|----------:|---------:
`e166ccc` | Sources |                 18  |           85 |          54
`e166ccc` | Tests     |                  22  |         164 |           195
`f3b8a21`  |Sources |                 39  |          190 |          121
`f3b8a21` | Tests     |                  42  |         336 |          332

By forcing feature flag values to be passed explicitly throughout the code, these changes expose some of the complexity in the design of directives and semantic analysis. This complexity was always there, it was just more hidden because it was accessed through shared mutable global state.

## Dependencies

None.

## Testing

## Checklist

- [x] ~Added~ Updated tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
+++ b/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
@@ -138,8 +138,7 @@ struct PathHierarchy {
 
                     let node = Node(symbol: symbol, name: symbol.pathComponents.last!)
                     // Disfavor synthesized symbols when they collide with other symbol with the same path.
-                    // FIXME: Get information about synthesized symbols from SymbolKit https://github.com/swiftlang/swift-docc-symbolkit/issues/58
-                    if symbol.identifier.precise.contains("::SYNTHESIZED::") {
+                    if symbol.identifier.isSynthesized {
                         node.specialBehaviors.formUnion([.disfavorInLinkCollision, .excludeFromAutomaticCuration])
                     }
                     nodes[id] = node
@@ -995,4 +994,40 @@ private func assertAllNodes(
         file: file,
         line: line
     )
+}
+
+private extension SymbolGraph.Symbol.Identifier {
+    // FIXME: Get information about synthesized symbols from SymbolKit https://github.com/swiftlang/swift-docc-symbolkit/issues/58
+    var isSynthesized: Bool {
+        return precise.withCString {
+            // This is a rather quick way to check if the symbol's unique identifier string contains the "::SYNTHESIZED::" string or not.
+            // Because this check is performed for every symbol once per language representation and platforms, it's good if it can be fast.
+            
+            // Only Swift types use the "::SYNTHESIZED::" separator, so if this unique identifier doesn't have a "s:" prefix, then we can exit early.
+            let pointer = UnsafeRawPointer($0)
+            guard pointer.hasASCIIPrefix("s:") else {
+                return false
+            }
+            
+            // We know that _if_ the symbol's unique identifier contains this separator, then it also contains another unique identifier afterwards.
+            // Because the "::SYNTHESIZED::" separator is 15 characters and the following unique identifier has to be _at least_ 1 character long,
+            // we know that there's less than 16 characters left in the symbol's unique identifier, we can't find a valid "::SYNTHESIZED::" separator anymore.
+            let lengthToCheck = precise.utf8.count &- 16
+            // We know that the first 2 characters are "s:" so there's no reason to check them again. That colon can't be part of a valid "::SYNTHESIZED::" separator.
+            var offset = 2
+            while offset < lengthToCheck {
+                // Because there will always be at more than 8 characters remaining in the, we can safely check 8 characters at a time when looking for the first ":" without risking reading out-of-bounds.
+                defer { offset += 8 }
+                // Our `ByteMatches` reads 8 bytes and uses bitwise operations to determine if any of the next 8 bytes is a ":"
+                let match = ByteMatches(pointer.loadUnaligned(fromByteOffset: offset, as: UInt64.self), ByteMatches.colonSearchPattern)
+                // If _any_ of the next 8 bytes is a ":", we make larger reads and check if the 15 bytes that starts at that ":" is exactly that separator.
+                if match.hasMatches, pointer.hasASCIIPrefix("::SYNTHESIZED::", byteOffset: offset + match.numberOfLeadingNonMatches) {
+                    return true
+                }
+            }
+            // If we reached the last 16 bytes without finding the separator, then we won't find it.
+            assert(!precise.contains("::SYNTHESIZED::"), "The custom implementation didn't produce the same result as a 'String.contains(_:)' check.")
+            return false
+        }
+    }
 }
--- a/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
+++ b/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift
@@ -1124,15 +1124,14 @@ extension Dictionary: FastJSONDecodable where Key == String, Value: FastJSONDeco
 /// A very low level type that represents matching locations of a byte pattern in a collection of 8 bytes.
 ///
 /// Use this type to look for a specific byte in a sequence, checking 8 bytes at a time.
-private struct ByteMatches: ~Copyable {
+package struct ByteMatches: ~Copyable {
     private let raw: UInt64
     
     /// Creates a new byte match result to find the first byte that matches the given search pattern.
     /// - Parameters:
     ///   - bytes: The 8 consecutive bytes of data to match against the search pattern.
     ///   - searchPattern: The byte to search for, repeated 8 times.
-    @inlinable
-    init(_ bytes: UInt64, _ searchPattern: UInt64) {
+    package init(_ bytes: UInt64, _ searchPattern: UInt64) {
         // To understand how this type finds the locations of the bytes that match the search pattern,
         // consider the a decoder that has advanced to the first byte of the first key in the JSON below:
         //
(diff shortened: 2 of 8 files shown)


d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -313,7 +314,7 @@ public enum RenderBlockContent: Equatable {
             self.lineAnnotations = annotations
         }
 
-        public init(copyToClipboard: Bool = FeatureFlags.current.isExperimentalCodeBlockAnnotationsEnabled, showLineNumbers: Bool = false, wrap: Int, highlight: [Int], strikeout: [Int]) {
+        public init(copyToClipboard: Bool, showLineNumbers: Bool = false, wrap: Int, highlight: [Int], strikeout: [Int]) {
```
anferbui, line 317: This is a backwards incompatible API change, the `copyToClipboard` parameter was previously optional but now it is not.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -1020,7 +1023,7 @@ extension RenderBlockContent: Codable {
             }
             self = .aside(aside)
         case .codeListing:
-            let copy = FeatureFlags.current.isExperimentalCodeBlockAnnotationsEnabled
+            let copy = decoder.userInfo[Self.isExperimentalCodeBlockAnnotationsEnabledUserInfoKey] as? Bool == true
```
anferbui, line 1031: Where does this property of the decoder get set?

Sources/SwiftDocC/Semantics/Article/MarkupConvertible.swift
```diff
@@ -12,6 +12,7 @@ public import Foundation
 public import Markdown
 
 /// A type that can be initialized from markup.
+@available(*, deprecated, message: "This deprecated API will be removed after 6.5 is released.")
 public protocol MarkupConvertible {
```
anferbui, line 16: Why are we deprecating this protocol? Should we do this change separately?

Sources/SwiftDocC/Semantics/DirectiveConvertable.swift
```diff
@@ -53,3 +57,18 @@ public extension DirectiveConvertible {
     }
 }
 
+// Default implementation so that conforming types don't need to implement deprecated API.
+public extension DirectiveConvertible {
+    init?(from directive: BlockDirective, source: URL?, for bundle: DocumentationBundle, problems: inout [Problem]) {
+        self.init(from: directive, source: source, for: bundle, featureFlags: .init(), problems: &problems)
```
anferbui: Should this be `FeatureFlags.current` for backward compatibility?

Sources/SwiftDocC/Semantics/Article/Article.swift
```diff
@@ -130,14 +130,15 @@ public final class Article: Semantic, MarkupConvertible, Abstracted, Redirected,
 
             return nil
         }
+        let featureFlags = FeatureFlags()
```
anferbui: This shadows the `init` parameter with the same name, is this intended?

Sources/SwiftDocC/Model/DocumentationNode.swift
```diff
@@ -769,7 +774,7 @@ public struct DocumentationNode {
         // Prefer content sections coming from an article (documentation extension file)
         var deprecated: DeprecatedSection?
         
-        let (markup, docChunks, _) = Self.contentFrom(documentedSymbol: symbol, documentationExtension: article, engine: engine)
+        let (markup, docChunks, _) = Self.contentFrom(documentedSymbol: symbol, documentationExtension: article, featureFlags: FeatureFlags(), engine: engine)
```
anferbui: This looks like it should pass the `init` parameter `featureFlags` instead.

Sources/SwiftDocC/Infrastructure/ConvertActionConverter.swift
anferbui: `HTMLRenderer` accepts feature flags as part of its init, should we pass them through here?

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -920,6 +926,9 @@ extension PathHierarchy {
         self.tutorialContainer = lookup[identifiers[fileRepresentation.tutorialContainer]]!
         self.tutorialOverviewContainer = lookup[identifiers[fileRepresentation.tutorialOverviewContainer]]!
         
+        // If we're decoding another path hierarchy from a file, then we want to be able to
```
anferbui: This sentence is incomplete.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -69,6 +69,7 @@ public struct ConvertAction: AsyncAction {
     ///   - treatWarningsAsErrors: `true` if the convert action should treat warnings as errors, otherwise `false`.
     ///   - experimentalEnableCustomTemplates: `true` if the convert action should enable support for custom "header.html" and "footer.html" template files, otherwise `false`.
     ///   - experimentalModifyCatalogWithGeneratedCuration: `true` if the convert action should write documentation extension files containing markdown representations of DocC's automatic curation into the `documentationBundleURL`, otherwise `false`.
+    ///   - featureFlags: A collection of feature flags that the convert action can uses to enable or disable certain optional behaviors.
```
anferbui: Nit: Grammar
```suggestion
    ///   - featureFlags: A collection of feature flags that the convert action can use to enable or disable certain optional behaviors.
```

Sources/SwiftDocC/Semantics/Article/MarkupConvertible.swift
d-ronnqvist: The only type that conforms to this protocol is `Article` and it needs to add a feature flags parameter so that it can pass the feature flags to its members. However, that means that it `Article` no longer conform to the protocol. 

I could have deprecated the old initializer requirement, added a new initializer requirement (with the feature flag), and added default implementations for both the old and new initializer so that type don't need to implement the deprecated initializer and so that existing types don't break conformance if they don't implement the new protocol. 

However, this protocol is never used as an existential—there are conforming types to it but never any code that accepts with `any MarkupConvertible` parameters, never any code that extends the protocol, never any generics that are constrained by the protocol, etc. so the protocol itself is completely pointless. 

Because it would take ~10 LOC to do the full deprecate-and-phase-out work to keep this pointless protocol alive but only a single LOC to deprecate the protocol, I opted for the shorter and simpler solution. The benefit is that we also get rid of a pointless protocol.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
d-ronnqvist: Yes, I'm aware. I originally made this API breaking change because I couldn't have a deprecated version with the default value and a non-deprecated version without. However, looking at it now, I realize that it's only API breaking when the developer didn't pass a value, so I can create a deprecated initializer without a `copyToClipboard` so that this isn't API breaking. I'll make that change.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
d-ronnqvist: Nothing ever sets this property. It only exists to allow developers to configure what default value render block code listings should decode. I don't think anyone uses this and it's likely that no-one ever will. 

As an alternative I could let the default be `FeatureFlags().isExperimentalCodeBlockAnnotationsEnabled`. That way developers wouldn't be able to customize how old render block code listings are decoded but the value will automatically reflect the default feature flags value when/if that ever changes in the future.

Sources/SwiftDocC/Semantics/DirectiveConvertable.swift
d-ronnqvist: I would need to deprecate this extension but it looks like I can do that without causing any deprecation warnings in the rest of the code.

Sources/SwiftDocC/Semantics/Article/Article.swift
d-ronnqvist: Thanks for catching this. This was a leftover from the first commit when I hadn't updated the `Article` initializer to have a new parameter yet.

Sources/SwiftDocC/Model/DocumentationNode.swift
d-ronnqvist: Same as https://github.com/swiftlang/swift-docc/pull/1482#discussion_r2999785140

Sources/SwiftDocC/Semantics/Article/MarkupConvertible.swift
d-ronnqvist: Because this deprecation would need to happen before this PR, and because review times even for small PRs are often over a week, I didn't feel like it was worth delaying this main PR on a separate PR that only deprecates this protocol.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/ConvertActionConverter.swift
d-ronnqvist: I removed the default value for that parameter so that all callers have to explicitly pass feature flags.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
anferbui: Thanks for the explanation, keeping the decoder approach sounds good to me.

d-ronnqvist:
@swift-ci please test

PR #1483, opened 2026-03-23
Deprecate DocumentationNode initializer that's only called from tests

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

While working on https://github.com/swiftlang/swift-docc/pull/1482 I noticed that one of the `DocumentationNode` initializers that it added a feature flag parameter was never called (only from a few tests). This PR deprecates that initializer so that we can remove it in the future. 

## Dependencies

None.

## Testing

## Checklist

- [x] ~Added~ tests

### DIFF
--- a/Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
+++ b/Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
@@ -772,8 +772,15 @@ class ReferenceResolverTests: XCTestCase {
             """)
         }
         let (_, context) = try await loadBundle(catalog: catalog)
+        XCTAssertEqual(context.problems.map(\.diagnostic.summary).sorted(), [
+            "'NotFoundArticle' doesn't exist at '/ModuleName/Something'",
+            "'NotFoundSymbol' doesn't exist at '/ModuleName/Something'",
+            "'OtherNotFoundArticle' doesn't exist at '/ModuleName/Something'",
+            "'OtherNotFoundSymbol' doesn't exist at '/ModuleName/Something'",
+            "Resource 'not-found-image' couldn't be found",
+            "Resource 'other-not-found-image' couldn't be found",
+        ])
         let node = try XCTUnwrap(context.documentationCache["some-symbol-id"])
-        
         XCTAssertEqual(node.docChunks.count, 2, "This node has content from both the in-source comment and the documentation extension file.")
         
         var resolver = ReferenceResolver(context: context)


Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
```diff
@@ -750,67 +750,29 @@ class ReferenceResolverTests: XCTestCase {
...
+            # ``Something``
+            
+            Continue the documentation for the "something" class.
+            
+            These other links to ``OtherNotFoundSymbol`` and <doc:OtherNotFoundArticle> also won't resolve.
+            
+            This other image name also won't resolve: ![Some other image that's not found](other-not-found-image)
+            """)
+        }
+        let (_, context) = try await loadBundle(catalog: catalog)
```
anferbui, line 774: Please verify that the context doesn't have any unexpected warnings.

Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
```diff
@@ -824,7 +786,7 @@ class ReferenceResolverTests: XCTestCase {
         do {
             let problem = try XCTUnwrap(problems.first)
             XCTAssertEqual(problem.diagnostic.summary, "Can't resolve 'NotFoundArticle'")
-            XCTAssertEqual(problem.diagnostic.source?.path, "/Users/username/path/to/Something.swift")
+            XCTAssertEqual(problem.diagnostic.source?.path, "/Users/username/path/to/SomeFile.swift")
```
anferbui, line 796: Not a request for a change, but an observation: Given how often we reference this value, I wonder if we should make this property public:
https://github.com/swiftlang/swift-docc/blob/4cd90430f9a02803f0b961741226089b5c86e98b/Sources/DocCTestUtilities/SymbolGraphCreation.swift#L226

Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
d-ronnqvist: If you look either above or above few line below you'll see that this content is full of links to pages that don't exist. The test exists to verify those warnings.

anferbui: I did see that; I still think it is worth verifying that the errors in the context are as expected, at least the number of them, same as we do in other tests.

d-ronnqvist: Fair enough. Added in https://github.com/swiftlang/swift-docc/pull/1483/commits/f8885dd2702f7ab636e855349fab781d81d1adca

Tests/SwiftDocCTests/Infrastructure/ReferenceResolverTests.swift
d-ronnqvist: I think it would be out of scope for this PR.

d-ronnqvist:
@swift-ci please test

PR #1485, opened 2026-03-23
Ensure that pages from dependencies organized into local topics sections display symbol names from the external page's abstract

d-ronnqvist:
Bug/issue #, if applicable: rdar://170721787

## Summary

This fixes an edge case regarding pages with links in their abstract in "DependencyModule" when they are organized into a topic section in "MainModule" (which is another documentation build that depends on "DependencyModule").

_Before_ this fix, the name of the linked symbol didn't display in the summary below the externally organized symbol in the topic section. _With_ this fix, it does:

Before | After
---|---
before | after

## Dependencies

None.

## Testing

There are a handful of steps necessary to reproduce this

1. In _one_ target, create two public symbols where one links to the other in their abstract. For example:
   ```swift
   /// The first class links to the ``Second`` class in its abstract.
   public class First {}

   public class Second {}
   ```
   
2. Build documentation for the first target and pass the `--enable-experimental-external-link-support`. To make things clearer, you can specify `--output-path Dependency.doccarchive` for this build.

3. In _another_ target, create a Topics section that organizes the external symbol with the link in its abstract into the local documentation hierarchy. For example:
   ```md
   ## Topics
   
   - ``/DependencyModuleName/First``
   ```

4. Build documentation for the second target and pass both `--enable-experimental-external-link-support` and `--dependency Dependency.doccarchive` . To make things clearer, you can specify `--output-path Main.doccarchive` for this build.

5. Create a combined archive for the two targets using `docc convert merge Dependency.doccarchive Main.doccarchive --output-path Combined.doccarchive`

6. Preview the local documentation using a local web server of your choice. Note because this is a combined archive you cannot use `docc preview` to preview it.

7. Go to the "Main" page that curated the external symbol with the link in its abstract. The abstract should display that link—and not omit the linked page's name completely (see "before the fix" example above)

rendered page with external link in topic section with symbol in its abstract

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
+++ b/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
@@ -352,6 +352,17 @@ extension MarkdownOutputMarkupWalker {
         }
         
     }
+    
+    // HTML is not included in render JSON output, so is omitted here
+    mutating func visitHTMLBlock(_ html: HTMLBlock) {
+        return
+    }
+    
+    // Inline HTML is stripped of any tags, but its content is included in render JSON output, so the same path is followed here
+    mutating func visitInlineHTML(_ inlineHTML: InlineHTML)  {
+        return
+    }
+
 }
 
 // Semantic handling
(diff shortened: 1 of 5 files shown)


d-ronnqvist:
@swift-ci please test

PR #1486, opened 2026-03-24
Add option to install without building.

3405691582:
Bug/issue #, if applicable: 

## Summary

This allows the build action to be decoupled from the install action, so packaging systems that expect to be able to do discrete installation steps post-build can do so cheaply, without having to effectively start the build over from scratch.

To match existing behavior, this defaults to being disabled. This was introduced previously in swiftlang/swift-package-manager#9050, and the flag has the same name used there.

## Dependencies

No hard dependencies. Changes are being upstreamed to introduce this behavior on some other projects (as mentioned the first change was swiftlang/swift-package-manager#9050, see also e.g., swiftlang/swift-format#1171, swiftlang/sourcekit-lsp#2579).

## Testing

Run build-script-helper's install action without the flag, i.e., use build-script normally. The script's behavior will be unchanged because the flag is disabled by default.

Run build-script-helper's install action with the flag, pointing at the built build artifacts.
The installation will just perform the installation steps without rebuilding, i.e., copy the binaries.

## Checklist

- [n/a] Added tests
- [n/a] Ran the `./bin/test` script and it succeeded


heckj:
@swift-ci please test

Issue #1217, opened 2025-05-14
Docc convert fails in docker image with swift 6


tverboon:
### Description

I am working on a Swift docc project that consists of only manually created markdown files (`--template articleOnly`). No symbols or something else.

We want to convert the docc with a GitHub action for static hosting. When using the Swift 6.1 or 6.0 docker base image, this operation fails.

Using Swift 5.9 as base image, convert succeeds. Very nice! 👍

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

The docc convert operation succeeds, like it did with Docker image `swift:5.9`

### Actual behavior

The docc convert operation failes with the following error:
 
```
Error: Error Domain=NSCocoaErrorDomain Code=512 "(null)"UserInfo={NSSourceFilePathErrorKey=/github/workspace/src/Midmid.Platform.Api.Docs/docs.docc/Resources/optional.svg, NSUserStringVariant=["Copy"], NSDestinationFilePath=/tmp/C445329D-DFCE-4573-B177-6AE338E38B10-7-00001ED6DC3F9EB7/images/docs/optional.svg, NSFilePath=/github/workspace/src/Midmid.Platform.Api.Docs/docs.docc/Resources/optional.svg, NSURL=file:///github/workspace/src/Midmid.Platform.Api.Docs/docs.docc/Resources/optional.svg, NSUnderlyingError=Error Domain=NSPOSIXErrorDomain Code=22 "Invalid argument"}
```

### Steps To Reproduce

I will provide my scripts and docker file:

### Dockerfile

```dockerfile
FROM swift:6.1

COPY run.sh /run.sh

ENTRYPOINT ["/run.sh"]
```

```bash
#!/usr/bin/env bash

echo "Building documents in: $1"

docc convert /github/workspace/$1 --transform-for-static-hosting --output-dir /github/workspace/$2

echo "Build result in: $2"
```

### Swift-DocC Version Information

6.1

### Swift Compiler Version Information

```shell
Swift version 6.1 (swift-6.1-RELEASE)
Target: aarch64-unknown-linux-gnu
```

heckj:
I think I've just run into the same issue (or very close) - same underlying issue:

[...]

PR #1491, opened 2026-03-27
fix for recursive copy issue on Linux

heckj:
Bug/issue #, if applicable: #1217

## Summary

There's a workaround in the code for [swift-foundation#1125](https://github.com/swiftlang/swift-foundation/issues/1125), introduced in swiftlang/swift-docc#1169 that didn't accommodate a recursive copy failure mode.

With this update, when the workaround ignores a `CocoaError` for a directory copy, it walks the source tree and individually copies the missing children. 
Each individual file copy goes through `_copyItem` itself, so per-file `fchown` failures are correctly handled (the single file exists at the destination after content is written, before the attribute-setting step fails).

## Testing

I wasn't sure about idea path for testing, and this relies on some workaround that only happen in Linux, so the test builds up custom "FileManager" that exhibits the same behavior we're seeing on Linux with the foundation bug. If you comment out lines 161-163 in the FileManagerProtocol.swift, the test illustrates the failure identical to how I encountered the original issue, and passes with this fix in place.

## Checklist

- [x] Added tests

I'm fairly confident that this issue goes back quite a few editions of Swift - I just hadn't been able to hunt it down before. If this PR is acceptable for a fix, I'd like to propose that it also get cherry picked for a 6.3 release branch back-port.


Sources/SwiftDocC/Utility/FileManagerProtocol.swift
```diff
@@ -150,19 +150,54 @@ extension FileManager: FileManagerProtocol {
...
+    /// This handles the case where `copyItem(at:to:)` fails mid-copy when copying a directory,
+    /// leaving some children uncopied. Each missing child is copied individually so that a
+    /// per-file `fchown` failure doesn't prevent copying the remaining items.
+    private func _copyMissingChildren(from source: URL, to destination: URL) throws {
+        let sourceChildren: [String]
+        do {
+            sourceChildren = try contentsOfDirectory(atPath: source.path)
+        } catch {
+            return // Can't list source — nothing more we can do
+        }
```
d-ronnqvist: Question: do you know of a reason why the file manager would fail to list the source contents? Is this working around a bug or would it make sense to surface this error to the caller (remove the do-catch blocks)?

Tests/SwiftDocCTests/Utility/FileManagerCopyItemTests.swift
```diff
@@ -0,0 +1,154 @@
...
+                       "favicon.svg should exist")
+        XCTAssertTrue(fm.fileExists(atPath: destDir.appendingPathComponent("index.html").path),
+                       "index.html should exist")
+        XCTAssertTrue(fm.fileExists(atPath: destDir.appendingPathComponent("index-template.html").path),
+                       "index-template.html should exist")
+
+        // Verify content integrity
+        let copiedHTML = try Data(contentsOf: destDir.appendingPathComponent("index.html"))
+        XCTAssertEqual(String(decoding: copiedHTML, as: UTF8.self),
+                       "<html><head></head><body></body></html>")
```
d-ronnqvist, line 108: We have some test helpers for creating and verifying file content that I think could make it easier to see what this test is verifying (that the files within a copied directory are all copied when `fchown` fails).

Using those test helpers you can define the "source" as a loop of unspecific files and verify that that entire hierarchy exist at the other location after the copy:

```swift
// A source directory that contains a number of top-level files without any directories.
let source = Folder(name: "source") {
    for number in 0...10 {
        TextFile(name: "file-\(number)", utf8Content: "File \(number)")
    }
}
let tempDir = try createTemporaryDirectory()
let sourceDir = try source.write(inside: tempDir)
let destDir = tempDir.appendingPathComponent("dest")

let fm = PartialCopyFileManager()

// Without the fix, _copyItem would silently return after the partial copy,
// leaving all but "file-0" missing from the destination.
try fm._copyItem(at: sourceDir, to: destDir)

source.assertExist(at: destDir)
```

d-ronnqvist: The test below would look almost the same but with a different definition of the `source`. For example, like below (or however complex you want the directory hierarchy to be):
```swift
// A source directory that contains a number of files in a directory hierarchy.
let source = Folder(name: "source") {
    Folder(name: "even") {
        for number in 0...5 {
            let number = number * 2
            TextFile(name: "file-\(number)", utf8Content: "Even file \(number)")
        }
    }
    Folder(name: "deep") {
        Folder(name: "path") {
            Folder(name: "to") {
                Folder(name: "odd") {
                    for number in 0..<5 {
                        let number = (number * 2) + 1
                        TextFile(name: "file-\(number)", utf8Content: "Even file \(number)")
                    }
                }
            }
        }
    }
}
```

d-ronnqvist: We seldom use the `assertExist(at:)` helper anymore, so we haven't updated it to support Swift Testing yet. Because of this, it probably makes sense for these new tests to use XCTest.

Tests/SwiftDocCTests/Utility/FileManagerCopyItemTests.swift
```diff
@@ -0,0 +1,154 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2025 Apple Inc. and the Swift project authors
```
sofiaromorales:
```suggestion
 Copyright (c) 2026 Apple Inc. and the Swift project authors
```

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
heckj: Can absolutely pitch the error upwards here if we prefer. This specific call isn't working around anything that I know of - more that it *can* throw an error (but doesn't in the workaround scenario I've hit) 

Since this method is entirely the workaround shim to copy in children to deal with the oddness of the upstream issue in Foundation, I was trying to keep it's "API surface" to a minimum - we'd already be in a failure state due to that bug previously, and I didn't want to compound anything, so went with silent results here. If we can't get the contents of the directory to copy in the files within it, felt like we were pretty much dead in the water anyway.

heckj:
@swift-ci please test

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
heckj: Simplified and throws the error now

All the call paths source up to actions directly, and theoretically have file path related information in them as standard "Foundation" errors, so this should provide good specifics if we DO hit that scenario for any reason.

heckj:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

PR #1492, opened 2026-03-28
Ensure theme-settings.json is always present in build output

Himadrimani:
Fixes swiftlang/swift-docc-render#1000

Requests for 'theme-settings.json' from the web renderer resulted in 404 errors 
if the file was not explicitly provided.

This change ensures:
- An empty 'theme-settings.json' is always generated during conversion
- The preview server returns an empty JSON object instead of 404

Added regression and handler tests to validate behavior.


d-ronnqvist:
I wonder if it would make more sense for Swift-DocC Render repo to include an empty theme settings file in the template. That way the contents of the template can remain opaque to Swift-DocC.

Himadrimani:
> I wonder if it would make more sense for Swift-DocC Render repo to include an empty theme settings file in the template. That way the contents of the template can remain opaque to Swift-DocC.

Good point, thanks. I’ll move this concern to the Render template side and update this PR accordingly.

d-ronnqvist:
Closing this because https://github.com/swiftlang/swift-docc-render/pull/1001 solves this issue in DocC-Render instead.

PR #1493, opened 2026-03-28
Enable snippets beyond Swift: resolve files outside documentation catalog

### LINKED ISSUE #1436: The ability to include source content (markdown, or programming sources) from outside the catalog (beyond Swift snippets)
### Feature Name

snippets-beyond-swift

### Description

I'd like to have an easy path so that content that is stored within a Swift package can be easily referenced and used within the content presented in a DocC catalog. 

### Motivation

The two primarily use cases here are markdown files that aren't housed within the catalog due to overlap and value in their location within a broader project structure, and including source code from languages other than Swift, which is all the snippets provides today.

For markdown files, being able to reference and/or include the README while leaving it at the root of the git repo (typically VERY outside of a catalog) has been an extremely common request. And this just as easily applies to generated sources that might be part of a build process that aren't easy to incorporate into DocC content documentation today.

The source file desire comes from wanting to lean into the broader interop stories for Swift - specifically C and C++, but it applies to any other possible language, where today it's impossible to reference a validated, tested source file - or a slice of such a file, akin to how Snippets allows you to slice a Swift file with comments - in a DocC article. Being able to define something within a catalog that points to a specific source file, and use that as a file that can be converted into a "snippet" for inclusion within any of the DocC catalogs that may be supported within a project, would open up documenting how to support interop with other languages significantly from where we are today - which relies on the high maintenance "copy the snippet into place".

Additionally, something beyond the snippets directory would support allowing a more direct "here's how this code works" linkage, referencing the source files and displaying portions of them inline with content, allowing more contributor-focused documentation to be able to be created.

### Importance

Reduces the friction for managing source code snippets, particularly for non-Swift languages, and include their content in DocC articles.

### Alternatives Considered

Manually copying and maintaining a replica of any relevant source code into a DocC catalog.

Himadrimani:
Fixes #1436

This change extends the @Snippet directive to support resolving source content from outside the documentation catalog.

Key features:
- SnippetResolver now falls back to filesystem resolution relative to the project root (checkoutPath).
- Added support for parsing // snippet.<name> markers in non-Swift files.
- Automatically detects language based on file extension for proper syntax highlighting.
- Relaxed validation for command-line arguments to allow providing --checkout-path independently.

Added comprehensive unit tests in ExternalSnippetTests.swift.

### DIFF
--- a/Sources/DocCHTML/MarkdownRenderer+Topics.swift
+++ b/Sources/DocCHTML/MarkdownRenderer+Topics.swift
@@ -94,7 +94,7 @@ package extension MarkdownRenderer {
     }
     
     private func _taskGroupItem(for element: LinkedElement) -> XMLElement {
-        var items: [XMLNode]
+        let items: [XMLNode]
         switch element.subheadings {
         case .single(.conceptual(let title)):
             items = [.element(named: "p", children: [.text(title)])]
@@ -122,15 +122,17 @@ package extension MarkdownRenderer {
             }
         }
         
-        // Add the formatted abstract if the linked element has one.
-        if let abstract = element.abstract {
-            items.append(visit(abstract))
-        }
-        
-        return .element(named: "li", children: [
-            // Wrap both the name and the abstract in an anchor so that the entire item is a link to that page.
+        let listItem = XMLNode.element(named: "li", children: [
+            // DocC-Render only makes the item's name an anchor, not its abstract
             .element(named: "a", children: items, attributes: ["href": path(to: element.path)])
         ])
+        
+        // Add the formatted abstract if the linked element has one.
+        if let abstract = element.abstract {
+            listItem.addChild(visit(abstract))
+        }
+        
+        return listItem
     }
     
     /// Transforms the symbol name fragments into a `<code>` HTML element that represents a symbol's subheading.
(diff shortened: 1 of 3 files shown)


heckj:
@Himadrimani please be careful to only include commits for the content under question within a PR. This one includes what appears to be two different PRs in one go. This shouldn't include the proposal for how to fix the theme settings.

d-ronnqvist (commented):
This changes the design of snippets in DocC and makes two different components both responsible for parsing the snippet source (the snippets-extract tool in [the Swift-DocC Plugin](https://github.com/swiftlang/swift-docc-plugin/) and now also Swift-DocC itself). 

These type design changes (which are also changes in user-facing behavior) need to be discussed in the Swift Forums before this PR can move forward.

Himadrimani:
Thanks for the review — agreed. I’ll split out unrelated commits, keep this PR scoped only to snippet resolution changes, and resolve conflicts. I’ll also open a Swift Forums thread for the design discussion and link it here.

Himadrimani:
I opened the Swift Forums discussion here: https://forums.swift.org/t/swift-docc-proposal-allow-snippet-to-resolve-files-outside-documentation-catalog/85737.

I also force-pushed a cleaned branch:

* Removed unrelated theme-settings commit(s)
* Kept only snippet-related changes
* Rebased on latest main and resolved conflicts

PR #1499, opened 2026-04-01
Ignore nodes with empty titles in indexing records

snprajwal:
Bug/issue #, if applicable: rdar://171823225

## Summary

Sometimes, anonymous C symbols can be incorrectly included in the symbol graph file with an empty title and path components. When these symbols are registered in DocC, their render nodes are created successfully, but the lack of a title means that they cannot be reasonably represented in either the navigator or the indexing records. These issues currently cause DocC to emit an irrecoverable error and have the build fail. However, they do not affect DocC's ability to build documentation correctly, and hence do not warrant the severity that they are currently emitted at. This patch updates both the navigator index and indexing record methods to include a debug assertion failure, and return nil when running in release mode to prevent terminating the build.

## Dependencies

N/A

## Testing

N/A. Debug assertions are relevant only when developing.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Indexing/IndexingError.swift
+++ b/Sources/SwiftDocC/Indexing/IndexingError.swift
@@ -10,6 +10,7 @@
 
 import Foundation
 
+@available(*, deprecated, message: "This error type is no longer used. This deprecated API will be removed after 6.5 is released.")
 public enum IndexingError: DescribedError {
     /**
      A page or other piece of documentation doesn't have a title, so it can't
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -55,6 +55,7 @@ public class NavigatorIndex {
         case missingBundleIdentifier
         
         /// A RenderNode has no title and won't be indexed.
+        @available(*, deprecated, message: "This error type is no longer used. This deprecated API will be removed after 6.5 is released.")
         case missingTitle(description: String)
         
         /// The navigator index has not been initialized.
@@ -712,8 +713,11 @@ extension NavigatorIndex {
                 return nil // skip as item exists already.
             }
             
-            guard let title = usePageTitle ? renderNode.metadata.title : renderNode.navigatorTitle() else {
-                throw Error.missingTitle(description: "\(renderNode.identifier.absoluteString.singleQuoted) has an empty title and so can't have a usable entry in the index.")
+            guard let title = usePageTitle ? renderNode.metadata.title : renderNode.navigatorTitle(), !title.isEmpty else {
+                // Nodes without a title are erroneous entries in the symbol graph.
+                // An index entry cannot be constructed without a title, so the node is skipped.
+                assertionFailure("\(renderNode.identifier.absoluteString.singleQuoted) has an empty title, and cannot have a usable index entry")
+                return nil
             }
             
             // Get the identifier path
--- a/Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
+++ b/Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
@@ -34,7 +34,7 @@ extension RenderNode {
 }
 
 extension RenderNode: Indexable {
-    func topLevelIndexingRecord() throws -> IndexingRecord {
+    func topLevelIndexingRecord() -> IndexingRecord? {
         let kind: IndexingRecord.Kind
         switch self.kind {
         case .tutorial:
@@ -50,8 +50,11 @@ extension RenderNode: Indexable {
         }
         
         guard let title = metadata.title, !title.isEmpty else {
-            // We at least need a title for a search result.
-            throw IndexingError.missingTitle(identifier)
+            // Nodes without a title are erroneous entries in the symbol graph.
+            // A search result cannot be constructed without a title, meaning that
+            // an indexing record for this node is useless. The node is skipped.
+            assertionFailure("\(identifier.absoluteString.singleQuoted) has an empty title, and cannot have a usable search result")
+            return nil
         }
         
         let summaryParagraph: RenderBlockContent?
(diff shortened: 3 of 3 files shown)


snprajwal:
@swift-ci please test

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -712,8 +712,15 @@ extension NavigatorIndex {
                 return nil // skip as item exists already.
             }
             
-            guard let title = usePageTitle ? renderNode.metadata.title : renderNode.navigatorTitle() else {
-                throw Error.missingTitle(description: "\(renderNode.identifier.absoluteString.singleQuoted) has an empty title and so can't have a usable entry in the index.")
+            guard let title = usePageTitle ? renderNode.metadata.title : renderNode.navigatorTitle(), !title.isEmpty else {
+                // Nodes without a title are erroneous entries in the symbol graph.
+                // An index entry cannot be constructed without a title, so the node is skipped.
+                self.problems.append(Problem(diagnostic: Diagnostic(
```
d-ronnqvist: This isn't a user-actionable issue so it shouldn't be a diagnostic.

Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
```diff
@@ -50,8 +50,15 @@ extension RenderNode: Indexable {
         }
         
         guard let title = metadata.title, !title.isEmpty else {
-            // We at least need a title for a search result.
-            throw IndexingError.missingTitle(identifier)
+            // Nodes without a title are erroneous entries in the symbol graph.
+            // A search result cannot be constructed without a title, meaning that
+            // an indexing record for this node is useless. The node is skipped.
+            problems.append(Problem(diagnostic: Diagnostic(
```
d-ronnqvist: This isn't a user-actionable issue so it shouldn't be a diagnostic.

Sources/SwiftDocC/Indexing/IndexingError.swift
```diff
@@ -10,6 +10,7 @@
 
 import Foundation
 
+@available(*, deprecated, message: "This error type is no longer used, and will be removed in Swift 6.4")
```
d-ronnqvist: This removal is too soon. The upcoming release is 6.4 so the earliest we would remove this is after 6.5.

Also, please use the same phrasing as the other deprecation messages. That makes it easier to search for and find them all when it's time to remove them.

Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
```diff
@@ -68,20 +75,34 @@ extension RenderNode: Indexable {
         return IndexingRecord(kind: kind, location: .topLevelPage(identifier), title: title, summary: summary, headings: self.headings, rawIndexableTextContent: self.rawIndexableTextContent, platforms: metadata.platforms)
     }
     
+    @available(*, deprecated, message: "This method will be removed in Swift 6.4, use ``RenderNode.indexingRecords(onPage:problems:)`` instead")
```
d-ronnqvist: Same as above; this removal is too soon. The upcoming release is 6.4 so the earliest we would remove this is after 6.5.

Also, please use the same phrasing as the other deprecation messages. That makes it easier to search for and find them all when it's time to remove them.

d-ronnqvist:
Note that the "Testing" section of the pull request template is intended for instructions for how a human can verify the changes end-to-end. This is in addition to the expectation that all changes are covered by automated tests.

Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
snprajwal: Do we have an alternative that allows this to be visible while not erroring out like before? We can silently drop it, but I'm not sure if that's a good idea given that it's _expected_ for every node to have a title.

d-ronnqvist: I don't think we have any mechanism for reporting issues that aren't intended for the user. 

The only similar example that I can think of was [this previous symbol graph validation error](https://github.com/swiftlang/swift-docc/pull/966). It used to be a non-user-facing error that we first lowered to be a warning but before merging the team discussed it more offline and agreed that because it was confusing and unactionable we'd make it a debug assertion instead (which is what the merged PR did in the end).

snprajwal: I think that makes sense here too. This is neither actionable nor build-breaking, the only reason we would want to surface this is so that the user is aware that there's either an incorrect entry in the symbol graph (if handwritten) or an issue with the tool generating the symbol graph. I'll update this. It would also mean we don't have to change the public API.

snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test

Sources/SwiftDocC/Indexing/IndexingError.swift
```diff
@@ -10,6 +10,7 @@
 
 import Foundation
 
+@available(*, deprecated, message: "This error type is no longer used. It will be removed in Swift 6.5")
```
d-ronnqvist: minor: If we use the same phrasing in these deprecations as elsewhere it will be easier to find them when it's time to remove them

```suggestion
@available(*, deprecated, message: "This error type is no longer used. This deprecated API will be removed after 6.5 is released.")
```

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -55,6 +55,7 @@ public class NavigatorIndex {
         case missingBundleIdentifier
         
         /// A RenderNode has no title and won't be indexed.
+        @available(*, deprecated, message: "This error type is no longer used. It will be removed in Swift 6.5")
```
d-ronnqvist: minor: If we use the same phrasing in these deprecations as elsewhere it will be easier to find them when it's time to remove them

```suggestion
        @available(*, deprecated, message: "This error type is no longer used. This deprecated API will be removed after 6.5 is released.")
```

Sources/SwiftDocC/Indexing/IndexingError.swift
snprajwal: Ah yes, sorry. I thought you meant using the same wording between the multiple messages I added previously. Updated!

snprajwal:
@swift-ci please test

PR #1504, opened 2026-04-05
Add support for module-name selectors (::) in documentation links

Himadrimani:
This PR adds support for the :: separator in documentation links, consistent with the Swift 6.4 module selector syntax.

- Updated PathComponentScanner to recognize :: as a separator.
- Updated PathParser to treat links starting with ModuleName:: as absolute links.
- Added unit tests to ModuleSelectorTests.swift.

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
+++ b/Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
@@ -164,7 +164,7 @@ extension PathHierarchy.PathParser {
         // the path parser can identify that "ModuleName" (without the leading slash) is a valid module name, but "=(_:_:)" (without the leading slash) isn't.
         // This tells the parser to remove the leading slash from "/ModuleName" and return that this link string represented an absolute link, whereas it leaves
         // "/=(_:_:)" as-is and returns that that link string represented a relative link.
-        let isAbsolute: Bool
+        var isAbsolute: Bool
         if path.first == PathComponentScanner.separator {
             guard let maybeModuleName = components.first?.dropFirst(), !maybeModuleName.isEmpty else {
                 return ([], true, nil)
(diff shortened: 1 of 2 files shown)


d-ronnqvist (changes requested):
Thank you for this pull request. However, this is not a change that DocC needs. 

If we wanted to make this change it would require a discussion in the Swift Forums—like two of your other recent PRs—because it makes user-facing changes to syntax.

I thought that this had already been talked about in the pitch/discussion thread about module separators but I checked and the conversation I was thinking of happened offline. Here's a brief summary of that conversation for future reference:
- DocC only uses symbol-relative links (for example ` ``SomeTypeName`` ` or ` ``someMethod()`` ` ) for the current module (that the documentation is about)
- All links to symbols in other modules are "absolute"; starting with a leadings slash, the other module name, and the rest of the path to that symbol (` ``/OtherModule/SomeType`` `)
- It's possible to use absolute links (with a leading slash) to symbols in the current module as well. This means that when there's a top-level symbol with the same name as the containing module, it can be unambiguously referred to using an absolute link ` ``/Something/Something`` `.
- If we wanted to introduce a dedicated `::` separator to the link syntax for the purposes of closer resembling the Swift syntax we should probably revisit the `/` separator character first. That would be a much larger discussion.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy+PathComponent.swift
```diff
@@ -328,13 +334,34 @@ private struct PathComponentScanner: ~Copyable {
...
+                index = m
+                separatorLength = Self.moduleSeparator.count
+            }
+        case (let s?, nil):
+            index = s
+            separatorLength = 1
+        case (nil, let m?):
+            index = m
+            separatorLength = Self.moduleSeparator.count
+        case (nil, nil):
```
d-ronnqvist, line 358: If we wanted to support a `::` separator that specifically separates the module name from the first symbol name, we likely wouldn't implement it like this because this implementation supports the `::` separator every where in the link (even between symbols and their members).

d-ronnqvist:
This PR is increasingly looking to be abandoned. There has been no activity on this PR for over a month and as far as I can tell there hasn't been any Swift Forums post pitching this. 

Feel free to reopen this PR either after or in parallel to a Swift Forums thread that pitch this idea. In the meantime I will close this PR to reduce noise in the list of open PRs.

PR #1501, opened 2026-04-02
Remove workaround for unnamed C symbols (#1100)

snprajwal:
## Summary

This reverts commit 093049d1c34869a703eee05785af91eebbaacb65.

The issue has been fixed upstream in https://github.com/llvm/llvm-project/pull/120801. The previous attempt to remove this workaround in #1456 discovered that there were some code paths that incorrectly depended on this workaround and broke the build otherwise. It was reintroduced in #1467 while the issue was being investigated. The parts that depended on this workaround have been removed in #1499.

## Dependencies

N/A

## Testing

Create a symbol graph containing a symbol with an empty title, and run a documentation build with the `--emit-digest` flag. The build should not error out or terminate in release mode, and should crash with a debug assertion failure in debug mode.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

~- [ ] Added tests~
~- [ ] Ran the `./bin/test` script and it succeeded~
~- [ ] Updated documentation if necessary~

PR #1506, opened 2026-04-16
Avoid creating a cyclic topic graph when symbols are unexpectedly members of themselves

d-ronnqvist:
Bug/issue #, if applicable: rdar://174319706

## Summary

This updates the logic for synthesized protocol implementation pages to avoid creating a cyclic topic graph when the symbol graph contains unexpected relationships that define a symbol a the member of itself.

## Dependencies

None.

## Testing

I don't know what makes the Swift symbol graph extractor create such unexpected relationships, so in order to test this you would need to manually modify a symbol graph file to include a `SymbolGraph.Symbol.Swift.Extension` trait for a symbol and a `memberOf` relationship to itself for that symbol that also include a `SymbolGraph.Relationship.SourceOrigin` trait—like the setup in `testRelationshipToSelfDoesNotCauseCyclicTopicGraph()`

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Symbol Graph/GeneratedDocumentationTopics.swift
+++ b/Sources/SwiftDocC/Infrastructure/Symbol Graph/GeneratedDocumentationTopics.swift
@@ -145,6 +145,7 @@ enum GeneratedDocumentationTopics {
         // Curate all inherited symbols under the collection node
         for childReference in identifiers {
             if let childTopicGraphNode = context.topicGraph.nodeWithReference(childReference) {
+                assert(childReference != parent, "Parent type '\(parent.path)' is unexpectedly included among its members: [\(identifiers.map(\.path).sorted().joined(separator: ", "))]")
                 context.topicGraph.addEdge(from: collectionTopicGraphNode, to: childTopicGraphNode)
             }
         }
@@ -242,7 +243,9 @@ enum GeneratedDocumentationTopics {
                // Get the child symbol
                let childSymbol = child.symbol,
                // Check that there is Swift extension information
-               childSymbol[mixin: SymbolGraph.Symbol.Swift.Extension.self] != nil
+               childSymbol[mixin: SymbolGraph.Symbol.Swift.Extension.self] != nil,
+               // Ignore any faulty relationships
+               relationship.source != relationship.target
             {
                 let originSymbol = context.documentationCache[origin.identifier]?.symbol
                 
@@ -255,6 +258,8 @@ enum GeneratedDocumentationTopics {
         for (typeReference, collections) in inheritanceIndex.implementingTypes where !collections.inheritedFromTypeName.isEmpty {
             for (_, collection) in collections.inheritedFromTypeName where !collection.identifiers.isEmpty {
                 // Create a collection for the given provider type's inherited symbols
+                assert(!collection.identifiers.contains(typeReference), "Parent type '\(typeReference.path)' is unexpectedly included among its members: [\(collection.identifiers.map(\.path).sorted().joined(separator: ", "))]")
+                
                 try createCollectionNode(parent: typeReference, title: collection.title, identifiers: collection.identifiers, context: context)
             }
         }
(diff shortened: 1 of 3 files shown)


d-ronnqvist:
@swift-ci please test

snprajwal (approved):
LGTM, this is a good fix beyond the bug itself too -- there are languages that support directly recursive types (e.g. TypeScript), this will let us work with such symbol graphs reliably.

Sources/SwiftDocC/Infrastructure/Symbol Graph/GeneratedDocumentationTopics.swift
```diff
@@ -242,7 +243,9 @@ enum GeneratedDocumentationTopics {
                // Get the child symbol
                let childSymbol = child.symbol,
                // Check that there is Swift extension information
-               childSymbol[mixin: SymbolGraph.Symbol.Swift.Extension.self] != nil
+               childSymbol[mixin: SymbolGraph.Symbol.Swift.Extension.self] != nil,
+               // Ignore any faulty relationships
+               relationship.source != relationship.target
```
patshaughnessy, line 248: This is great to detect when a symbol has a relationship to itself. But are there any more complex cycles that we should detect? Such as a grandparent symbol is a member of one of its grandchildren?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1248,17 +1248,6 @@ public struct RenderNodeTranslator: SemanticVisitor {
         var node = RenderNode(identifier: identifier, kind: .symbol)
         var contentCompiler = RenderContentCompiler(context: context, identifier: identifier)
         
-        /*
```
patshaughnessy, line 1251: Was this really just dead code? Anything to do with the invalid relationship you identify in GeneratedDocumentationTopics.swift?

patshaughnessy (approved):
Thanks for the fix!

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
d-ronnqvist: Yeah, this was just dead code. Nothing used the local `ref` value other than the loop that computed the ref value based on itself.

Sources/SwiftDocC/Infrastructure/Symbol Graph/GeneratedDocumentationTopics.swift
d-ronnqvist: I don't know of any programming language where that would be valid. I don't even think most languages would allow more than one definition of a type so I'm having a hard time imagining a hypothetical syntax for how a cyclic type definition would even be defined.

Note that types that _reference_ themselves or reference one of their member types is totally valid but the compilers for those languages shouldn't emit `memberOf` relationships for those references.

----

Initially I added a debug assertion in `TopicGraph.addEdge(from:to:)` that verified that there wasn't a cycle in the graph after adding a edge but we hit that assertion in some of our tests that verify the diagnostics raised about forming cyclic curation, so I had to remove that assertion in order for DocC to be able to raise that diagnostic in debug builds (and in order for the tests to pass without asserting). 

As far as I know this is the first time we've encountered an issue with cycles in the symbol graph relationships. If it becomes a problem again in the future we can see if there are more safeguards that we can add.

d-ronnqvist:
> [...] -- there are languages that support directly recursive types (e.g. TypeScript), this will let us work with such symbol graphs reliably.

There are many programming languages that support recursive types, including Swift, but I don't think any of them would model it as the type _defining_ itself within its own definition. Instead, I would imagine that those types _reference_ their own type name within their own definition. The latter (referencing a type in a property, enum case, etc.) wouldn't use a `memberOf` relationship to the type so that type of recursiveness wouldn't have this problem.

snprajwal:
Right, so this relationship is possible only in the impossible case of re-defining a type within its own definition. Given that this looks to be a bug in Swift's symbol graph extraction, do we want to track this change as something we'd revert once we fix it at the source? I'm not sure if DocC should be responsible for handling cases like this when we reasonably expect the input to be well-formed (somewhat along the lines of #1501, where a bug in Clang necessitated a workaround in DocC).

d-ronnqvist:
It could make sense to remove the equality check once the Swift compiler fix has been live for long enough but I don't think it's too bad to keep it as a guard against unexpected inputs. 

If we remove it in the future I think it would be good to keep the assertions and the removal of dead code, rather than revert all the changes.

PR #1497, opened 2026-03-31
Add lots of test covering a variety of current availability behaviors

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This adds lots of test covering a variety of current availability behaviors ahead of an upcoming planned refactor of the availability code.

## Dependencies

None.

## Testing

Nothing in particular. This PR doesn't change any behaviors. It only adds a number of new automated tests.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- ~[ ] Updated documentation if necessary~

PR #1507, opened 2026-04-16
Do not inherit default availability if it is present in-source

snprajwal:
Bug/issue #, if applicable: rdar://174841126

## Summary

With the availability refactor in #1470, we now always use the default availability, and override it with the in-source availability if present for a platform. However, this requires a symbol to be explicitly marked unavailable for that platform. If a symbol graph is present for a platform and does not contain a symbol, we still show the symbol as being available on that platform if it is present in the default availability list. This patch modifies the logic to only include the default availability if there is no in-source availability information at all, and to otherwise start with an empty list and populate the in-source availability

## Dependencies

N/A

## Testing

1. Create a catalog with symbol graphs for more than one platform.
2. Include default availability for all the platforms in `Info.plist`.
3. Include a platform-specific symbol in only one symbol graph.
4. Without this patch, the symbol is shown to be available on all platforms present in the default availability.
5. With this patch, the symbol is shown to only be available on the one platform it is available on.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -1295,10 +1295,11 @@ public struct RenderNodeTranslator: SemanticVisitor {
         
         node.metadata.platformsVariants = VariantCollection<[AvailabilityRenderItem]?>(from: symbol.availabilityVariants) { _, inSourceAvailability in
             // Different sources of availability information are added in-order to compute the complete availability information.
+
+            // The default availability is merged with the in-source availability when loading a symbol graph (see ``SymbolGraphLoader.addDefaultAvailability(to:moduleName:)``).
+            // If no in-source availability is present, we fall back to the default availability for the module (in Info.plist).
             // FIXME: Move this logic out of the rendering code (rdar://172280267)
-            
-            // The "default" information provided by the Info.plist is the base information because it applies to every thing in the module.
-            var information = baseAvailabilityByPlatform
+            var information = inSourceAvailability.availability.isEmpty ? baseAvailabilityByPlatform : [String: AvailabilityRenderItem]()
             
             var unavailablePlatformNamesToRemove = [String]()
             
--- a/Tests/SwiftDocCTests/Model/AvailabilityTests.swift
+++ b/Tests/SwiftDocCTests/Model/AvailabilityTests.swift
@@ -170,6 +170,48 @@ struct AvailabilityTests {
         #expect(availability.first(where: { $0.domain?.rawValue == "iPadOS"      })?.introducedVersion?.description ==  "6.0.0")
         #expect(availability.first(where: { $0.domain?.rawValue == "tvOS"        })?.introducedVersion?.description == "10.0.0")
     }
+
+    // Regression test for rdar://174841126
+    // Symbols should not use the module-level default availability (in Info.plist) if in-source availability is present.
+    @Test
+    func symbolWithInSourceAvailabilityDoesNotUseModuleDefaultAvailability() async throws {
+        let macOSOnlySymbol = makeSymbol(
+            id: "macos-only-symbol",
+            kind: .class,
+            pathComponents: ["MacOSOnlyClass"],
+            availability: [makeAvailabilityItem(domainName: "macOS", introduced: SymbolGraph.SemanticVersion(string: "14.0"))]
+        )
+
+        let catalog = Folder(name: "unit-test.docc") {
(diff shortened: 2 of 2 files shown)


snprajwal:
@swift-ci please test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1295,10 +1295,10 @@ public struct RenderNodeTranslator: SemanticVisitor {
         
         node.metadata.platformsVariants = VariantCollection<[AvailabilityRenderItem]?>(from: symbol.availabilityVariants) { _, inSourceAvailability in
             // Different sources of availability information are added in-order to compute the complete availability information.
+
+            // The default availability is inherited only if there is no in-source availability coming from the symbol graphs.
```
d-ronnqvist: FYI: This comment is only true because `SymbolGraphLoader.addDefaultAvailability(to:moduleName:)` has already integrated the default availability in the in-source availability as necessary. In general it is not true that pages don't consider default availability when the symbol has any in-source availability information.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -97,6 +97,49 @@ class SymbolAvailabilityTests: XCTestCase {
...
+            JSONFile(name: "ModuleName-watchos.symbols.json", content: makeSymbolGraph(
+                moduleName: "ModuleName",
+                platform: .init(operatingSystem: .init(name: "watchos"))
+            )),
+        ])
+
+        let (_, context) = try await loadBundle(catalog: catalog)
+
+        let reference = try XCTUnwrap(context.soleRootModuleReference).appendingPath("MacOSOnlyClass")
+        let node = try context.entity(with: ResolvedTopicReference(bundleID: context.inputs.id, path: reference.path, sourceLanguage: .swift))
```
d-ronnqvist: FYI: you can also access a symbol by its USR:
```swift
let node = try XCTUnwrap(context.documentationCache["macos-only-symbol"])
```

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -97,6 +97,49 @@ class SymbolAvailabilityTests: XCTestCase {
         ])
     }
     
+    // Regression test for rdar://174841126
+    // Symbols should not inherit default availability if in-source availability is present.
+    func testSymbolWithInSourceAvailabilityDoesNotInheritDefaultAvailability() async throws {
```
d-ronnqvist: Would it make sense to add this test in AvailabilityTests.swift instead? That's where I'm trying to accumulate all availability-logic behaviors (see https://github.com/swiftlang/swift-docc/pull/1497). Those tests are also already updated to Swift Testing, so this test could use slightly more modern patterns there.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -97,6 +97,49 @@ class SymbolAvailabilityTests: XCTestCase {
...
+        let macOSOnlySymbol = makeSymbol(
+            id: "macos-only-symbol",
+            kind: .class,
+            pathComponents: ["MacOSOnlyClass"],
+            availability: [makeAvailabilityItem(domainName: "macOS", introduced: SymbolGraph.SemanticVersion(string: "14.0"))]
+        )
+
+        let catalog = Folder(name: "unit-test.docc", content: [
```
d-ronnqvist: FYI: there's a (recently added) result builder that allows you to omit the commas between files and allows you to use conditionals and loops to create the files.

For example, you could create the two empty symbol graph files like:
```swift
for platformName in ["ios", "watchos"] {
    JSONFile(name: "ModuleName-\(platformName).symbols.json", content: makeSymbolGraph(
        moduleName: "ModuleName", 
        platform: .init(operatingSystem: .init(name: platformName))
    ))
}
```

or all three symbol graph files like:

```swift
for (platformName, symbols) in [
    ("macos",   [macOSOnlySymbol]), 
    ("ios",     []), 
    ("watchos", [])
] {
    JSONFile(name: "ModuleName-\(platformName).symbols.json", content: makeSymbolGraph(
        moduleName: "ModuleName", 
        platform: .init(operatingSystem: .init(name: platformName)),
        symbols: symbols
    ))
}
```

d-ronnqvist (approved):

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
snprajwal: That's an important distinction, thanks for pointing it out. I've updated the comment to include this detail.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
snprajwal: Nice one, thanks!

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
snprajwal: Makes sense, moved it

snprajwal:
@swift-ci please test

PR #1508, opened 2026-04-16
New `@Card` directive

mportiz08:
Bug/issue #, if applicable: 173272119

## Summary

This PR adds support for a new `@Card` block directive, which groups and highlights blocks of content
to be rendered within a single "card" block.

This opt-in feature can be enabled with the `--enable-experimental-card-directive` feature flag for convert/preview commands.

See the associated Swift Forums pitch/discussion [here][1].

### Examples

```markdown
@Card {
	Hello

	World
}
```

example-01

```markdown
@Card {
	### This is a heading

	More heading content

	---

	This content is the main body of the card.

	Here is some more body content.
}
```

example-02

## Dependencies

https://github.com/swiftlang/swift-docc-render/pull/1004

## Testing

Steps:
1. Verify unit tests pass with `bin/test`.
2. Checkout the associated DocC-Render [branch][2] and build it with the following commands: `npm ci; npm run build`
3. Manually compile DocC catalogs using the new directive as described above with the following command: `DOCC_HTML_DIR=/path/to/swift-docc-render/dist swift run preview --enable-experimental-card-directive /path/to/ExampleCatalog.docc`
4. Verify that the card is only rendered as expected when using the `--enable-experimental-card-directive` flag.

## Checklist

- [X] Added tests

[1]: https://forums.swift.org/t/card-directive/86076
[2]: https://github.com/swiftlang/swift-docc-render/pull/1004


Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
```diff
@@ -484,6 +484,13 @@ extension Docc {
                 """)
             )
             var enableParametersAndReturnsValidation = true
+
+            /// A user-provided value that is true if the user enables support for the `@Card` directive.
+            @Flag(
+                name: .customLong("enable-card-directive"),
+                help: "Enable support for the @Card directive."
+            )
+            var enableCardDirective = false
```
d-ronnqvist: As discussed in the Documentation Workgroup yesterday; we generally introduce features as _experimental_. If you look elsewhere in this file you'll find a handful of examples of how we name such flags and their variable.

If you look in ConvertSubcommandTests.swift you'll find a handful of examples of how we test such flags.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -835,6 +838,17 @@ public enum RenderBlockContent: Equatable {
             public var content: [RenderBlockContent]
         }
     }
+
+    /// A card element with separate head and content sections.
+    ///
+    /// The head section typically contains a heading and its first associated paragraph,
+    /// while the content section contains the remaining block elements.
+    public struct Card: Codable, Equatable {
```
d-ronnqvist, line 846: minor: This type doesn't have a public memberwise initializer. Without it, clients of DocC can't create Card values.

Sources/SwiftDocC/Utility/FeatureFlags.swift
```diff
@@ -36,6 +36,9 @@ public struct FeatureFlags: Codable {
     
     /// Whether or not support for validating parameters and return value documentation is enabled.
     public var isParametersAndReturnsValidationEnabled = true
+
+    /// Whether or not support for the `@Card` directive is enabled.
+    public var isCardDirectiveEnabled = false
```
d-ronnqvist: We generally introduce new features as _experimental_. There are other examples in this file that you can look at to see how we name such flags.

Tests/SwiftDocCTests/Testing+ParseDirective.swift
```diff
@@ -70,6 +70,17 @@ func parseDirective<Directive: RenderableDirectiveConvertible>(
     return (renderedContent, problems, directive)
 }
 
+func parseDirective<Directive: RenderableDirectiveConvertible>(
```
d-ronnqvist: Question: why did you add a new test helper for this? How is it different from the existing test helpers?

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
```diff
@@ -0,0 +1,213 @@
...
+ Copyright (c) 2026 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import Testing
+@testable import SwiftDocC
```
d-ronnqvist: Question: is there a particular internal API that this test needs to access? If there isn't; can we remove the `@testable` keyword and use a regular (public access level) import?

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
```diff
@@ -0,0 +1,213 @@
...
+struct CardTests {
+    /// A configuration with the card directive feature flag enabled.
+    static var cardEnabledConfiguration: DocumentationContext.Configuration {
+        var configuration = DocumentationContext.Configuration()
+        configuration.featureFlags.isCardDirectiveEnabled = true
+        return configuration
+    }
+
+    @Test
+    func noContent() async throws {
```
d-ronnqvist: Please see our guidelines about testing in CONTRIBUTING.md. Specifically, it would help to rename these tests so that they each describe the behavior that they're verifying:

>- Use a descriptive and readable phrase as the test name.
>
>   It can be easier to understand a test's implementation if its name describes the _behavior_ that the test verifies. A phrase that start with a verb can often help make a test's name a more readable description of what it's verifying. For example: `sortsSwiftFirstAndThenByID`, `raisesDiagnosticAboutCyclicCuration`, `isDisabledByDefault`, and `considersCurationInUncuratedAPICollection`.

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+            (contentCompiler.visit(block) as? [RenderBlockContent]) ?? []
+        }
+    }
+
+    private func partition<Separator: Markup>(
+        elements: [any Markup],
+        separatedBy separator: Separator.Type
+    ) -> (before: [any Markup], after: [any Markup]) {
+        var beforeSeparator = true
+        return elements.reduce((before: [], after: [])) { partitioned, element in
```
d-ronnqvist: minor: This implementation is accidentally _O(N²)_ due to repeatedly copying the accumulated elements for every iteration. Those copies could be avoided by using [`reduce(into:_:)`](https://developer.apple.com/documentation/swift/array/reduce(into:_:)) instead. 

That said, it's still unnecessary to make _any_ copies of the elements for a partitioning operation. If you look at Swift's [`partition(by:)`](https://developer.apple.com/documentation/swift/array/partition(by:)-90po8) function, it instead reorders the elements so that the elements that match the predicate are in one partition (one side of the pivot index) and the elements that don't match the predicate are in the other partition. That allows the caller to iterate over the two partition slices without making separate copies.

However, in this case it's not necessary to do a full partition operation because once we've found the first thematic break we know all elements that belong in each partition. This mean that this operation can more simply and efficiently be described as first finding that thematic break (for example using [`firstIndex(where:)`](https://developer.apple.com/documentation/swift/int/words-swift.struct/firstindex(where:))) and then rendering the two slices (one that goes up to the thematic break and the other that starts after the thematic break)

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+    init(originalMarkup: BlockDirective) {
+        self.originalMarkup = originalMarkup
+    }
+}
+
+extension Card: RenderableDirectiveConvertible {
+    func render(with contentCompiler: inout RenderContentCompiler) -> [any RenderContent] {
+        // Gate the Card directive behind its feature flag.
+        guard contentCompiler.context.configuration.featureFlags.isCardDirectiveEnabled else {
+            return render(blocks: content.elements, with: &contentCompiler)
```
d-ronnqvist: Question: under what scenario would we end up rendering a card if the feature flag wasn't enabled?

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+public import Markdown
+
+/// A directive that renders its content as a card with separate head and content sections.
```
d-ronnqvist: It might be good to mention _somewhere_ in the user-facing documentation that it's necessary to pass a  command line feature flag to use this directive. Otherwise, someone who finds the documentation and copies the example might have a bad experience when it doesn't work for them.

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+///
+/// The `@Card` directive partitions its child markup into two sections:
+/// - **head**: The first heading and the first paragraph following it.
+/// - **content**: All remaining block elements.
```
d-ronnqvist: While this documentation is correct for the example below, it's not accurate in the general case. As it's currently written I suspect that some readers might think that the first heading and first paragraph always goes in the "head" section but that's not true in the general case. Instead, it would be more accurate to say that any content before the first thematic break goes into the "head" section.

Also, I'm wondering if these names ("head" and "content") are relevant to the reader or if it's sufficient to say that they're two different sections.

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+        }
+    }
+
+    private func partition<Separator: Markup>(
+        elements: [any Markup],
+        separatedBy separator: Separator.Type
+    ) -> (before: [any Markup], after: [any Markup]) {
+        var beforeSeparator = true
+        return elements.reduce((before: [], after: [])) { partitioned, element in
+            if beforeSeparator && type(of: element) == separator.self {
```
d-ronnqvist: minor: This function is only called exactly once so it doesn't benefit from having a `separator` parameter. Instead, the separator parameter makes the implementation more abstract by requiring a `type(of:)` call instead of a more readable `element is ThematicBreak` check.

Sources/SwiftDocC/Semantics/Reference/Card.swift
```diff
@@ -0,0 +1,125 @@
...
+///
+///     ---
+///
+///     Some body content.
+///
+///     Another body paragraph.
+/// }
+/// ```
+public final class Card: Semantic, AutomaticDirectiveConvertible, MarkupContaining {
+    public static let introducedVersion = "6.4"
```
d-ronnqvist, line 48: I'll mention for broader visibility that we talked about what release this feature can land in, yesterday, in the Documentation Workgroup. While the 6.4 release _is_ in convergence and primarily accepts bug fixes, the group didn't see any concerns with merging this feature as "experimental" behind an off-by-default feature flag in the 6.4 release.

Sources/SwiftDocC/Semantics/Reference/Card.swift
d-ronnqvist, line 1: This file needs to be listed in the CMakeLists files for the SwiftDocC target. See CONTRIBUTING.md for more information.

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
```diff
@@ -0,0 +1,213 @@
...
+            @Card {
+                ### Example heading
+
+                Some content.
+            }
+            """
+        }
+
+        #expect(card != nil)
+        #expect(problems == [])
```
d-ronnqvist: minor: If another code change in the future causes this test assertion to fail; it would be more helpful is it included some readable version of what went wrong if it. For more information see the testing section in CONTRIBUTING.md:

> - Think about what information would be helpful to someone else who might debug that test case if it fails in the future.
>  
>    In an open source project like Swift-DocC, it's possible that a person you've never met will continue to work on code that you wrote.
  It could be that they're working on the same feature as you, or it could also be that they're working on something entirely different but their changes broke a test that you wrote.
  To help make their experience better, we appreciate any time that you spend considering if there's any information that you would have wanted to tell that person, as if they were a colleague.
>  
>    One way to convey this information could be to verify assumptions (like "this test content has no user-facing warnings") using `#expect`.
  Additionally, if there's any information that you can surface right in the test failure that will save the next developer from needing to add a breakpoint and run the test again to inspect the value,
  that's a nice small little thing that you can do for the developer coming after you:
>    ```swift
>    #expect(problems.isEmpty, "Unexpected problems: \(problems.map(\.diagnostic.summary))")
>    ```

d-ronnqvist (changes requested):
Thanks for opening this PR. The code so far is a good start but there's also some things that are missing.

Like we talked about in the Documentation Workgroup yesterday; we generally expect different experimental features to work well together. Two such existing features that are particularly relevant for this feature are:
- the in-page HTML output (`--experimental-transform-for-static-hosting-with-content`)
- the supplementary Markdown output (`--enable-experimental-markdown-output`)


In addition to adding tests that the `Card` directive works with those features, there are a lot of important behaviors that aren't covered by tests:
- There are no tests that verify that a `Card` directive in an article, in-source documentation comment, and documentation extension file is included in the RenderNode output for that page when the feature flag is enabled.
  - There are no tests that verify that DocC warns when a `Card` directive is used without the feature flag.
- There are no tests that verify which Tutorial directives (if any) support the `Card` directive. 
- There are no tests that verify that links or `Snippet` directives in the `Card` directive's content resolve correctly.
  - There are no tests that verify that DocC warns about unresolved links (or unresolved `Snippet` paths) encountered in the `Card` directive's content.
- There are no tests that verify that DocC warns about unexpected content (for example `Metadata` directive or own of its child directives) in the `Card` directive's content.

d-ronnqvist:
Note that the "Testing" section of the pull request template is intended for instructions for how a human can verify the changes end-to-end. This is in addition to the expectation that all changes are covered by automated tests. I couldn't get this feature to work but I'm assuming that that's because it depends on this [Swift-DocC Render counterpart](https://github.com/swiftlang/swift-docc-render/pull/1004). If that's correct, that would be something worth listing in the (removed) "Dependencies" section from pull request template.

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
d-ronnqvist: Actually, I would expect DocC to raise some kind of warning if the developer used the Card directive without passing the feature flag. For guidance on adding new user-facing diagnostics, see Adding-Diagnostics.md

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: My thought here was that it would be preferable to still render the contents inside of the card without rendering the actual card itself when the feature flag is not enabled (as opposed to not rendering anything).

It's not a strongly held opinion though—if you think it's better to omit the whole card contents if the feature flag is not enabled, I can change the behavior here. I wasn't aware of a similar pattern that might have been used for a similar scenario that I should try to follow.

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Do you mean in the user facing DocC catalog? I thought we tended to omit most experimental features from these docs, but I can definitely add more docs there if we want.

Tests/SwiftDocCTests/Testing+ParseDirective.swift
mportiz08: The difference is the `configuration` parameter, which I used to test configurations with and without the feature flag enabled.

I'm open to alternative implementations if you can think of an improved way of doing this without the added helper.

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
mportiz08: These tests aren't directly calling any internal API, so you're right that I can take it out (will do).

Test files generated with Xcode or Swift Package Manager usually include the `@testable` keyword when importing the relevant module in the stub template content, so it's mostly just a pattern I've followed even when it isn't strictly necessary.

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
mportiz08: I see a mix of naming conventions, but I'll add the word "experimental" so the flag will change to `--enable-experimental-card-directive` instead of `--enable-card-directive` (along with adding the unit tests that specifically test this flag).

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: I didn't use the builtin `partition(by:)` since it would still include the first thematic break element, which I explicitly do not want to consider when rendering. Also, this function is just for extracting the 2 distinct groups (without actually rendering them yet). I don't expect performance to ever _reasonably_ be an issue with the amount of content practically contained here, but I agree it would be better to still have this written in a more performant manner regardless.

I'll likely rewrite this to not use these helper functions. I think finding the first break and rendering things inline within the main render function should most likely address most of the feedback here.

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Yes sorry. These docs are out-of-date—when I first approached the syntax, I was explicitly looking for the first header and first paragraph to split the sections. I later changed that to use the thematic break element but forgot to update these comments—I'll fix that.

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
mportiz08: Addressed with a288489e77f263128e5bf0fb802d4f44588dbac8

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
mportiz08: Addressed with 836e3b9fb559a78ab9d0f0a94c350c108adbdb2a

Sources/SwiftDocC/Utility/FeatureFlags.swift
mportiz08: Addressed with a288489e77f263128e5bf0fb802d4f44588dbac8

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
mportiz08: Addressed with 6efadd6ccabd9d2cff807eba35a575499cf8c82f

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
mportiz08: Addressed with 7d27ab9f2996348bae86afdf7b4d52989942a3e6

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Addressed with 8876430ce1f935ce91892031a9918ab1a69bba13

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Addressed with ce3455b11f0ec5bb32a1a4b46b2ae1cce386371a

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Addressed with 8876430ce1f935ce91892031a9918ab1a69bba13

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Addressed with fe12dca6e4c01495f1eb1e66dbb7953983cd59d6

Tests/SwiftDocCTests/Semantics/Reference/CardTests.swift
mportiz08: Addressed with:
* ad0e989e85b4bff57bba23c47bfc42aec97c6d21
* 92fc54abeb5f7f2fe756918d3280730d34f73009

mportiz08:
> Note that the "Testing" section of the pull request template is intended for instructions for how a human can verify the changes end-to-end. This is in addition to the expectation that all changes are covered by automated tests. I couldn't get this feature to work but I'm assuming that that's because it depends on this [Swift-DocC Render counterpart](https://github.com/swiftlang/swift-docc-render/pull/1004). If that's correct, that would be something worth listing in the (removed) "Dependencies" section from pull request template.

I added the "Dependencies" section back and updated the testing instructions with more explicit details on how to manually test content using this branch and the associated swift-docc-render branch.

mportiz08:
Still working on addressing the two higher level areas mentioned:
* integration with other experimental features
* adding more tests

I'll mark this as needing re-review once that's done.

I've addressed a number of the other specific feedback if you want to check those changes out in the meantime.

Tests/SwiftDocCTests/Testing+ParseDirective.swift
d-ronnqvist: I see. Can you check if we can add add a new configuration parameter—with a default value—to one of the existing test helpers instead? That way we don't need to repeat the entire test helper just to add an optional parameter.

Sources/SwiftDocC/Semantics/Reference/Card.swift
d-ronnqvist: Ah, I see the confusion here. This documentation would be included in DocC's user-facing documentation once this PR is merged. If we don't want that documentation to be displayed yet, then the `Card` type should override `hiddenFromDocumentation` and return `true`.

Sources/SwiftDocC/Semantics/Reference/Card.swift
d-ronnqvist: Sorry, I should have been more clear in my question. 

What I meant to ask is; does this code, that renders a card, ever run if the feature flag wasn't enabled when DocC parsed the content that contained the card markup? 

When the feature flag isn't passed I would expect DocC to behave the same as before this PR; meaning that I would expect DocC to warn that "Card" is not a recognized/supported directive and then skip past it. In that case, I would imagine that it'd be a programming error to ever render a card without the feature flag; either because of a bug that unexpectedly parsed the card markup or because of a bug that didn't propagate the feature flags to the necessary places.

Tests/SwiftDocCTests/Testing+ParseDirective.swift
mportiz08: Addressed with a58e3bc6c80b69c0b5576dc0fecc62832d6dbbdf

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Hm. I might need to move the logic for the feature flag if we want it to work that way. Is there a place you would suggest to move the feature flag logic so it behaves the way you're expecting it to?

Right now, it works as follows when compiling content using the directive without the flag enabled:
1. [Warns][1] at instances of directive in markdown that it can only be used with feature flag
2. Transparently [renders][2] content inside of card (but not the card itself)

To make it behave the way you expect, would I just need to update the `validate` method to return false (in addition to the warning) with the flag not enabled? Or is there an even better strategy for using the flag so things don't even get to the validate/render calls here?

I'm open to whatever approach best fits with existing DocC patterns, but I need some guidance on implementing it the right way.

[1]: https://github.com/mportiz08/swift-docc/blob/a58e3bc6c80b69c0b5576dc0fecc62832d6dbbdf/Sources/SwiftDocC/Semantics/Reference/Card.swift#L75
[2]: https://github.com/mportiz08/swift-docc/blob/a58e3bc6c80b69c0b5576dc0fecc62832d6dbbdf/Sources/SwiftDocC/Semantics/Reference/Card.swift#L103

Sources/SwiftDocC/Semantics/Reference/Card.swift
mportiz08: Addressed with 0f7d197b6a2a0597fe1f5b29c0bc88e23a58f1cd

mportiz08:
> Like we talked about in the Documentation Workgroup yesterday; we generally expect different experimental features to work well together. Two such existing features that are particularly relevant for this feature are:
> 
> * the in-page HTML output (`--experimental-transform-for-static-hosting-with-content`)
> * the supplementary Markdown output (`--enable-experimental-markdown-output`)

Addressed with:
* 39416cf0a1abd9a2f88f7bbfa08d036efe11ee9c
* 41431c4da06f6bcbacc942505201b85e8bc6d103

mportiz08:
>   * There are no tests that verify that DocC warns about unresolved links (or unresolved `Snippet` paths) encountered in the `Card` directive's content.

Addressed with 673e4eace528e1acea1d961a0cc4ab8118cdc7a0

mportiz08:
Going to close this PR for now and possibly revisit it after there's been more of a consensus on the right syntax for this based on the forum discussion.

PR #1509, opened 2026-04-22
Deprecate `Problem` type in favor of using `Diagnostic` as the only diagnostic type in DocC.

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This aims to clean up the DocC API that's used to represent diagnostics by:
- Deprecating `Problem` (which _was_ wrapping a `Diagnostic` even though that sounds a bit backwards)
- Moving the problem's `possibleSolutions` to the diagnostic; so that it holds all the information about the issue including its notes and solutions
- Renaming `DiagnosticNote` to a nested `Note` type inside of `Diagnostic`
- Moving `Replacement` to a nested type inside of `Solution`.

----

I made the initial deprecations by hand and after that updated most of the calling code through a series of regular expression replacements of frequently occurring update patterns. For example:

 pattern | replacement
----|----
`Problem\(diagnostic: (.*), possibleSolutions: \[\]\)` | `$1`
`Problem\(diagnostic: (.*)\), possibleSolutions: (.*)\)` | `$1, solutions: $2)`
`(#expect\|XCTAssert)\(problem\.diagnostic\.` | `$1(diagnostic.`
`func (\w+)Problem(\(.*\)) -> Problem` | `func $1Diagnostic$2 -> Diagnostic` 
`var problems(: \[Problem\] = \[\]\| = \[Problem\]\(\))` | `var diagnostics = [Diagnostics]()`
etc.

I also utilized regular expression replacements for some patterns of deprecating API. For example, replacing
```
public func analyze\(_ directive: BlockDirective, children: some Sequence\<any Markup\>, source: URL\?, for bundle: DocumentationBundle, problems: inout \[Problem\]\) -> (.*) {
```
with a deprecated wrapper that transforms the inout diagnostics into inout problems.

```swift
@available(*, deprecated, renamed: "analyze(_:children:source:for:diagnostics:)", message: "Use 'analyze(_:children:source:for:diagnostics:)' instead. This deprecated API will be removed after 6.5 is released.")
public func analyze(_ directive: BlockDirective, children: some Sequence<any Markup>, source: URL?, for bundle: DocumentationBundle, problems: inout [Problem]) -> $1 {
    var diagnostics = [Diagnostic]()
    defer {
        problems.append(contentsOf: diagnostics.map { .init(diagnostic: \$0) })
    }
    return analyze(directive, children: children, source: source, for: bundle, diagnostics: &diagnostics)
}

public func analyze(_ directive: BlockDirective, children: some Sequence<any Markup>, source: URL?, for bundle: DocumentationBundle, diagnostics: inout [Diagnostic]) -> $1 {
```

For anything that I couldn't quickly replace with a pattern (or I couldn't spot a sufficiently frequently recurring pattern) I simply made the changes by hand.

Lastly, I tried to clean up documentation and code comments by searching for `// .*problem\n.*diagnostic` and updating the documentation or code comments by hand.

I've inspected all of the source changes and most of the test changes. In some cases I've tried to clean up some of the tests code but many of those tests would benefit from a larger cleanup and an update to use Swift Testing so I eventually stopped doing that and left the later test code as-is (apart from renaming variables and updating it to not call deprecated API).

## Dependencies

None.

## Testing

## Checklist

- [x] ~Added~ Updated tests


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Diagnostics/Solution.swift
```diff
@@ -1,27 +1,27 @@
...
  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
-/**
- A solution to a `Problem`.
- */
+/// A possible solution to a problem or issue that's reported about in a diagnostic.
 public struct Solution: Hashable {
-    /// A *brief* description of what the solution is.
+    /// A brief summary that describe what the solution is to the end-user.
```
snprajwal:
```suggestion
    /// A brief summary that describes what the solution is to the end-user.
```

Sources/SwiftDocC/Checker/Checker.swift
```diff
@@ -17,16 +17,33 @@ public import Markdown
...
+    }
+}
+
+// Default implementation so that existing conformances  by the additional initializer parameter in the protocol requirements.
+public extension Checker {
+    @available(*, deprecated) // This needs to be marked deprecated because it calls deprecated API.
+    var diagnostics: [Diagnostic] {
+        problems.map(\.diagnostic)
+    }
+}
```
snprajwal: I'm not sure I fully understand why we're doing this. It looks like this uses `problems` to build the diagnostics array, and `problems` is constructed from the `diagnostics` array that has been introduced above. Is there a reason we're doing this sort of indirection?

snprajwal (commented):
Mostly all NFC, LGTM with one question below

Sources/SwiftDocC/Checker/Checker.swift
d-ronnqvist: The issue is that `Checker` is a `public` protocol so adding a new protocol requirement or removing an existing protocol requirement are both breaking changes. 

----

Step by step; I would describe the changes like this:

Adding the new `diagnostics` requirement is a breaking change because existing conformances wouldn't have that property. 
```diff 
  public protocol Checker: MarkupWalker {
      var problems: [Problem] { get }
+     var diagnostics: [Diagnostic] { get }
  }
```

To not break existing implementations, we can add a default implementation of the new `diagnostics` property:
```diff 
  public protocol Checker: MarkupWalker {
      var problems: [Problem] { get }
      var diagnostics: [Diagnostic] { get }
  }
 
+ public extension Checker {
+     var diagnostics: [Diagnostic] {
+         problems.map(\.diagnostic)
+     }
+ } 
```

If we now deprecate the old `problems` requirement, that will signal to all existing implementation that they need to implement `diagnostics` instead of `problems`. 

```diff 
  public protocol Checker: MarkupWalker {
+     @available(*, deprecated, renamed: "diagnostics", message: "Use 'diagnostics' instead. This deprecated API will be removed after 6.5 is released.")
      var problems: [Problem] { get }
      var diagnostics: [Diagnostic] { get }
  }
 
  public extension Checker {
      var diagnostics: [Diagnostic] {
          problems.map(\.diagnostic)
      }
  } 
```

Because the default implementation is now calling a deprecated property, it also needs to be deprecated. This has the added benefit that existing `Checker` implementation will get a "Deprecated default implementation is used to satisfy property 'diagnostics' required by protocol 'Checker'" warning.

```diff 
  public protocol Checker: MarkupWalker {
      @available(*, deprecated, renamed: "diagnostics", message: "Use 'diagnostics' instead. This deprecated API will be removed after 6.5 is released.")
      var problems: [Problem] { get }
      var diagnostics: [Diagnostic] { get }
  }
 
  public extension Checker {
+     @available(*, deprecated) // This needs to be marked deprecated because it calls deprecated API.
      var diagnostics: [Diagnostic] {
          problems.map(\.diagnostic)
      }
  } 
```

Now, the protocol still _requires_ the deprecated `problems` property. This means that both existing and new types need to implement both `problems` and `diagnostics`. If we remove the `problems` requirement from `Checker` that solves the issue for implementing types _but_ it's a breaking change for any existing calling code that accesses the `problems` property:

```diff 
  public protocol Checker: MarkupWalker {
-     @available(*, deprecated, renamed: "diagnostics", message: "Use 'diagnostics' instead. This deprecated API will be removed after 6.5 is released.")
-     var problems: [Problem] { get }
      var diagnostics: [Diagnostic] { get }
  }
 
  public extension Checker {
      @available(*, deprecated) // This needs to be marked deprecated because it calls deprecated API.
      var diagnostics: [Diagnostic] {
          problems.map(\.diagnostic)
      }
  } 
```

To solve that breaking change, we can add back the `problems` property as an extension

```diff 
  public protocol Checker: MarkupWalker {
      var diagnostics: [Diagnostic] { get }
  }
 
+ public extension Checker {
+     @available(*, deprecated, renamed: "diagnostics", message: "Use 'diagnostics' instead. This deprecated API will be removed after 6.5 is released.")
+     var problems: [Problem] {
+         diagnostics.map { Problem(diagnostic: $0) }
+     }
+ } 
  
  public extension Checker {
      @available(*, deprecated) // This needs to be marked deprecated because it calls deprecated API.
      var diagnostics: [Diagnostic] {
          problems.map(\.diagnostic)
      }
  } 
```

After this step, there are no breaking changes and:
- Existing `Checker` implementations get the "Deprecated default implementation is used to satisfy property 'diagnostics' required by protocol 'Checker'" warning that informs them of the deprecation and that they should implement `diagnostics` instead.
- Existing calling code gets a deprecation warning from accessing the `problems` property that informs them that they should call `diagnostics` instead.

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Checker/Checker.swift
d-ronnqvist: You'll see this same pattern whenever we need to modify a public protocol in SwiftDocC.

snprajwal: That makes sense now, thanks for the explanation! 😄

snprajwal (approved):

PR #1510, opened 2026-04-23
Show Deprecated label in navigator for items that use @DeprecationSummary

franklinsch:
Bug/issue #, if applicable: rdar://170218646

## Summary

The navigator sidebar currently doesn't show the "Deprecated" label for symbols or articles that have `@DeprecationSummary` directive but are not deprecated on any platform via availability annotations. This is because the navigator index derived deprecation solely from per-platform availability data in the availability index, ignoring the render node's deprecation summary.

## Dependencies

N/A

## Testing

Run DocC on a symbol that doesn't specify deprecation annotations in code, but uses `@DeprecationSummary` in a documentation extension file. The navigator item for that symbol should show the deprecated label.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
@@ -76,6 +76,9 @@ public final class NavigatorItem: Serializable, Codable, Equatable, CustomString
         - availabilityID:  The identifier of the availability information of the page.
         - path: The path to load the content.
         - icon: A reference to a custom image for this navigator item.
+        - isExternal: A flag indicating whether the navigator item belongs to an external documentation archive.
+        - isBeta: A flag indicating whether the navigator item is in beta.
+        - isDeprecated: A flag indicating whether the navigator item is deprecated.
      */
     init(
         pageType: UInt8,
@@ -113,6 +116,7 @@ public final class NavigatorItem: Serializable, Codable, Equatable, CustomString
         - icon: A reference to a custom image for this navigator item.
         - isExternal: A flag indicating whether the navigator item belongs to an external documentation archive.
         - isBeta: A flag indicating whether the navigator item is in beta.
+        - isDeprecated: A flag indicating whether the navigator item is deprecated.
      */
     public init(
         pageType: UInt8,
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
@@ -59,6 +59,12 @@ public final class NavigatorItem: Serializable, Codable, Equatable, CustomString
     /// Used for determining whether stray navigation items should remain part of the final navigator.
     var isExternal: Bool = false
     
+    /// A value that indicates whether this item has been deprecated.
+    ///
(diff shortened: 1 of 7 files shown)


franklinsch:
@swift-ci please test

Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
```diff
@@ -71,7 +77,18 @@ public final class NavigatorItem: Serializable, Codable, Equatable, CustomString
...
+        pageType: UInt8,
+        languageID: UInt8,
+        title: String,
+        platformMask: UInt64,
+        availabilityID: UInt64,
+        path: String,
+        icon: RenderReferenceIdentifier? = nil,
+        isExternal: Bool = false,
+        isBeta: Bool = false,
+        isDeprecated: Bool = false
```
snprajwal, line 93: The doc comment is missing the new field (and a couple of old ones), could you please add them?

Sources/SwiftDocC/Indexing/Navigator/NavigatorItem.swift
```diff
@@ -96,7 +114,17 @@ public final class NavigatorItem: Serializable, Codable, Equatable, CustomString
...
+    public init(
+        pageType: UInt8,
+        languageID: UInt8,
+        title: String,
+        platformMask: UInt64,
+        availabilityID: UInt64,
+        icon: RenderReferenceIdentifier? = nil,
+        isExternal: Bool = false,
+        isBeta: Bool = false,
+        isDeprecated: Bool = false
```
snprajwal, line 130: The doc comment is missing the new field, could you please add it?

Sources/SwiftDocC/Indexing/Navigator/RenderNode+NavigatorIndex.swift
```diff
@@ -133,6 +145,12 @@ extension NavigatorIndexableRenderMetadataRepresentation {
     }
 }
 
+extension NavigatorIndexableRenderNodeRepresentation {
+    var isDeprecated: Bool {
+        metadata.platforms?.contains { $0.deprecated != nil } == true
+    }
```
snprajwal: Nit: we're repeating this three times, it _might_ be good to extract it into a helper in the protocol:

```suggestion
    var isPlatformDeprecated: Bool {
        metadata.platforms?.contains { $0.deprecated != nil } == true
    }
    var isDeprecated: Bool {
        isPlatformDeprecated
    }
```
With this, the `||` checks above can also be reduced to `|| isPlatformDeprecated`. Not a big deal though, I'm fine with the current pattern too.

snprajwal (approved):
LGTM with some nits, thanks for fixing this!

Sources/SwiftDocC/Indexing/Navigator/RenderNode+NavigatorIndex.swift
franklinsch: Yes, good point. I initially didn't do that because I thought I'd need to add it to be protocol's API surface, but actually, as you point out, we can include it in a protocol extension instead.

franklinsch:
@swift-ci please test

d-ronnqvist (approved):

PR #1512, opened 2026-04-28
Filter by available traits in automatic curation

snprajwal:
Bug/issue #, if applicable: rdar://161926175

## Summary

When manually curating the topics section for a page, the nodes are filtered such that only the nodes in the current language are curated. The exception to this is if a node is in a language that is not otherwise available in the documentation, in which case it is curated in order to avoid dropping it entirely. However, automatic curation does not perform this filtering, meaning that nodes from a different language are incorrectly curated in the topics section. The navigator _does_ perform this filtering, leading to inconsistency between the topics section and navigator for such pages. This change consolidates the filtering logic into a common method and updates the automatic curation workflow to perform the filtering in the same way as manual curation currently does.

## Dependencies

N/A

## Testing

1. `swift run docc preview Tests/SwiftDocCTests/Test\ Bundles/MixedLanguageFrameworkWithArticlesUsingSupportedLanguages.docc`
2. Without this patch, the topics section in the root page incorrectly curates the ObjC-only article in the Swift page and the Swift article in the ObjC-only page, and doesn't match the navigator.
3. With this patch, the topics section in the root page correctly curates the pages that are available in the language, and matches the navigator.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -706,6 +706,8 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 sections.append(
                     contentsOf: renderAutomaticTaskGroupsSection(
                         article.automaticTaskGroups.filter { $0.renderPositionPreference == .top },
+                        allowedTraits: allowedTraits,
+                        availableTraits: documentationNode.availableVariantTraits,
                         contentCompiler: &topicSectionContentCompiler
                     )
                 )
@@ -746,6 +748,8 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 sections.append(
                     contentsOf: renderAutomaticTaskGroupsSection(
                         article.automaticTaskGroups.filter { $0.renderPositionPreference == .bottom },
+                        allowedTraits: allowedTraits,
+                        availableTraits: documentationNode.availableVariantTraits,
                         contentCompiler: &topicSectionContentCompiler
                     )
                 )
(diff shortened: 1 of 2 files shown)


snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test macos

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1018,15 +1022,59 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+        if availableSourceLanguageTraits.isDisjoint(with: referenceSourceLanguages) {
+            // The set of available source language traits has no members in common with
+            // the set of source languages the given reference is available in.
+            // Since there are no overlapping traits, this reference is made available
+            // in order to prevent dropping it entirely.
+            return true
+        }
+
+        return allowedTraits.contains { trait in
+            trait.sourceLanguage.map { referenceSourceLanguages.contains($0) } ?? false
```
sofiaromorales: This looks a bit condensed to me, loosing readability.

Can we make it more explicit? 

```
guard let language = trait.sourceLanguage else { return false }
return referenceSourceLanguages.contains(language)
```

Tests/SwiftDocCTests/Model/SemaToRenderNodeMultiLanguageTests.swift
```diff
@@ -878,6 +878,34 @@ class SemaToRenderNodeMixedLanguageTests: XCTestCase {
...
+            objcRenderNode.topicSections.first(where: { $0.title == "Articles" })
+        )
+        XCTAssertEqual(
+            objcArticlesSection.identifiers.sorted(),
+            [
+                "doc://org.swift.MixedLanguageFramework/documentation/MixedLanguageFramework/ArticleWithoutSupportedLanguages",
+                "doc://org.swift.MixedLanguageFramework/documentation/MixedLanguageFramework/ObjCArticle",
+                "doc://org.swift.MixedLanguageFramework/documentation/MixedLanguageFramework/SwiftAndObjCArticle",
+            ]
+        )
```
sofiaromorales, line 908: The logic ` guard !filteredReferences.isEmpty else { return nil }` makes sure we don't render an empty task group. This is a new check that's good to have.

We should add a test to make sure this logic works properly and we don't regress it. Please add a test that checks an automatic group that only has a Swift articles does not exists for the ObjC variant.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
sofiaromorales: Given that this method checks against not only allowed traits, but also any available trait, I propose renaming it to `isTopicAvailable`.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1018,15 +1022,59 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+        }
+    }
+
     /// Renders automatically generated task groups
-    private mutating func renderAutomaticTaskGroupsSection(_ taskGroups: [AutomaticTaskGroupSection], contentCompiler: inout RenderContentCompiler) -> [TaskGroupRenderSection] {
-        return taskGroups.map { group in
-            contentCompiler.collectedTopicReferences.append(contentsOf: group.references)
+    private mutating func renderAutomaticTaskGroupsSection(
```
sofiaromorales, line 1091: NIT: Update the DocC comment to document the filtering logic based on language traits.

snprajwal:
@swift-ci please test

sofiaromorales (approved):

Issue #1336, opened 2025-06-10
Documentation with a `--hosting-base-path` causes CSP errors


cuberoot:
This plugin adds a `<script>` tag to support alternative base paths for deployed documentation (using `--hosting-base-path`). That causes issues on servers that implement a [content security policy ](https://developer.mozilla.org/en-US/docs/Web/HTTP/Guides/CSP). Those typically restrict the use of plain script tags in source.

The preferred and more standard approach would be to use the [<base> tag instead](https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/base).

### Checklist
- [x] If possible, I've reproduced the issue using the `main` branch of this package.
- [x] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/swiftlang/swift-docc-plugin/issues).

### Expected behavior
I would expect the output to use the `<base>` tag when you use the `--hosting-base-path`

### Actual behavior
A `<script>` tag is introduced to inject the base path, which violates any secure content security policy

### Steps to Reproduce
Build the documentation for a package using a flag like `--hosting-base-path /documentation` and look at the generated `index.html`. It will contain code like:
```html
        <script>
            var baseUrl = "/documentation/"
        </script>
```

That would be better replaced by a `<base>` tag

### Swift-DocC Plugin Version Information

**Swift-DocC plugin version:** `1.4.3`
**Swift Compiler version:**`6.1.0`.

PR #1514, opened 2026-05-02
Replace inline script with <base> tag for CSP compliance (#1336)

VictorPuga:
# Replace inline script with <base> tag for CSP compliance

Bug/issue swiftlang/swift-docc-render#1007

## Summary

When building documentation with `--hosting-base-path`, DocC previously injected an inline `<script>var baseUrl = "{{BASE_PATH}}"</script>` element by token substitution in the HTML template. This violates Content Security Policy for servers that restrict inline scripts.

This PR replaces that mechanism with a standards-compliant `<base href="/path/" />` HTML element injected as the first child of `<head>`. The `<base>` element tells browsers how to resolve relative URLs without requiring JavaScript.

**Key changes:**

- Inject `<base href="/hosting/path/" />` immediately after the `<head>` opening tag when a hosting base path is provided
- Injection only occurs if a `<head>` tag is found in the template (graceful degradation for partial templates)
- The existing `{{BASE_PATH}}` token substitution continues to work for asset paths in external scripts and stylesheets
- Fully backward compatible — existing templates and deployments are unaffected

## Dependencies

None. This is a self-contained fix within the swift-docc repository.

### Automated Tests

Two new/updated tests verify the fix:

1. **`testIndexHTMLDataInjectsBaseTagWhenHeadPresent`** (new)
   - Tests that `<base href="/test/base-path/" />` is injected immediately after `<head>` when a hosting base path is provided
   - Verifies `{{BASE_PATH}}` token substitution still works for asset paths
   - Confirms no remaining `{{BASE_PATH}}` tokens in the output

2. **`testIncludesBasePathInPerPageIndexHTMLFile`** (updated)
   - Existing test updated to expect `<base href="/some/test/base-path/" />` in the generated HTML
   - Tests both per-page HTML mode (`includeHTMLContent = true/false`) with hosting base path

### Manual Verification

To verify the fix in a real documentation build:

```bash
# Build a test site with a hosting base path
swift run docc convert \
  /path/to/Package.docc \
  --output-path /tmp/docs-output/documentation-test \
  --hosting-base-path /documentation-test
```

Then inspect the generated HTML:

```bash
# Should show <base> tag as first element in <head>
head -20 /tmp/docs-output/documentation-test/*/index.html | grep -A 2 "<head>"
```

Expected output:

```html
<head>
  <base href="/documentation-test/" />
  <meta charset="utf-8" />
  ...
</head>
```

### CSP Compliance Testing

For servers with strict CSP headers that disallow inline scripts:

```bash
# Serve locally and test with CSP header
cd /tmp/docs-output
python3 -m http.server 8000

# Verify no CSP violations in browser console
curl -H "Content-Security-Policy: script-src 'self'" \
  http://localhost:8000/documentation-test/index.html
```

The documentation site should load and function correctly without CSP console errors related to inline scripts.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
+++ b/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
@@ -125,6 +125,15 @@ extension StaticHostableTransformer {
             }
 
             indexHTML = indexHTML.replacingOccurrences(of: HTMLTemplate.tag.rawValue, with: replacementString)
+
+            // Inject <base href="/path/"> immediately after <head> for CSP compliance.
+            // Per HTML spec, <base> must precede other URL-referencing elements like <link> and <script src>.
+            if let headTagRange = indexHTML.range(of: "<head", options: .caseInsensitive),
+               let headCloseAngle = indexHTML[headTagRange.upperBound...].firstIndex(of: ">") {
+                let insertionIndex = indexHTML.index(after: headCloseAngle)
+                indexHTML.insert(contentsOf: "<base href=\"\(replacementString)/\" />", at: insertionIndex)
+            }
+            // If no <head> is found (e.g., partial template), silently skip — graceful degradation.
         }
         
         return Data(indexHTML.utf8)
--- a/Tests/DocCCommandLineTests/StaticHostableTransformerTests.swift
+++ b/Tests/DocCCommandLineTests/StaticHostableTransformerTests.swift
@@ -191,5 +191,33 @@ class StaticHostableTransformerTests: StaticHostingBaseTests {
             XCTAssertEqual(testHTMLString, indexHTML, "Unexpected content in index.html at \(indexFileURL)")
         }
     }
+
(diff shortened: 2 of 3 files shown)


d-ronnqvist (changes requested):
AFAIK Swift-DocC Render depends on the base path being represented in that script tag. 

Regardless, the template script tag comes from the render template. If we want to change the way the base path is represented we should make that change in the template file that Swift-DocC Render creates rather than change _how_ it's represented in Swift-DocC. (IMO Swift-DocC should only be responsible for modifying the _value_, not the _way_ it's represented in the data.

VictorPuga:
Makes sense. I was a bit confused about why this was transferred to swift-docc (https://github.com/swiftlang/swift-docc/issues/1336#issuecomment-3499997994). I’ll try to provide a fix in swift-docc-render.

VictorPuga:
Closing in favor of https://github.com/swiftlang/swift-docc-render/pull/1008

PR #1505, opened 2026-04-15
Use `collection` navigator type for API collection

snprajwal:
Bug/issue #, if applicable: rdar://174485280

## Summary

41331819c fixed an issue in the navigator where API collections were being treated as symbols. It was remapped to the existing `article` render type. However, this diverges from the mapping used by swift-docc-render [1], where API collections are expected to use the `collection` render type in order to have the correct icon in the navigator. The render JSON uses this type, meaning that the icon is correctly present in the topics section, but the navigator now renders the article icon for API collections. This patch introduces the `collection` render type in the navigator.

[1]: https://github.com/swiftlang/swift-docc-render/blob/dc47cbc06e4e574ab9c75e0585fb708c9172051d/src/components/TopicTypeIcon.vue#L53

## Dependencies

N/A

## Testing

Preview a catalog containing an API collection, and view the navigator. Previously, the article icon would be rendered. With this patch, the API collection icon is rendered.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1516, opened 2026-05-08
[6.4] Use `collection` navigator type for API collection (#1505)

snprajwal:
- **Explanation**: This fixes an issue in the navigator where API collections were using the article icon. This was due to API collections previously being treated as articles as per the fix in 19c593357cc59762f7d1c2a52e1250d19e.
- **Scope**: This change does not modify existing public API, but does introduce a new variant in a public enum. This could lead to breakage if library consumers use the enum in an exhaustive switch without a default case. As the enum is not frozen, the library authors are allowed to introduce new fields. To make this more obvious to consumers, a non-frozen enum variant has been added to ensure a default case is always present.
- **GitHub Issue**: N/A
- **Risk**: N/A
- **Testing**: Unit tests have been added, and the test bundle in the repository has also been used to verify correct behaviour.
- **Reviewer**: @d-ronnqvist @sofiaromorales

### DIFF
--- a/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
+++ b/Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
@@ -251,6 +251,7 @@ public class NavigatorIndex {
         case overview = 5
         case resources = 6
         case symbol = 7 // This indicates a generic symbol
+        case collection = 8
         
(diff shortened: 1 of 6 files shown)


snprajwal:
@swift-ci please test

franklinsch (approved):

PR #1518, opened 2026-05-08
[6.4] Filter by available traits in automatic curation (#1512)

snprajwal:
- **Explanation**: When manually curating the topics section for a page, the nodes are filtered such that only the nodes in the current language are curated. The exception to this is if a node is in a language that is not otherwise available in the documentation, in which case it is curated in order to avoid dropping it entirely. Automatic curation does not perform this filtering,but the navigator _does_, leading to inconsistency between the topics section and navigator for such pages. This change updates the automatic curation workflow to perform the filtering in the same way as manual curation currently does.
- **Scope**: Impacts curation and UI, no API changes.
- **GitHub Issue**: N/A
- **Risk**: N/A
- **Testing**: Unit and integration tests, along with the test bundle in the repo
- **Reviewer**: @sofiaromorales


snprajwal:
@swift-ci please test

franklinsch (approved):

snprajwal:
@swift-ci please test

PR #1521, opened 2026-05-08
[6.4] Deprecate `Problem` type in favor of using `Diagnostic` as the only diagnostic type in DocC

d-ronnqvist:
- **Explanation**: This deprecates `Problem` type in favor of using `Diagnostic` as the only diagnostic type in DocC and updates API to use `Diagnostic` values instead of `Problem` values. 

  The reason for suggesting that we cherry pick this into the 6.4 release is that it will make it easier to cherry pick _other_ changes into the 6.4 release. If we don't cherry pick this, then new code on the main branch is going to be using `Diagnostic` everywhere and is going to need to be modified to instead use `Problem` when cherry picked onto the 6.4 branch—meaning that it wouldn't be possible to make a clean cherry pick for many changes.
- **Scope**: Renaming and deprecating of API.
- **Issues**: rdar://176555746
- **Original PRs**: #1509
- **Risk**: Low. Even thought this change is very large in number of modified lines, each steps was performed through deterministic IDE refactoring options or a series of detailed regular expressions to rename APIs, update call sites, and keep deprecated API shims where necessary. 
- **Testing**: All existing automated test pass. Some minimal manual testing confirms that a handful of example projects continue to build documentation and continue to surface warnings about issues in the documentation.
- **Reviewers**: @snprajwal


d-ronnqvist:
@swift-ci please test

franklinsch (approved):

PR #1522, opened 2026-05-13
Allow building from CMake without the toolchain's module config files

QuietMisdreavus:
Bug/issue #, if applicable: None

## Summary

This adds a couple new CMake variables to control how dependencies are loaded:

- `SWIFTDOCC_DEPS_CHECKOUTS` points to a directory where the dependency repositories are checked out. Defaults to `.build/checkouts` (which is populated by `swift package resolve` and other SwiftPM commands that build the project)
- `SWIFTDOCC_VENDOR_DEPENDENCIES`, if set to `YES`, will load dependencies from pre-existing clones in the above directory instead of falling back to the standard `find_package` behavior.

The latter option adds an alternate dependency-loading path which allows for local testing of the CMake configuration as follows:

```sh
swift package resolve
rm -rf .build/cmake
cmake -B .build/cmake -G Ninja -S . -DSWIFTDOCC_VENDOR_DEPENDENCIES=YES
(cd .build/cmake && ninja docc)
```

The goal of making these changes is to allow for testing project changes without needing to wait for a complete toolchain build. I stopped short of adding this to `bin/test`, but it could be added either there or in a CI configuration if we want to test this regularly.

Note that i had to update the pinned version of swift-lmdb because the commit in our Package.resolved file is from before the CMake configuration was added.

## Dependencies

None

## Testing

See the commands above to test the new path. Toolchain builds should still work as before; i will open a test PR on swiftlang/swift to verify.

## Checklist

- [ n/a ] Added tests
- [ ? ] Updated documentation if necessary (While writing this i realized that this could be added to the CONTRIBUTING.md file if desired. Not sure if we want that, though.)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -21,23 +21,99 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
 
 set(CMAKE_Swift_MODULE_DIRECTORY ${CMAKE_BINARY_DIR}/swift)
 set(CMAKE_Swift_COMPILE_OPTIONS_MSVC_RUNTIME_LIBRARY MultiThreadedDLL)
-# Overridden per-target.
-set(CMAKE_Swift_LANGUAGE_VERSION invalid)
 
 include(GNUInstallDirs)
 
+# The full toolchain build has find_package config files for all our
+# dependencies, but at-desk we don't have that luxury. Set this option to use
+# SwiftPM's checkouts instead.
+option(SWIFTDOCC_VENDOR_DEPENDENCIES
+  "Build dependencies from .build/checkouts instead of using find_package"
+  NO)
+set(SWIFTDOCC_DEPS_CHECKOUTS "${CMAKE_CURRENT_SOURCE_DIR}/.build/checkouts"
+  CACHE PATH "Location of SwiftPM-fetched dependency sources")
+
+if(SWIFTDOCC_VENDOR_DEPENDENCIES)
+  include(FetchContent)
+
+  macro(_docc_declare_vendored pkg subdir)
+    if(NOT EXISTS "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}")
+      message(FATAL_ERROR
+        "SWIFTDOCC_VENDOR_DEPENDENCIES is ON but ${subdir} was not found at "
+        "${SWIFTDOCC_DEPS_CHECKOUTS}. Run `swift package resolve` first.")
+    endif()
+    # OVERRIDE_FIND_PACKAGE routes both this CMakeLists' `find_package` calls
+    # *and* nested calls inside the vendored dependencies (e.g. swift-crypto
+    # looking up SwiftASN1) through our local checkouts instead of the network.
+    FetchContent_Declare(${pkg}
+      SOURCE_DIR "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}"
+      OVERRIDE_FIND_PACKAGE)
+  endmacro()
+
+  # swift-cmark needs to disable testing to build via CMake
+  set(BUILD_TESTING NO)
+
+  _docc_declare_vendored(CLMDB          swift-lmdb)
+  _docc_declare_vendored(ArgumentParser swift-argument-parser)
+  _docc_declare_vendored(SwiftASN1      swift-asn1)
+  _docc_declare_vendored(SwiftCrypto    swift-crypto)
+  _docc_declare_vendored(cmark-gfm      swift-cmark)
+  _docc_declare_vendored(SwiftMarkdown  swift-markdown)
+  _docc_declare_vendored(SymbolKit      swift-docc-symbolkit)
+
+  find_package(CLMDB          REQUIRED)
+  find_package(ArgumentParser REQUIRED)
+  find_package(SwiftASN1      REQUIRED)
+  find_package(SwiftCrypto    REQUIRED)
+  find_package(cmark-gfm      REQUIRED)
+
+  # swift-markdown's `cmake/modules/CMakeLists.txt` calls `export(TARGETS ...
+  # EXPORT_LINK_INTERFACE_LIBRARIES)` on its `Markdown` target, which pulls the
+  # private `libcmark-gfm-extensions` link dep into the exported interface and
+  # fails unless that dep belongs to a build-tree export. swift-cmark itself
+  # already exports `libcmark-gfm` (in `cmarkTargets.cmake`) but not the
+  # extensions target, so place it in a throwaway build-tree export to satisfy
+  # the check; the generated stub file is never consumed.
+  export(TARGETS libcmark-gfm-extensions
+    NAMESPACE _docc_vendor_stub::
+    FILE "${CMAKE_BINARY_DIR}/_docc_vendor_cmark_extensions_stub.cmake")
+
+  find_package(SwiftMarkdown  REQUIRED)
+  find_package(SymbolKit      REQUIRED)
+
+  # LMDB, SymbolKit, and Swift-Markdown are exported into namespaces in the
+  # toolchain build, but not when included directly like this. Re-export them
+  # into the namespaces that the DocC modules expect.
+  if(TARGET CLMDB AND NOT TARGET LMDB::CLMDB)
+    add_library(LMDB::CLMDB ALIAS CLMDB)
+  endif()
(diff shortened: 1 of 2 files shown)


QuietMisdreavus:
@swift-ci Please test

QuietMisdreavus:
I will note, this PR was written with the help of Claude, so if there's an odd construct or something that could be done better, i would love to hear it 😅

CMakeLists.txt
```diff
@@ -21,24 +21,100 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
...
+  macro(_docc_declare_vendored pkg subdir)
+    if(NOT EXISTS "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}")
+      message(FATAL_ERROR
+        "SWIFTDOCC_VENDOR_DEPENDENCIES is ON but ${subdir} was not found at "
+        "${SWIFTDOCC_DEPS_CHECKOUTS}. Run `swift package resolve` first.")
+    endif()
+    # OVERRIDE_FIND_PACKAGE routes both this CMakeLists' `find_package` calls
+    # *and* nested calls inside the vendored dependencies (e.g. swift-crypto
+    # looking up SwiftASN1) through our local checkouts instead of the network.
+    FetchContent_Declare(${pkg}
```
etcwilde, line 48: You'll need a final call to `FetchContent_MakeAvailable` to consume all of these declarations at once and resolve them.

CMakeLists.txt
```diff
@@ -21,24 +21,100 @@ set(CMAKE_RUNTIME_OUTPUT_DIRECTORY ${CMAKE_BINARY_DIR}/bin)
...
+  CACHE PATH "Location of SwiftPM-fetched dependency sources")
+
+if(SWIFTDOCC_VENDOR_DEPENDENCIES)
+  include(FetchContent)
+
+  macro(_docc_declare_vendored pkg subdir)
+    if(NOT EXISTS "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}")
+      message(FATAL_ERROR
+        "SWIFTDOCC_VENDOR_DEPENDENCIES is ON but ${subdir} was not found at "
+        "${SWIFTDOCC_DEPS_CHECKOUTS}. Run `swift package resolve` first.")
```
etcwilde, line 43: Is it only updates to the `Package.resolved` and `Package.swift` file that invalidates the checkout? We should be able to use a custom command that depends on those two files and automatically runs `swift package resolve` for you when either of those files is updated, fixing the incremental build behavior.

The custom command would look something like this:

```cmake
add_custom_command(
    OUTPUT "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
    DEPENDS 
        Package.resolved
        Package.swift
    COMMAND swift package resolve
    COMMAND "${CMAKE_COMMAND}" -E touch "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
    WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
)

add_custom_target(SwiftSourceResolve 
    DEPENDS "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
)
```

Then instead of FetchContent, we use ExternalProject. FetchContent resolves the dependency at configuration time, ExternalProject does everything at build time so that our custom command can tie into the build graph and ensure that things are always up-to-date.

```cmake
function(_docc_declare_vendored pkg subdir)
  cmake_parse_arguments(PARSE_ARGV 2 ARG "" "" "CMAKE_ARGS")

  ExternalProject_Add(${pkg}
    SOURCE_DIR       "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}"
    BINARY_DIR       "${CMAKE_BINARY_DIR}/deps/${subdir}"
    DOWNLOAD_COMMAND ""
    UPDATE_COMMAND   ""
    DEPENDS          SwiftResolve
    CMAKE_ARGS       ${ARG_CMAKE_ARGS}
  )
endfunction()
```

Something like this?

```cmake
option(SwiftDocC_VENDOR_DEPENDENCIES "Use SwiftPM to vendor dependencies" YES)

if(SwiftDocC_VENDOR_DEPENDENCIES)
  add_custom_command(
      OUTPUT "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
      DEPENDS
          Package.resolved
          Package.swift
      COMMAND swift package resolve
      COMMAND "${CMAKE_COMMAND}" -E touch "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
      WORKING_DIRECTORY "${PROJECT_SOURCE_DIR}"
  )

  add_custom_target(SwiftSourceResolve
      DEPENDS "${PROJECT_BINARY_DIR}/swift_resolve.stamp"
  )

  function(_docc_declare_vendored pkg subdir)
    cmake_parse_arguments(PARSE_ARGV 2 ARG "" "" "CMAKE_ARGS")

    ExternalProject_Add(${pkg}
      SOURCE_DIR       "${SWIFTDOCC_DEPS_CHECKOUTS}/${subdir}"
      BINARY_DIR       "${CMAKE_BINARY_DIR}/deps/${subdir}"
      DOWNLOAD_COMMAND "" # SwiftPM manages the source versions
      UPDATE_COMMAND   ""
      DEPENDS          SwiftSourceResolve
      CMAKE_ARGS       ${ARG_CMAKE_ARGS}
    )
  endfunction()

  _docc_declare_vendored(CLMDB          swift-lmdb)
  _docc_declare_vendored(ArgumentParser swift-argument-parser)
  _docc_declare_vendored(SwiftASN1      swift-asn1)
  _docc_declare_vendored(SwiftCrypto    swift-crypto)
  _docc_declare_vendored(cmark-gfm      swift-cmark
    CMAKE_ARGS -DBUILD_TESTING=NO)
  _docc_declare_vendored(SwiftMarkdown  swift-markdown)
  _docc_declare_vendored(SymbolKit      swift-docc-symbolkit)

else()
  find_package(...)
endif()
```

etcwilde (commented):
You might get asked about splitting it up so you can have some dependencies resolved by SwiftPM and others resolved by `find_package`, otherwise this seems like a reasonable direction.

Two comments. 
The first, maybe I missed it, but you'll need to pass the declared dependencies into `FetchContent_MakeAvailable` to actually pull them into your build.

The second, we might be able to tie updates to the Package.swift and Package.resolve into the build graph, which would then let the CMake build automatically download the new sources whenever those change so it's one fewer things to forget.

CMakeLists.txt
QuietMisdreavus: Looks like that doesn't work with the existing dependency declarations for DocC's own modules:

```
CMake Error at Sources/DocCCommandLine/CMakeLists.txt:70 (target_link_libraries):
  Target "ArgumentParser" of type UTILITY may not be linked into another
  target.  One may link only to INTERFACE, OBJECT, STATIC or SHARED
  libraries, or to executables with the ENABLE_EXPORTS property set.
```

Is there another call we need to add to make the `ExternalProject` version work for that?

CMakeLists.txt
snprajwal: `FetchContentDeclare` calls `FetchContent_MakeAvailable` when `OVERRIDE_FIND_PACKAGE` is present: https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_declare:

> OVERRIDE_FIND_PACKAGE
>
> When a FetchContent_Declare(<name> ...) call includes this option, subsequent calls to find_package(<name> ...) will ensure that FetchContent_MakeAvailable(<name>) has been called, then use the config package files in the [CMAKE_FIND_PACKAGE_REDIRECTS_DIR](https://cmake.org/cmake/help/latest/variable/CMAKE_FIND_PACKAGE_REDIRECTS_DIR.html#variable:CMAKE_FIND_PACKAGE_REDIRECTS_DIR) directory (which are usually created by FetchContent_MakeAvailable()). This effectively makes [FetchContent_MakeAvailable()](https://cmake.org/cmake/help/latest/module/FetchContent.html#command:fetchcontent_makeavailable) override [find_package()](https://cmake.org/cmake/help/latest/command/find_package.html#command:find_package) for the named dependency, allowing the former to satisfy the package requirements of the latter.

CMakeLists.txt
snprajwal: I don't think `ExternalProject_Add` can be made to work here. It only exposes a `UTILITY` target, which we cannot link to. Running `swift package resolve` is already a one-time prerequisite as part of the existing changes, so I feel we can defer this change for a later date. That would allow us to land the existing flow with `FetchContent`.

PR #1525, opened 2026-05-20
Print a deprecation message when running `docc process-archive index`

d-ronnqvist:
Bug/issue #, if applicable: rdar://177519864

## Summary

As proposed in [this Swift Forums thread](https://forums.swift.org/t/pitch-deprecating-the-docc-process-archive-index-subcommand/86701); this prints a deprecation message when running both the `docc process-archive index` subcommand and the hidden `docc index` command that was previously left for backwards compatibility (before we had a `process-archive` group of commands). 

> [!NOTE]
> The idea is to also cherry pick this into the 6.4 release so that developers have a longer time to see the deprecation message in case someone who isn't reading the Swift Forums is using this command.

## Dependencies

None.

## Testing

- Run `docc process-archive index /path/to/Something.doccarchive` 
  - DocC should print a warning about the deprecation and planned removal of this command.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] Added tests
- [x] Ran the `./bin/test` script and it succeeded
- [ ] Updated documentation if necessary

PR #1526, opened 2026-05-21
[6.4] Print a deprecation message when running `docc process-archive index`

d-ronnqvist:
- **Explanation**: As proposed in [this Swift Forums thread](https://forums.swift.org/t/pitch-deprecating-the-docc-process-archive-index-subcommand/86701); this prints a deprecation message when running both the `docc process-archive index` subcommand and the hidden `docc index` command that was previously left for backwards compatibility (before we had a process-archive group of commands).

  The reason for suggesting that we cherry pick this into the 6.4 release is to give any developer who uses this command (which we expect to be very few or none at all) but doesn't read the Swift Forums as much time as possible to both reach out about their use case and/or to transition to passing the the `--emit-lmdb-index` flag in their `docc convert` call instead.
- **Scope**: The `docc process-archive index` subcommand.
- **Issues**: rdar://177519864
- **Original PRs**: #1525
- **Risk**: Low. This only informs developer of a deprecation and upcoming removal of a niche command. It doesn't change any other behavior.
- **Testing**: All existing automated test pass. New tests verify that both variations of the `index` subcommand print the deprecation warning.
- **Reviewers**: @sofiaromorales

### DIFF
--- a/Sources/DocCCommandLine/ArgumentParsing/Subcommands/Index.swift
+++ b/Sources/DocCCommandLine/ArgumentParsing/Subcommands/Index.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2024 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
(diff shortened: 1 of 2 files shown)


d-ronnqvist:
@swift-ci please test

franklinsch (approved):

Issue #470, opened 2023-01-27
Fixits produce invalid references when used with the regular markdown link syntax


theMomax:
### Description

Fixits (see #438) produce invalid references when used with the regular markdown link syntax (e.g. `[doc:my/reference](doc:my/reference)`, or `[custom title](doc:my/reference)`).

The reason for that is that the code generating the replacements can only differentiate between symbol-link (` ``my/reference`` `) and non-symbol-link. It always assumes the reference to be written in DocC's custom link syntax (`<doc:my/reference>`) in the non-symbol-link case and thus uses wrong offsets when encountering a link written in the standard markdown link syntax (`[doc:my/reference](doc:my/reference)`, or `[custom title](doc:my/reference)`).

### Checklist

- [ ] If possible, I've reproduced the issue using the `main` branch of this package.
- [X] This issue hasn't been addressed in an [existing GitHub issue](https://github.com/apple/swift-docc/issues).

### Expected Behavior

Assuming that e.g. `MyClass/init()` requires the disambiguation suffix `-33vaw` for `init()`, the fixit should transform the link as follows:

`[custom title](doc:MyClass/init())` -> `[custom title](doc:MyClass/init()-33vaw)`

### Actual behavior

`[custom title](doc:MyClass/init())` -> `[custom title](doc:-33vawMyClass/init())`

### Steps To Reproduce

With `DOCC_USE_HIERARCHY_BASED_LINK_RESOLVER=YES`, run `SymbolTests/testUnresolvedReferenceWarningsInDocumentationExtension()` after adding the faulty link `- [custom title](doc:MyClass/init())` to the initial extension comment and analyse the resulting problem as it is done for the others.

### Swift-DocC Version Information

This bug was introduced in #455

### Swift Compiler Version Information

```shell
swift-driver version: 1.62.15 Apple Swift version 5.7.2 (swiftlang-5.7.2.135.5 clang-1400.0.29.51)
Target: arm64-apple-macosx13.0
```

PR #1527, opened 2026-05-24
Fix fixit offsets for references using markdown link syntax

adityasingh2400:
Bug/issue #, if applicable: #470

## Summary

Suggested replacements (fixits) for an unresolved topic reference were anchored to the wrong column when the reference was authored with the `[link text](doc:my/reference)` markdown link syntax rather than the `<doc:my/reference>` autolink syntax.

The replacements that link resolution produces (for example a disambiguation suffix to append, or a near-miss spelling correction) use columns relative to the start of the reference's path. `unresolvedReferenceDiagnostic(...)` turned those into source ranges by offsetting from the link's source range, but for non-symbol links it assumed the autolink form and inset the range by 5 to skip `<doc:`. For a markdown link that inset lands inside the display text instead of at the path, so the replacement was inserted in the wrong place:

```
[custom title](doc:MyClass/init())  ->  [custom title](doc:-33vawMyClass/init())   // before
[custom title](doc:MyClass/init())  ->  [custom title](doc:MyClass/init()-33vaw)   // after
```

This PR computes the reference body's source range from the parsed markup. Swift-Markdown doesn't expose the destination's range directly (https://github.com/swiftlang/swift-markdown/issues/109), so for a markdown link the range is derived from the link's range and the length of its destination: the destination ends just before the closing parenthesis, so it starts `destination.count` columns before it, which is correct regardless of how long the display text is. The autolink (`<doc:...>`) and symbol-link (`` ``...`` ``) insets are unchanged, so existing diagnostics for those syntaxes keep the same ranges. A markdown link with a title is left without a suggested replacement rather than risk an incorrectly placed one.

## Dependencies

None.

## Testing

Steps:
1. Add a documentation extension whose content links to symbols using the markdown link syntax, including a misspelled path and an ambiguous reference, e.g.

   ```md
   - [a near miss](doc:otherFunction())
   - [an ambiguous reference](doc:MyClass/init())
   - [a wrong disambiguation](doc:MyClass/init()-swift.init)
   ```

2. Build the documentation and apply the suggested replacement from each unresolved-reference warning. The replacement now lands at the reference's path, producing e.g. `[an ambiguous reference](doc:MyClass/init()-33vaw)`, with the display text and link syntax left intact.

This is covered by the new `SymbolTests.testUnresolvedReferenceWarningsForMarkdownLinkSyntax`, which mirrors the existing `testUnresolvedReferenceWarningsInDocumentationExtension` (that test continues to cover the autolink and symbol-link syntaxes).

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Semantics/SymbolTests.swift
```diff
@@ -948,7 +948,126 @@ class SymbolTests: XCTestCase {
         - <doc:MyClass/init()-swift.init>
         """)
     }
-    
+
+    // Verifies that suggested replacements (fixits) for unresolved references are anchored correctly when the
+    // reference is authored using the `[link text](doc:my/reference)` markdown link syntax instead of the
+    // `<doc:my/reference>` autolink syntax. https://github.com/swiftlang/swift-docc/issues/470
+    func testUnresolvedReferenceWarningsForMarkdownLinkSyntax() async throws {
+        let (url, _, context) = try await testBundleAndContext(copying: "LegacyBundle_DoNotUseInNewTests") { url in
```
d-ronnqvist: Please don't use this test fixture in any new test.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
```diff
@@ -20,15 +20,28 @@ func unresolvedReferenceDiagnostic(source: URL?, range: SourceRange?, severity:
...
+    return unresolvedReferenceDiagnostic(source: source, referenceSourceRange: referenceSourceRange, severity: severity, errorInfo: errorInfo)
+}
+
+/// Creates a diagnostic for an unresolved topic reference, anchoring any suggested replacements and notes to the source
+/// range of the link's destination computed from its parsed markup.
+///
+/// Use this overload when the original ``Markup`` link is available so that the destination's range can be computed
+/// regardless of whether the link uses the `<doc:my/reference>` autolink syntax or the `[link text](doc:my/reference)`
+/// markdown syntax.
+func unresolvedReferenceDiagnostic(source: URL?, link: any Markup, severity: DiagnosticSeverity, errorInfo: TopicReferenceResolutionErrorInfo) -> Diagnostic {
```
d-ronnqvist: FYI: If this is only intended to be passed `Link` and `SymbolLink` markup, then we have an `AnyLink` protocol that both those markup types conform to.

Sources/SwiftDocC/Utility/MarkupExtensions/SourceRangeExtensions.swift
```diff
@@ -48,3 +48,59 @@ extension SourceRange {
         return location ..< location
     }
 }
+
+extension Markup {
```
d-ronnqvist: minor: This would probably be a better extension on `AnyLink` instead to avoid having a property that return `nil` for all other types of markup.

Sources/SwiftDocC/Utility/MarkupExtensions/SourceRangeExtensions.swift
```diff
@@ -48,3 +48,59 @@ extension SourceRange {
...
+    var referenceBodySourceRange: SourceRange? {
+        guard let range = self.range else {
+            return nil
+        }
+
+        if let symbolLink = self as? SymbolLink {
+            guard symbolLink.destination != nil else { return nil }
+            // Inset the range by 2 at the start and end to skip both "``".
+            return SourceLocation(line: range.lowerBound.line, column: range.lowerBound.column + 2, source: range.lowerBound.source)
+                ..< SourceLocation(line: range.upperBound.line, column: range.upperBound.column - 2, source: range.upperBound.source)
```
d-ronnqvist: minor: It _might_ be worth adding an `inset` method (in aways similar to `offsetWithRange(_:)` above) to make the call sites of the two inset operations easier to read.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
```diff
@@ -20,15 +20,28 @@ func unresolvedReferenceDiagnostic(source: URL?, range: SourceRange?, severity:
             // Inset the range by 2 at the start and end to skip both "``".
             return SourceLocation(line: range.lowerBound.line, column: range.lowerBound.column+2, source: range.lowerBound.source) ..< SourceLocation(line: range.upperBound.line, column: range.upperBound.column-2, source: range.upperBound.source)
         } else {
-            // FIXME: This assumes that the link uses the `<doc:my/reference>` syntax.
-            // Links that use the [link text](doc:my/reference) syntax will have incorrect suggestion replacements.
-            // https://github.com/swiftlang/swift-docc/issues/470
-            
+            // This assumes that the link uses the `<doc:my/reference>` autolink syntax. Links that use the
```
d-ronnqvist: Question: Do you know how many places there are that still need to call this function? I'm wondering if it would be possible to have one code path for all these diagnostics so that callers can't accidentally use the wrong function and get an unexpectedly offset suggestion range.

Tests/SwiftDocCTests/Semantics/SymbolTests.swift
```diff
@@ -948,7 +948,126 @@ class SymbolTests: XCTestCase {
         - <doc:MyClass/init()-swift.init>
         """)
     }
-    
+
+    // Verifies that suggested replacements (fixits) for unresolved references are anchored correctly when the
+    // reference is authored using the `[link text](doc:my/reference)` markdown link syntax instead of the
+    // `<doc:my/reference>` autolink syntax. https://github.com/swiftlang/swift-docc/issues/470
+    func testUnresolvedReferenceWarningsForMarkdownLinkSyntax() async throws {
```
d-ronnqvist: Question: Was there a particular reason for adding this test in SymbolTests.swift? If not, it might be easier to write a Swift Testing test in another file that's already updated to use Swift Testing (instead of XCTest)

d-ronnqvist (commented):
Thanks for fixing this issue. The code generally looks good to me except for the added test. We don't use the "LegacyBundle_DoNotUseInNewTests" fixture in new tests. We also aim to write new tests in Swift Testing (there's some additional guidance and information on this in the contributing docs).

adityasingh2400:
Thanks for the review. I moved the test out of `SymbolTests.swift` and rewrote it in Swift Testing as `Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift`.

It no longer uses `LegacyBundle_DoNotUseInNewTests`. Instead it builds a small in-memory catalog with `load(catalog:)` and `makeSymbolGraph`/`makeSymbol`, reusing the original symbol identifiers so the disambiguation hashes stay deterministic. The coverage is unchanged: the same three suggested replacement cases for references written with the markdown link syntax (a near miss, an ambiguous reference, and a wrong disambiguation), each verified by applying the fixit to the source. I confirmed it fails on the pre-fix code and passes with the fix.

Let me know if you would prefer it to live somewhere else.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
```diff
@@ -0,0 +1,160 @@
...
+                ]
+            )),
+
+            // A documentation extension that curates three references using the markdown link syntax. Each reference is
+            // unresolvable in a different way so that it exercises a different kind of suggested replacement.
+            TextFile(name: "MyClass.md", utf8Content: markdownSource),
+        ])
+        let context = try await load(catalog: catalog)
+
+        let unresolvedTopicDiagnostics = context.diagnostics.filter { $0.identifier == "org.swift.docc.unresolvedTopicReference" }
```
d-ronnqvist: We generally don't want to ignore diagnostics in tests. Instead, it's better to make an assertion that these inputs only produce the diagnostics that we expect. For example:
```swift
#expect(diagnostics.map(\.identifier) == ["TechnologyRootInExtensionFile", "TechnologyRootInExtensionFile"],
        "Encountered unexpected problems: \(context.diagnostics.map(\.summary))")
```

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
```diff
@@ -0,0 +1,160 @@
...
+struct MarkdownLinkFixitTests {
+    /// Verifies that suggested replacements (fixits) for unresolved references are anchored correctly when the
+    /// reference is authored using the `[link text](doc:my/reference)` markdown link syntax instead of the
+    /// `<doc:my/reference>` autolink syntax.
+    ///
+    /// Because the markdown link syntax has more characters before the reference's path than the autolink syntax, a
+    /// fixit that's anchored as if the reference used the autolink syntax would be offset into the link's display text.
+    /// See https://github.com/swiftlang/swift-docc/issues/470
+    @Test
+    func anchorsFixitsToReferencePathForMarkdownLinkSyntax() async throws {
```
d-ronnqvist: We try to name tests so that the name itself would be a description of what's being verified—so that there isn't a need for a comment explaining the purpose of the test.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
```diff
@@ -0,0 +1,160 @@
...
+
+    ### Near Miss
+
+    - [a near miss](doc:otherFunction())
+
+    ### Ambiguous curation
+
+    - [an ambiguous reference](doc:MyClass/init())
+    - [a wrong disambiguation](doc:MyClass/init()-swift.init)
+    """
```
d-ronnqvist: The topics sections makes no difference for this test—and there's no documentation to override—so this markup could be significantly shortened. 

```suggestion
    # ``MyKit/MyClass``

    - [a near miss](doc:otherFunction())
    - [an ambiguous reference](doc:MyClass/init())
    - [a wrong disambiguation](doc:MyClass/init()-swift.init)
    """
```

We sometimes include an abstract that contextualizes the content in terms of what's being tested—but "A cool API to call." doesn't describe the purpose of this file and would be better to remove.

```suggestion
    # ``MyKit/MyClass``
    
    Three links with different issues; their solutions should only modify the link portion of the syntax.

    - [a near miss](doc:otherFunction())
    - [an ambiguous reference](doc:MyClass/init())
    - [a wrong disambiguation](doc:MyClass/init()-swift.init)
    """
```

Strictly speaking `doc:MyClass/init()-swift.init` is also an ambiguous link. Because both overloads are initializers, it doesn't help disambiguate the link to specify `-init` (the `swift.` prefix is redundant).

If you wanted to test a disambiguation that's incorrect you could either use a type that doesn't match either of the overloads (for example `-class`) or a hash that doesn't match (for example `-h4sh`).

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
```diff
@@ -0,0 +1,160 @@
...
+            TextFile(name: "MyClass.md", utf8Content: markdownSource),
+        ])
+        let context = try await load(catalog: catalog)
+
+        let unresolvedTopicDiagnostics = context.diagnostics.filter { $0.identifier == "org.swift.docc.unresolvedTopicReference" }
+
+        // A near miss replaces only the misspelled path component, keeping the surrounding markdown link intact.
+        var diagnostic = try #require(unresolvedTopicDiagnostics.first(where: { $0.summary == "'otherFunction()' doesn't exist at '/MyKit/MyClass'" }))
+        #expect(diagnostic.solutions.count == 1)
+        #expect(diagnostic.solutions.map { [$0.summary, $0.replacements.first!.replacement] } == [
```
d-ronnqvist: If this warning regresses in the future, this force unwrap will trap and prevent the other tests from running. It would be better to either compare against an optional value or to use `#require` to unwrap the value if it's inconvenient to compare with the optional value (for example if multiple comparisons are made to different properties of the unwrapped value)

Tests/SwiftDocCTests/Semantics/SymbolTests.swift
```diff
@@ -948,7 +948,7 @@ class SymbolTests: XCTestCase {
         - <doc:MyClass/init()-swift.init>
         """)
     }
-    
+
```
d-ronnqvist: nit: This file is not modified other than the whitespace change.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
```diff
@@ -0,0 +1,160 @@
...
+                    makeSymbol(id: "s:5MyKit0A5ClassC10myFunctionyyF",  kind: .method, pathComponents: ["MyClass", "myFunction()"]),
+                    makeSymbol(id: "s:5MyKit0A5ClassCACycfc",           kind: .`init`, pathComponents: ["MyClass", "init()"]),
+                    makeSymbol(id: "s:5MyKit0A5ClassCACycfcDUPLICATE",  kind: .`init`, pathComponents: ["MyClass", "init()"]),
+                ],
+                relationships: [
+                    SymbolGraph.Relationship(source: "s:5MyKit0A5ClassC10myFunctionyyF", target: "s:5MyKit0A5ClassC", kind: .memberOf, targetFallback: nil),
+                    SymbolGraph.Relationship(source: "s:5MyKit0A5ClassCACycfc",          target: "s:5MyKit0A5ClassC", kind: .memberOf, targetFallback: nil),
+                    SymbolGraph.Relationship(source: "s:5MyKit0A5ClassCACycfcDUPLICATE", target: "s:5MyKit0A5ClassC", kind: .memberOf, targetFallback: nil),
+                ]
+            )),
```
d-ronnqvist, line 29: The reasons for not using the legacy fixture also applies to most of its content. 

It would be better to think about what the smallest amount of inputs are that are necessary to reproduce this issue. In this case; 
- This could be reproduced with only top-level symbols, so you don't need any of the `memberOf` relationships.
- The extension file could document the module instead, so you don't need to define "MyClass".
- The near-miss link could be a misspelling of the overloaded symbol name, so you don't need _both_ the function and the initializers. 
- The symbol IDs are opaque strings, so we generally try to use more descriptive spellings (for example `"first-symbol-id"` & `"second-symbol-id"` or `"some-overload-id"` & `"other-overload-id"`)

adityasingh2400:
Thanks for the thorough review, this all made the change a lot cleaner. Here's what I did for each theme:

**AnyLink.** The markup-aware `unresolvedReferenceDiagnostic(source:link:...)` now takes `any AnyLink` instead of `any Markup`, since it's only ever meant for `Link` and `SymbolLink`.

**One diagnostic path.** I traced the callers and consolidated. Every link-bearing caller now goes through the `link:` overload, which computes the destination range from the parsed markup. `MarkupReferenceResolver` always passes the parsed link, so its old range-based fallback branch is gone. The only remaining caller without link markup is the `@TutorialReference` directive path in `ReferenceResolver`, so I kept the range-based function but made it `private` to `ReferenceResolver.swift`. That way a link-bearing caller can't accidentally reach for the range-based function and get an offset suggestion range, which I think was the concern.

**referenceBodySourceRange moved to AnyLink.** It's now an extension on `AnyLink` rather than `Markup`, so it's no longer a property that returns `nil` for unrelated markup. I also added a small `SourceRange.inset(startColumns:endColumns:)` helper and used it for the inset operations in both the property and the range-based fallback.

**SymbolTests whitespace.** Reverted, that file now has no diff against `main`.

**Test rewrite.** Renamed it so the name describes what it verifies and dropped the explanatory doc comment. It now asserts the exact set of diagnostics (`context.diagnostics.map(\.identifier)`) so an unexpected diagnostic fails the test instead of being filtered away. The inputs are minimal: two overloaded top-level symbols sharing a path (no `memberOf` relationships, no `MyClass`), descriptive IDs (`some-overload-id` / `other-overload-id`), a module-documenting extension file, and a trimmed markdown source with just the near miss, the ambiguous reference, and a wrong disambiguation. For the wrong disambiguation I used `-class`, which genuinely can't match two functions (it now produces an "isn't a disambiguation" diagnostic). Force-unwraps are replaced with `try #require`. I rebuilt and ran the test to capture the real disambiguation hashes (`-1j23c` / `-7bbou`) and the assertions check that each solution only modifies the link's destination, leaving the display text intact, which is what verifies the fix.

Built and ran the test locally, it passes, and `bin/check-source` is clean.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
adityasingh2400: Good call, switched it to `any AnyLink`.

Sources/SwiftDocC/Semantics/ReferenceResolver.swift
adityasingh2400: I dug into the callers. There were two using the range-based function: `MarkupReferenceResolver` (which always has the parsed link) and the `@TutorialReference` directive path in `ReferenceResolver` (which doesn't). I routed `MarkupReferenceResolver` entirely through the markup-aware `link:` overload so its range-based branch is gone, and made the range-based function `private` to this file as the fallback for the directive path. So link-bearing callers can no longer reach the range-based function and get an offset suggestion range.

Sources/SwiftDocC/Utility/MarkupExtensions/SourceRangeExtensions.swift
adityasingh2400: Moved it onto `AnyLink` so it's no longer a property that returns `nil` for unrelated markup.

Sources/SwiftDocC/Utility/MarkupExtensions/SourceRangeExtensions.swift
adityasingh2400: Added `SourceRange.inset(startColumns:endColumns:)` and used it for the inset operations here and in the range-based fallback.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
adityasingh2400: Done. It now asserts the exact set with `context.diagnostics.map(\.identifier)`, so any unexpected diagnostic fails the test.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
adityasingh2400: Renamed to `suggestedReplacementsAnchorToTheLinkDestinationForMarkdownLinkSyntax` and removed the explanatory doc comment.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
adityasingh2400: Trimmed the markup down to the three links and used your one-line abstract about the solutions only modifying the link's destination. For the wrong disambiguation I switched to `-class`, which can't match two functions. It now produces an "isn't a disambiguation" diagnostic, which is a cleaner case than the redundant `-swift.init`.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
adityasingh2400: Replaced the force-unwraps with `try #require` so a future regression fails cleanly instead of trapping the whole suite.

Tests/SwiftDocCTests/Semantics/SymbolTests.swift
adityasingh2400: Reverted, SymbolTests.swift now has no diff against `main`.

Tests/SwiftDocCTests/Semantics/Reference/MarkdownLinkFixitTests.swift
adityasingh2400: Reworked along these lines: two overloaded top-level symbols sharing a path (no `memberOf`), the extension file documents the module instead of `MyClass`, the near miss is a misspelling of the overloaded name, and the IDs are now `some-overload-id` and `other-overload-id`.

parispittman:
Hey @adityasingh2400 - are you on the Swift forums? I see a lot of your contributions to Swift (thank you!! that rocks) and wanted to talk to you about shaping your work for better outcomes. You can post a topic about this in the #contributor-experience category or DM me. I have a similar UN there.

PR #1528, opened 2026-05-25
Add early work-in-progress integration to output static HTML to enable more people to contribute to it 

d-ronnqvist:
Bug/issue #, if applicable: rdar://165912669

## Summary

> [!NOTE]
> The primary purpose of this PR is to put together enough scaffolding in place to allow people to contribute on improving the static HTML output; both in terms of refining the layout and appearance and in terms of adding HTML output for more page elements. As such, some pieces are a bit more rough than what I would normally include in a PR. However, I believe that those can be cleaned up over time and feel that there are worthwhile benefits to opening this area up to contributions sooner.

This is a collection of rough changes pieced together from a handful of explorations into static HTML output from on and off over the past few of months. 

Most of those changes (800+ lines) are in the form of CSS to adjust the layout and appearance of the minimal semantic HTML that DocC can already output. Some of this is pixel-perfect recreations of the DocC-Render appearance—with some rules taken directly from there—and some of this very quick, good enough, first versions. I'm certain that this stylesheet can be greatly improved over time. Notably; many of the defined variables are never used and the stylesheet doesn't define any dark appearance colors. There's an additional ~100 lines of SVG icons where the paths are taken directly from DocC-Render. In the long run we might be able to make some of these more readable and/or better aligned to pixels for their display sizes.

The code comment in [StaticResources.swift](https://github.com/swiftlang/swift-docc/compare/main...d-ronnqvist:swift-docc:experimental-html-output?expand=1#diff-db113cd4f84968d9630a84ddfa834d5e4b51585c7d70ae39e71826f4b809ee7b) gives a brief explanation of why I've opted for code constants in this PR; it's much simpler to reference and submit to toolchains and it's two orders of magnitude less data than the DocC-Render Template (~18kB instead of ~1.2MB). If this becomes a problem in the future we can go down the same route as the DocC-Render Template but until then I'd prefer to avoid that if possible.

In terms of long term intentional design, there are two parts among these changes that in the ballpark of what my long term plan:
- Separating the HTML document creation from its serialization makes it easier to test behaviors of the HTML output.
- I believe that a `--output-format` CLI option will be extensible both if we want to add more complete formats (for example ePUB) and if we want options to output only the metadata (linkable entities, indexing records, etc.) without outputting pages in any format.

and one part that needs future design work:
- I'm not sure how the `ConvertAction` and `ConvertActionConverter` can best communicate which output files are needed, which responsibilities belong in `DocCCommandLine` and which belong in `SwiftDocC`, and how to manage this complexity with future hypothetical output formats. As pointed out in a few code comments mentioning "rdar://177867282"; I don't have a clear picture yet for how we'd configure which parts of the "render" output that a given output format wants and where that logic is best handled. At this point it would take some time, and additional prototyping / exploration, to try and answer those questions, which would be time where it's difficult for people to contribute to this output. By ensuring that nothing of this temporary design leaks up into public API and by adding code comments on pieces that aren't meant to be long-lived, my plan is to refine this design over time, in parallel to people's refinements of the static HTML output.

> [!NOTE]
> My hope is that this can be merged—so that people can start contributing in this area—without requiring a pitch and discussion in the Swift Forums. There's really only the `--output-format` flag that is user-facing _but hidden_. If you feel that this needs to be discussed in the Swift Forums before merging; I'd ask you to think about if there's anything I could do to further hide or dissuade from using this output format that would make that unnecessary that need. 
>
> I plan on making a broader pitch for this—including ideas for how it will support the on-page sidebar (very minimally at first with additional features being added over time), custom theming, custom page components, etc.—once it's getting closer to the point of being usable. However, right now, I feel that it's too early to have that discussion. I haven't explored the on-page sidebar solutions in sufficient detail yet, so it will take a lot more research before I can have a full pitch available to discuss and I would very much like to facilitate people contributing to this feature before then.

Lastly, this PR contains a few miscellaneous page refinements (on-hover icons for headings, icons for API collections in curation, backgrounds for "hero" sections, a visual—but not working—page headings and page footer) from various explorations. I felt that it made sense to collect these as well so that people wouldn't need to redo work that I had already done.

---- 

To give a few examples of what this early work-in-progress pages look like today; here are a two examples from the Swift standard library (the first also showing the in-progress recreation of DocC-Render's response design):

swift-method
swift-method-responsive
swift-module

## Dependencies

None.

## Testing

- Build some reference documentation and pass `--output-format experimental-html-for-development`
- Navigate to one of the HTML pages inside the output location and open it directly (without starting a local web server)
  - You should see a page that looks very similar to the DocC-Render version of those pages (although without an on-page sidebar, and a few other features).
  
As mentioned above; these changes aren't primarily for people to _use_. Instead, they're intended for local development workflows where people can iterate on improving this static HTML output.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
@@ -25,8 +25,9 @@ private import DocCHTML
 struct FullPageHTMLContentConsumer: HTMLContentConsumer {
     let _isPrimaryOutputFormat = true
     
-    private var customHeader: XMLNode?
-    private var customFooter: XMLNode?
+    private let customHeader: XMLNode?
+    private let customFooter: XMLNode?
+    // FIXME: Extract the file writing (and directory creation) functionality from this RenderNode (JSON) specific type.
     private let fileWriter: JSONEncodingRenderNodeWriter
     
     init(
--- a/Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
@@ -1,0 +1,58 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2026 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+#if canImport(FoundationXML)
+// TODO: Consider other HTML rendering options as a future improvement (rdar://165755530)
+import FoundationXML
+import FoundationEssentials
+#else
+import Foundation
+#endif
+
+import SwiftDocC
+private import DocCHTML
+
+// I'm not convinced that this is the best long-term design for HTML as a primary output format.
+// It currently exist because the consumers are created in the DocCCommandLine target but the core logic is in SwiftDocC.
+// There might be a different internal abstraction where we don't need this type at all. (rdar://177867282)
+struct FullPageHTMLContentConsumer: HTMLContentConsumer {
+    let _isPrimaryOutputFormat = true
+    
+    private var customHeader: XMLNode?
+    private var customFooter: XMLNode?
+    private let fileWriter: JSONEncodingRenderNodeWriter
+    
+    init(
+        targetFolder: URL,
+        fileManager: some FileManagerProtocol,
+        customHeader: URL?,
+        customFooter: URL?
+    ) throws {
+        (self.customHeader, self.customFooter) = try HTMLRenderer.prepareForFullPage(customHeader: customHeader, customFooter: customFooter, fileManager: fileManager)
+        
+        self.fileWriter = JSONEncodingRenderNodeWriter(
+            targetFolder: targetFolder,
+            fileManager: fileManager,
+            transformForStaticHostingIndexHTML: nil
+        )
+    }
+    
+    func consume(
+        mainContent: XMLNode,
+        metadata: (title: String, description: String?),
+        forPage reference: ResolvedTopicReference
+    ) throws {
+        let page = HTMLRenderer.makeFullPage(mainContent: mainContent, metadata: metadata, for: reference, customHeader: customHeader, customFooter: customFooter)
+        
+        let htmlString = page.xmlString(options: .nodeCompactEmptyElement)
+        let relativeFilePath = NodeURLGenerator.fileSafeReferencePath(reference, lowercased: true) + "/index.html"
+        try fileWriter.write(Data(htmlString.utf8), toFileSafePath: relativeFilePath)
+    }
+}
--- a/Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
+++ b/Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
@@ -144,8 +144,13 @@ func assert(_ html: XMLDocument, matches expectedHTML: String, sourceLocation: S
     func formatForTestComparison(_ xmlString: String) -> String {
         // This is overly simplified and won't result in "pretty" XML for general use but sufficient for test content comparisons
         xmlString
+            // Workaround document type differences
+            .replacingOccurrences(of: #"<?xml version="1.0" encoding="utf-8" standalone="no"?>"#, with: "")
+            .replacingOccurrences(of: #"<!DOCTYPE html PUBLIC "" "">"#, with: "<!DOCTYPE html>")
             // Put each tag on its own line
             .replacingOccurrences(of: ">", with: ">\n")
+            // Allow some meta tags to encode as void elements rather than self-closing elements
+            .replacingOccurrences(of: "\">", with: "\"/>")
             // Remove leading indentation
             .components(separatedBy: .newlines)
             .map { $0.trimmingCharacters(in: .whitespacesAndNewlines) }
@@ -154,6 +159,8 @@ func assert(_ html: XMLDocument, matches expectedHTML: String, sourceLocation: S
             // Explicitly escape a few HTML characters that appear in the test content
             .replacingOccurrences(of: "–", with: "&#x2013;") // en-dash
             .replacingOccurrences(of: "—", with: "&#x2014;") // em-dash
+            // Shorten empty string attribute values
+            .replacingOccurrences(of: #"="""#, with: "")
     }
     
     let actualHTML: String = html.xmlString(options: .nodeCompactEmptyElement)
--- a/Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
+++ b/Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
@@ -76,12 +76,6 @@ struct HTMLRenderFullPageTests {
     
     @Test
     func includesCustomHeaderAndFooterInFullPage() async throws {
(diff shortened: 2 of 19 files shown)


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test Linux

d-ronnqvist:
Hm.. it seems that `Foundation.XMLDocument` _on Linux_ encodes a "html" document type as 
```xml
<?xml version="1.0" encoding="utf-8" standalone="no"?>
<!DOCTYPE html PUBLIC "" "">
```
rather than 
```xml
<!DOCTYPE html>
```
(like how it's encoded on macOS) which is causing the new full-page tests to fail. 

I'll look for a workaround for that in the tests.

d-ronnqvist:
@swift-ci please test

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
```diff
@@ -0,0 +1,58 @@
...
+    
+    init(
+        targetFolder: URL,
+        fileManager: some FileManagerProtocol,
+        customHeader: URL?,
+        customFooter: URL?
+    ) throws {
+        (self.customHeader, self.customFooter) = try HTMLRenderer.prepareForFullPage(customHeader: customHeader, customFooter: customFooter, fileManager: fileManager)
+        
+        self.fileWriter = JSONEncodingRenderNodeWriter(
```
snprajwal, line 41: We're using `JSONEncodingRenderNodeWriter` to write HTML here, it might be worth using an alias to name the type more appropriately now.

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
```diff
@@ -0,0 +1,58 @@
...
+private import DocCHTML
+
+// I'm not convinced that this is the best long-term design for HTML as a primary output format.
+// It currently exist because the consumers are created in the DocCCommandLine target but the core logic is in SwiftDocC.
+// There might be a different internal abstraction where we don't need this type at all. (rdar://177867282)
+struct FullPageHTMLContentConsumer: HTMLContentConsumer {
+    let _isPrimaryOutputFormat = true
+    
+    private var customHeader: XMLNode?
+    private var customFooter: XMLNode?
```
snprajwal: This is used within `_concurrentPerform` in the converter, but `XMLNode` is [not sendable](https://developer.apple.com/documentation/foundation/xmlnode#conforms-to). If there are multiple pages being rendered with a custom header and footer, this might cause it to crash when `body.addChild(customHeader)` is called.

Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
```diff
@@ -0,0 +1,168 @@
...
+                Auto</label>
+              </fieldset>
+            </footer>
+          </body>
+        </html>
+        """)
+    }
+    
+    @Test
+    func includesCustomHeaderAndFooterInFullPage() async throws {
```
snprajwal, line 78: Adding another page in this test should reproduce the crash mentioned above.

snprajwal (commented):
This is great, thanks for working on this David. Mostly LGTM with one concurrency concern.

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
d-ronnqvist: Yeah, it's unfortunate that some general purpose core functionality is tied to various RenderNode-specific types. Instead of a typealias I added a FIXME comment here (and in the in-page `FileWritingHTMLContentConsumer`) to track extracting that functionality out of the render node writer.

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
d-ronnqvist: As far as I know and as far as I can tell, sharing an immutable `XMLNode` across threads doesn't cause any issue. We are also accessing `DocumentationContext`, `DocumentationContextConverter`, and several other transient values in the same `_concurrentPerform` body and that isn't causing any crashes either—even though all none of those types are marked as `Sendable`.

Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
d-ronnqvist: It doesn't. I added thousands of pages and ran this with the Thread Sanitizer enabled it it neither crashed not reported any data races.

Rendering multiple pages with the same header and footer `XMLNode` did however reveal a precondition failure that those nodes need to be explicitly copied (because it's not allowed to pass a node that already has a parent to `addChild(_:)`. I added these explicit copies in https://github.com/swiftlang/swift-docc/pull/1528/commits/be0f3d2ea6ff489d2c0c24a26a4bf0b591c8f91a

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
d-ronnqvist: I added that FIXME comment in https://github.com/swiftlang/swift-docc/pull/1528/commits/c58da3ad7cf0ef565275cececd796c0e8230c6d4

d-ronnqvist:
@swift-ci please test

Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
snprajwal: I added a second reference and call to `makeFullPage` in this test, and it crashed with the below message:

```
*** Terminating app due to uncaught exception 'NSInternalInconsistencyException', reason: 'Cannot add a child that has a parent; detach or copy first'
*** First throw call stack:
...<omitted for brevity>
libc++abi: terminating due to uncaught exception of type NSException
```

This has been fixed by the explicit copies, thanks :)

Sources/DocCCommandLine/Action/Actions/Convert/FullPageHTMLContentConsumer.swift
snprajwal: From what I understand, the reference to `XMLNode` is immutable, but it can be used to mutate the underlying value. `addChild` modifies the child node's pointer to its [parent](https://developer.apple.com/documentation/foundation/xmlnode/parent), so I was wondering if that would be a problem. As shared in the other comment thread below, this situation crashes due to a guardrail to prevent this anyway:

```
'Cannot add a child that has a parent; detach or copy first'
```

So we should be good now with the copies that were introduced.

snprajwal (approved):

Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
d-ronnqvist: Right, but as far as I understand that issue isn't about the node isn't sendable but rather that the node is a class and the shared reference cannot be added in two different places. 

The original comment about lack of sendability confused me, that's all.

d-ronnqvist:
@swift-ci please test macOS

Tests/SwiftDocCTests/Rendering/HTML/HTMLRenderer+FullPageTests.swift
snprajwal: Ah right, I implicitly assumed that if the type were sendable, then the API would be designed in a way where it would prevent this issue. Realised later on that they're not necessarily related.

PR #1530, opened 2026-05-27
Add alignment parameter to @Column directive

VictorPuga:
## Summary

This PR implements alignment control for the `@Column` directive, enabling documentation authors to align content.

The feature adds an optional `alignment` parameter to `@Column` that accepts `leading`, `center`, or `trailing` values, defaulting to `leading`.
This addresses the pitch at https://forums.swift.org/t/pitch-adding-alignment-control-to-the-column-directive/86754.

### Implementation Overview

**Semantic Layer Changes:**

- Adds `Row.ColumnAlignment` enum with three cases: `leading`, `center`, and `trailing`
- Parses the optional `alignment` parameter from `@Column` directive syntax
- Defaults to `.leading` when the parameter is not specified
- Produces a conversion-failed diagnostic when an invalid alignment value is encountered

**Render Layer Changes:**

- Introduces parallel `RenderBlockContent.Row.ColumnAlignment` enum to maintain clean separation between semantic and render models
- Converts semantic alignment values to their render equivalents

**Documentation Updates:**

- Refines Column directive documentation with concise parameter description
- Regenerates `docc.symbols.json` symbol graph to include the new alignment parameter and its doc comment entry
- Provides examples showcasing all three alignment options (default, center, and trailing)

**Test Coverage:**

- 14 comprehensive test cases covering:
  - All valid alignment values with proper conversion
  - Default `.leading` behavior when alignment is omitted
  - Invalid/unknown alignment values with appropriate diagnostics
  - Rendering layer conversion for all alignment states

## Dependencies

This PR is a pre-requisite for the corresponding swift-docc-render changes that will implement the actual rendering logic for the new alignment options. https://github.com/swiftlang/swift-docc-render/pull/1018

## Testing

1. Create a test file at `Sources/docc/DocCDocumentation.docc/TestAlignment.md` with the following content:

   ```markdown
   # Test Alignment

   ## Default (leading) alignment

   @Row {
       @Column {
           Foo
       }
   }

   ## Center alignment

   @Row {
       @Column(alignment: center) {
           Bar
       }
   }

   ## Trailing alignment

   @Row {
       @Column(alignment: trailing) {
           Baz
       }
   }
   ```

2. Preview the documentation to visually verify alignment rendering:

   ```bash
   ./bin/preview-docs DocC
   ```

3. Open http://localhost:8080/documentation/docc/testalignment in your browser and verify that all three rows display different column alignments correctly.

## Checklist

- [x] Added tests (14 new test cases in `RowTests.swift`)
- [x] Updated documentation with alignment parameter examples

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -818,7 +818,7 @@ public enum RenderBlockContent: Equatable {
             public var size: Int
 
             /// The alignment of this column's content.
-            public var alignment: ColumnAlignment?
+            public var alignment: ColumnAlignment = .leading
 
             /// The content that should be rendered in this column.
             public var content: [RenderBlockContent]
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -798,15 +798,28 @@ public enum RenderBlockContent: Equatable {
         /// individual columns that span multiple columns (specified with the column's
         /// ``Column/size`` property) or the row could be not fully filled with columns.
         public var numberOfColumns: Int
-        
+
         /// The columns that should be rendered in this row.
         public var columns: [Column]
-        
+
+        /// The alignment of content within a column.
+        public enum ColumnAlignment: String, Codable, Equatable {
+            /// Align to the leading edge.
+            case leading
+            /// Align to the center.
+            case center
+            /// Align to the trailing edge.
+            case trailing
+        }
+
         /// A column with a row in a grid-based layout system.
         public struct Column: Codable, Equatable {
             /// The number of columns in the parent row this column should span.
             public var size: Int
-            
+
+            /// The alignment of this column's content.
+            public var alignment: ColumnAlignment?
+
             /// The content that should be rendered in this column.
             public var content: [RenderBlockContent]
         }
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -798,19 +798,9 @@ public enum RenderBlockContent: Equatable {
         /// individual columns that span multiple columns (specified with the column's
         /// ``Column/size`` property) or the row could be not fully filled with columns.
         public var numberOfColumns: Int
-
+        
         /// The columns that should be rendered in this row.
         public var columns: [Column]
-
-        /// The alignment of content within a column.
-        public enum ColumnAlignment: String, Codable, Equatable {
-            /// Align to the leading edge.
-            case leading
-            /// Align to the center.
-            case center
-            /// Align to the trailing edge.
-            case trailing
-        }
 
         /// A column with a row in a grid-based layout system.
         public struct Column: Codable, Equatable {
@@ -818,10 +808,20 @@ public enum RenderBlockContent: Equatable {
             public var size: Int
 
             /// The alignment of this column's content.
-            public var alignment: ColumnAlignment = .leading
+            public var alignment: Alignment = .leading
 
             /// The content that should be rendered in this column.
             public var content: [RenderBlockContent]
+
+            /// The alignment of content within a column.
+            public enum Alignment: String, Codable, Equatable {
+                /// Align to the leading edge.
+                case leading
+                /// Align to the center.
+                case center
+                /// Align to the trailing edge.
+                case trailing
+            }
         }
     }
     
--- a/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
+++ b/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
@@ -803,12 +803,12 @@ public enum RenderBlockContent: Equatable {
         public var columns: [Column]
 
         /// A column with a row in a grid-based layout system.
-        public struct Column: Codable, Equatable {
+        public struct Column: Equatable {
             /// The number of columns in the parent row this column should span.
             public var size: Int
 
             /// The alignment of this column's content.
-            public var alignment: Alignment = .leading
+            public var alignment: Alignment
 
             /// The content that should be rendered in this column.
             public var content: [RenderBlockContent]
@@ -1004,6 +1004,26 @@ extension RenderBlockContent.Table: Codable {
                 try cellContainer.encode(data.rowspan, forKey: .rowspan)
             }
         }
+    }
+}
+
+extension RenderBlockContent.Row.Column: Codable {
+    enum CodingKeys: String, CodingKey {
+        case size, alignment, content
+    }
+
+    public init(from decoder: any Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        size = try container.decode(Int.self, forKey: .size)
+        alignment = try container.decodeIfPresent(Alignment.self, forKey: .alignment) ?? .leading
+        content = try container.decode([RenderBlockContent].self, forKey: .content)
+    }
+
+    public func encode(to encoder: any Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encode(size, forKey: .size)
+        try container.encode(alignment, forKey: .alignment)
+        try container.encode(content, forKey: .content)
     }
 }
 
--- a/Sources/SwiftDocC/Semantics/Reference/Row.swift
+++ b/Sources/SwiftDocC/Semantics/Reference/Row.swift
@@ -110,6 +110,9 @@ extension Row {
     ///     }
     /// }
     /// ```
+    ///
+    /// > Earlier Versions:
+    /// > Before Swift-DocC 6.4, the `alignment` parameter wasn't supported for `@Column`.
     public final class Column: Semantic, AutomaticDirectiveConvertible, MarkupContaining {
         public static let introducedVersion = "5.8"
         public let originalMarkup: BlockDirective
--- a/Sources/SwiftDocC/Semantics/Reference/Row.swift
+++ b/Sources/SwiftDocC/Semantics/Reference/Row.swift
@@ -106,7 +106,7 @@ extension Row {
     ///     }
     ///
     ///     @Column(size: 2, alignment: trailing) {
-    ///         Trailing alignment
+    ///         Trailing alignment; twice as wide
     ///     }
     /// }
     /// ```
--- a/Sources/SwiftDocC/Semantics/Reference/Row.swift
+++ b/Sources/SwiftDocC/Semantics/Reference/Row.swift
@@ -50,6 +50,22 @@ public import Markdown
 public final class Row: Semantic, AutomaticDirectiveConvertible, MarkupContaining {
     public static let introducedVersion = "5.8"
     public let originalMarkup: BlockDirective
+
+    /// The alignment of content within a column.
+    public enum ColumnAlignment: String, CaseIterable, DirectiveArgumentValueConvertible {
+        /// Align to the leading edge.
+        case leading
+
+        /// Align to the center.
+        case center
+
+        /// Align to the trailing edge.
+        case trailing
+
+        public init?(rawDirectiveArgumentValue: String) {
+            self.init(rawValue: rawDirectiveArgumentValue)
+        }
+    }
     
     /// The number of columns available in this row.
     @DirectiveArgumentWrapped(name: .custom("numberOfColumns"))
@@ -103,13 +119,18 @@ extension Row {
         /// in the parent ``Row``.
         @DirectiveArgumentWrapped
         public private(set) var size: Int = 1
-        
+
+        /// The alignment of this column's content.
+        @DirectiveArgumentWrapped
+        public private(set) var alignment: Row.ColumnAlignment? = nil
+
         /// The markup content in this column.
         @ChildMarkup(numberOfParagraphs: .zeroOrMore, supportsStructure: true)
         public private(set) var content: MarkupContainer
         
         static var keyPaths: [String : AnyKeyPath] = [
             "size"      : \Column._size,
+            "alignment" : \Column._alignment,
             "content"   : \Column._content,
         ]
         
@@ -133,8 +154,12 @@ extension Row {
 extension Row: RenderableDirectiveConvertible {
     func render(with contentCompiler: inout RenderContentCompiler) -> [any RenderContent] {
         let renderedColumns = columns.map { column in
+            let alignment = column.alignment.flatMap { semanticAlignment in
+                RenderBlockContent.Row.ColumnAlignment(rawValue: semanticAlignment.rawValue)
+            }
             return RenderBlockContent.Row.Column(
                 size: column.size,
+                alignment: alignment,
                 content: column.content.elements.flatMap { markupElement in
                     return contentCompiler.visit(markupElement) as! [RenderBlockContent]
                 }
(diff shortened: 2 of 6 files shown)


Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -798,15 +798,28 @@ public enum RenderBlockContent: Equatable {
...
+        }
+
         /// A column with a row in a grid-based layout system.
         public struct Column: Codable, Equatable {
             /// The number of columns in the parent row this column should span.
             public var size: Int
-            
+
+            /// The alignment of this column's content.
+            public var alignment: ColumnAlignment = .leading
```
sofiaromorales: Adding a non-optional stored property to this type breaks decoding of any render JSON produced by previous versions of DocC. `RenderBlockContent.Row.Column` is a public `Codable` type that gets decoded from .doccarchive render JSON here:

https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift#L1077.

This requires adding a custom decoder to `Column` to decode the alignment 
```swift
self.alignment = try container.decodeIfPresent(ColumnAlignment.self, forKey: .alignment) ?? .leading
```

sofiaromorales (changes requested):

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -798,15 +798,28 @@ public enum RenderBlockContent: Equatable {
         /// individual columns that span multiple columns (specified with the column's
         /// ``Column/size`` property) or the row could be not fully filled with columns.
         public var numberOfColumns: Int
-        
+
```
sofiaromorales: This change is not relevant for the implementation.

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
```diff
@@ -204,33 +204,34 @@ class RowTests: XCTestCase {
             """
             @Row {
                 @Column
-            
+
```
sofiaromorales: This change is not relevant for the implementation. Same applies for most of the changes made to this file.

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
```diff
@@ -798,15 +798,28 @@ public enum RenderBlockContent: Equatable {
...
         /// The columns that should be rendered in this row.
         public var columns: [Column]
-        
+
+        /// The alignment of content within a column.
+        public enum ColumnAlignment: String, Codable, Equatable {
```
sofiaromorales: This enum is only relevant to `Column`. I advise moving it inside `Column` and renaming it to `Alignment`.

This would make readability better when accessing the enum cases as it would be `Row.Column.Alignment.<case>`.

Sources/SwiftDocC/Semantics/Reference/Row.swift
sofiaromorales: Please, update the license header year to `2022-2026`

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
sofiaromorales: Please, update the license header to `2021-2026`.

sofiaromorales (dismissed):

Sources/SwiftDocC/Semantics/Reference/Row.swift
d-ronnqvist: FYI: we have a little Swift helper tool in the repo that can make these changes for you. If you run `swift run --package-path bin/update-license-comments update-license-for-modified-files --help` you can see 3 example uses cases and their corresponding commands.

Sources/docc/DocCDocumentation.docc/docc.symbols.json
```diff
@@ -2994,12 +3087,6 @@
           {
             "text" : "- Parameters:"
           },
-          {
-            "text" : "  - source: A reference to the source file for the media item."
-          },
-          {
-            "text" : "     **(required)**"
-          },
```
d-ronnqvist, line 3002: We don't manually modify this file. If this was previously removed from the in-source documentation then we don't expect it to be in the symbol graph file either.

Sources/SwiftDocC/Semantics/Reference/Row.swift
```diff
@@ -50,6 +50,22 @@ public import Markdown
 public final class Row: Semantic, AutomaticDirectiveConvertible, MarkupContaining {
     public static let introducedVersion = "5.8"
     public let originalMarkup: BlockDirective
+
+    /// The alignment of content within a column.
+    public enum ColumnAlignment: String, CaseIterable, DirectiveArgumentValueConvertible {
```
d-ronnqvist: Question: what's the reasons for defining two different `ColumnAlignment` types?

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
```diff
@@ -281,6 +283,325 @@ class RowTests: XCTestCase {
         )
     }
 
+    func testColumnWithAlignmentOnly() async throws {
```
d-ronnqvist: We aim to write new tests using Swift Testing instead of XCTest. 

One way of doing this would be to define a new suite and add the new tests there, leaving the existing tests in place. That said, there are currently only 5 tests in this file (with 9 more being added in this PR) so another option would be to update the existing tests to use Swift testing as well.

For these added tests specifically; `testAllAlignmentValues` could benefit from a parameterized test. There's also some overlap between `testAlignmentWithInvalidType`, `testAlignmentWithSizeInvalidType`, and `testMultipleColumnsWithInvalidAlignment` that could likely be combined into one—possible parameterized—test.

Sources/SwiftDocC/Semantics/Reference/Row.swift
```diff
@@ -93,6 +109,23 @@ extension Row {
...
+    ///
+    ///     @Column(alignment: center) {
+    ///         Centered content
+    ///     }
+    ///
+    ///     @Column(size: 2, alignment: trailing) {
+    ///         Trailing alignment
+    ///     }
+    /// }
+    /// ```
```
d-ronnqvist, line 112: Because this documentation comment is displayed in DocC's documentation and because this is a new feature, we like to call out what version of DocC first supported this feature. You can see a similar example (of an added directive argument) in the [`@Available` documentation](https://www.swift.org/documentation/docc/available), which corresponds to [this markup](https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/Semantics/Metadata/Availability.swift#L41-L42).

d-ronnqvist:
Regarding the "Dependencies" and "Testing" sections of the PR description; have I understood correctly that the columns will not display the _aligned_ content until there's a corresponding DocC-Render change? If that's correct; the provided testing instructions wouldn't work and we would want to list the DocC Render PR as dependency.

Sources/docc/DocCDocumentation.docc/docc.symbols.json
sofiaromorales: Thanks David!

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
VictorPuga: Fixed

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
VictorPuga: Fixed

Sources/SwiftDocC/Semantics/Reference/Row.swift
VictorPuga: Fixed

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
VictorPuga: Fixed

VictorPuga:
> Regarding the "Dependencies" and "Testing" sections of the PR description; have I understood correctly that the columns will not display the _aligned_ content until there's a corresponding DocC-Render change? If that's correct; the provided testing instructions wouldn't work and we would want to list the DocC Render PR as dependency.

I'll provide the swift-docc-render PR shortly and update the testing section. I got carried away with my fork of the renderer (which implements the alignment) and wrote the testing steps without realizing that there's no way to visualize it as-is.

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
VictorPuga: Fixed. Rewrote the tests for Swift Testing

Sources/SwiftDocC/Model/Rendering/Content/RenderBlockContent.swift
VictorPuga: Fixed

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
VictorPuga: Fixed

Sources/SwiftDocC/Semantics/Reference/Row.swift
VictorPuga: The original plan with was to have two separate types because they serve different purposes:

- `Row.Column.Alignment` parses directive arguments (`@Column(alignment: center)`), so it needs `DirectiveArgumentValueConvertible`
- `RenderBlockContent.Row.Column.Alignment` handles JSON in .doccarchive files, so it needs `Codable` with custom decoding for backward compatibility with old archives

Do you think it would be better to have a single type that conforms to both protocols?

Sources/SwiftDocC/Semantics/Reference/Row.swift
VictorPuga: Is there a way to mark the argument availability separate from the directive's? If I add `@Available` to this code snippet's metadata, then the whole page's availability changes. Also, which version should I use?

image

VictorPuga:
This is ready for review

d-ronnqvist:
Thanks. I will look at it on Monday

Sources/SwiftDocC/Semantics/Reference/Row.swift
d-ronnqvist: No, the page's `@Available` information applies to the entire symbol. Because DocC doesn't present its directives as different overloads when a new parameter is added, we instead use an aside that mentions that earlier versions don't support that parameter. This is what I was referring to above when I linked to a rendered page and its markup for a directive where we previously added a parameter:

> You can see a similar example (of an added directive argument) in the [@Available documentation](https://www.swift.org/documentation/docc/available), which corresponds to [this markup](https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/Semantics/Metadata/Availability.swift#L41-L42).

Sources/SwiftDocC/Semantics/Reference/Row.swift
```diff
@@ -93,6 +93,23 @@ extension Row {
...
+    ///     @Column {
+    ///         Default (leading) alignment
+    ///     }
+    ///
+    ///     @Column(alignment: center) {
+    ///         Centered content
+    ///     }
+    ///
+    ///     @Column(size: 2, alignment: trailing) {
+    ///         Trailing alignment
```
d-ronnqvist: Would it be worth mentioning that the size of this column (considering that the sentence before these examples mentions both `size` and `alignment`)?

I'm sure that there are shorter and better ways to phrase it, but maybe something along these lines?

```suggestion
    ///         Trailing alignment; twice as wide as either of the other columns
```

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
```diff
@@ -10,196 +10,153 @@
...
 
-import XCTest
+import Testing
 @testable public import SwiftDocC
 import Markdown
 
-class RowTests: XCTestCase {
-    func testNoColumns() async throws {
+struct RowTests {
+    @Test func noColumns() async throws {
```
d-ronnqvist: I know that this is an existing test that I asked you update to use Swift Testing. 

With that acknowledged; our idea of updating tests incrementally as we make other modifications is that a person gets to revisit the test and [also refine the test](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md?plain=1#L290-L291), for example by [using a more descriptive name](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md?plain=1#L279-L283) or by leveraging parameterized tests. If we only wanted to syntactically update these tests there are tools or scripts that would allow us to make those changes through the entire codebase. However, if we did that, there wouldn't be as clear of a signal that a test was old and could benefit from being revisited.

For this test in particular, the name doesn't really [describe the test's behavior](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md?plain=1#L279-L283). A more descriptive name could be `warnsAboutRowWithoutColumns()`.

Also, the `emptyRowBody()` test later in this file is verifying the same behavior and is an almost exact copy of this code (with the difference that the input directive has curly braces instead of no braces). Those tests would benefit from being unified into one parameterized test.

Sources/SwiftDocC/Semantics/Reference/Row.swift
VictorPuga: Fixed. I got things mixed up and didn't see that you pointed out the exact lines in the markup. 

Which version should I use? I added 6.4 for now.

Sources/SwiftDocC/Semantics/Reference/Row.swift
VictorPuga: Fixed

Tests/SwiftDocCTests/Semantics/Reference/RowTests.swift
VictorPuga: Fixed

Sources/SwiftDocC/Semantics/Reference/Row.swift
d-ronnqvist: The main branch is targeting the 6.5 release but we can still cherry pick this into the converging 6.4 release.

I think which version to use depends on the "exceptions will be considered on a case-by-case basis" portion of the "Philosophy on Taking Changes" from the [Swift 6.4 release process forum post](https://forums.swift.org/t/swift-6-4-release-process/85421) (emphasis added by me):

> ### Philosophy on Taking Changes into Swift 6.4
>
> - All language and API changes for Swift 6.4 will go through the [Swift Evolution](https://github.com/apple/swift-evolution) process. Evolution proposals should aim to be completed by the branch date in order to increase their chances of impacting the Swift 6.4 release. **Exceptions will be considered on a case-by-case basis,** particularly if they tie in with the core goal of the release.
> - Other changes (e.g., bug fixes, diagnostic improvements, SourceKit interface improvements) will be accepted based on their risk and impact.
> - Low-risk test tweaks will also be accepted late into the release branch if it aids in the qualification of the release.
> - As the release converges, the criteria for accepted changes will become increasingly restrictive.

(for context; the 6.4 release has been converging since mid March and the `release/6.4.x` branch was cut from `main` on May 4th.)

In the less formal evolution process for Swift-DocC; the discussion in [this forum thread](https://forums.swift.org/t/pitch-adding-alignment-control-to-the-column-directive/86754) would be considered a "completed proposal".

DocC doesn't have a specific core goal for this release, but we would still considered individual proposals to be cherry picked into the 6.4 release "on a case-by-case basis".

_I_ would be fine with cherry-picking this into the 6.4 release with the reason that it's a nice enhancement with very low risk; assuming that the DocC-Render counterpart is also low risk and is also cherry-picked into the 6.4 release. We'll get a second opinion from Franklin, who is the release manager for Swift-DocC, in the follow-up cherry-pick PR. If it turns out that for whatever reasons we don't cherry-pick this into the 6.4 release we can always make a tiny documentation-only PR that updated this version number to 6.5 instead.

franklinsch: Thanks for the detail, David, and thank you for making this change, Victor! 🙏  If you'd like to nominate this change for Swift 6.4, please open a PR for this change and the swift-docc-render change targeting the appropriate release branch and fill out the nomination template. Thank you!

VictorPuga: I'm happy to go either way. I have no problem targeting 6.5 unless you think it's better to have it for 6.4, just let me know what works best and I can adjust accordingly.

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test Linux

d-ronnqvist (approved):

PR #1533, opened 2026-06-02
Don't warn about DeprecationSummary when symbol is partially deprecated

d-ronnqvist:
Bug/issue #, if applicable: rdar://178260666

## Summary

This updates the warning about using a `DeprecationSummary` directive for an available symbol to _not_ raise that warning if the symbol is deprecated for _any_ of its available platforms. This matches the general behavior where a symbol is considered deprecated if it's deprecated for _any_ platform (as opposed to for _all_ platforms).

This also fixes a bug where the source location of `DeprecationSummary` didn't account for the start location of a symbol's documentation comment, resulting in the warning displaying somewhere very early in the same file (potentially very far from the markup that caused the warning—and possibly not even in any documentation comment).

## Dependencies

None.

## Testing

- Add a `DeprecationSummary` directive for a symbol that's marked as available for one platform. 
  - DocC should warn about the use of that directive for an available symbol
- Add another `@availability` annotation that marks the symbol as deprecated for some _other_ platform.
  -  DocC should no longer warn about the use of that directive, because the symbol is deprecated for _one_ of its platforms.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/DocumentationNode.swift
+++ b/Sources/SwiftDocC/Model/DocumentationNode.swift
@@ -930,7 +930,7 @@ public struct DocumentationNode {
             .union(availabilityFromDirectives.available.map(\.platform.rawValue))
             .subtracting(deprecatedDomainNames)
         
-        guard isUnconditionallyAvailable || !availableDomainNames.isEmpty || deprecatedDomainNames.isEmpty else {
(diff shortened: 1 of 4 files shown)


d-ronnqvist:
@swift-ci please test

snprajwal (approved):

d-ronnqvist:
@swift-ci please test

PR #1534, opened 2026-06-03
Fix versionless Catalyst fallback availability

sofiaromorales:
Bug/issue #, if applicable:  178528216

## Summary

1. Mac Catalyst was not applying the inheritance behaviour with iOS
when there was a default availability defined for the platform.

In the case were the default availability does not defines an
introduced version, but there's a version available we can inherit
from, use the inherited version instead of keeping the availability
as versionless.

This only applies for MacCatalyst and iPadOS as there are the only
platforms that have the inheritance mechanism from iOS.

2. Don't inherit from iOS when the symbol is not available in Catalyst

When a Catalyst symbol graph is provided for a module, that graph is the
source of truth for which symbols are available on Catalyst. Symbols
absent from the Catalyst graph should not silently inherit availability
from iOS, since the symbol isn't actually available on Catalyst.

## Dependencies

N/A

## Testing

To test manually:

1. Preview the attached catalog without the fix
2. Assert the Catalyst version is not there for the Greet symbol
3. Apply this diff
4. Assert that the iOS version gets correctly inherited by Mac Catalyst platform

[dummyCatalog.docc.zip](https://github.com/user-attachments/files/28559503/dummyCatalog.docc.zip)

## Checklist

- [X] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
+++ b/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
@@ -173,6 +173,9 @@ public class DocumentationContext {
     /// The unsuccessful links are tracked so that the context doesn't attempt to re-resolve the unsuccessful links during rendering which runs concurrently for each page.
     var externallyResolvedLinks = [ValidatedURL: TopicReferenceResolutionResult]()
     
+    /// The set of symbol graph platforms registered for each module.
+    private(set) var registeredPlatformsPerModule: [String: Set<PlatformName>] = [:]
+    
     /// A temporary structure to hold a semantic value that hasn't yet had its links resolved.
     ///
     /// These temporary values are only expected to exist while the documentation is being built. Once the documentation bundles have been fully registered and the topic graph
@@ -2030,6 +2033,7 @@ public class DocumentationContext {
             try signposter.withIntervalSignpost("Load symbols", id: signposter.makeSignpostID()) {
                 try autoreleasepool {
                     try symbolGraphLoader.loadAll()
+                    registeredPlatformsPerModule = symbolGraphLoader.platformsFoundInSymbolGraphsByModule
                 }
             }
             try shouldContinueRegistration()
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1376,7 +1376,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 }
                 
                 func addFallbackIfNeeded(named name: String) {
-                    guard information[name] == nil, !unavailableDefaultPlatformNames.contains(name) else {
+                    guard information[name]?.introduced == nil, !unavailableDefaultPlatformNames.contains(name) else {
                         return
                     }
                     var copy = iOSAvailability
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1384,6 +1384,8 @@ public struct RenderNodeTranslator: SemanticVisitor {
                     information[name] = copy
                 }
                 addFallbackIfNeeded(named: PlatformName.iPadOS.displayName)
+                
+                // FIXME: Move this logic out of the rendering code (rdar://172280267)
                 let catalystSGFExists = context.registeredPlatformsPerModule[moduleName.symbolName]?.contains(.catalyst) ?? false
                 let symbolExistsInCatalystSymbolGraph = information[PlatformName.catalyst.displayName] != nil
                 
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1384,7 +1384,16 @@ public struct RenderNodeTranslator: SemanticVisitor {
                     information[name] = copy
                 }
                 addFallbackIfNeeded(named: PlatformName.iPadOS.displayName)
-                addFallbackIfNeeded(named: PlatformName.catalyst.displayName)
+                let symbolAvailabilityContainsMacCatalyst = inSourceAvailability.availability.contains(where: {
+                    $0.domain?.rawValue == SymbolGraph.Symbol.Availability.Domain.macCatalyst
+                })
+                let catalystSGFExists = context.registeredPlatformsPerModule[moduleName.symbolName]?.contains(.catalyst) ?? false
+                
+                // Only inherit iOS if the Catalyst SGF don't define availability
+                // or if there's no catalyst SGF at all.
+                if symbolAvailabilityContainsMacCatalyst || !catalystSGFExists {
+                    addFallbackIfNeeded(named: PlatformName.catalyst.displayName)
+                }
             }
             
             // Lastly, remove any inferred or default information for platforms that were marked explicitly unavailable.
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -1384,14 +1384,14 @@ public struct RenderNodeTranslator: SemanticVisitor {
                     information[name] = copy
                 }
                 addFallbackIfNeeded(named: PlatformName.iPadOS.displayName)
-                let symbolAvailabilityContainsMacCatalyst = inSourceAvailability.availability.contains(where: {
-                    $0.domain?.rawValue == SymbolGraph.Symbol.Availability.Domain.macCatalyst
-                })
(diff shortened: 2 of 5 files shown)


Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -116,4 +156,203 @@ class SymbolAvailabilityTests: XCTestCase {
         ])
     }
     
+    func testMacCatalystInheritsIntroducedVersionFromIOSWhenDefaultAvailabilityIsVersionless() async throws {
```
d-ronnqvist: Can we add these tests in AvailabilityTests.swift instead? That's the file we're using to collect tests related to how various availability inputs render. That file is also already using Swift Testing, which we aim to use for new tests.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -17,6 +17,46 @@ import DocCCommon
...
+            ))
+        }
+        let catalog = Folder(
+            name: "unit-test.docc",
+            content: [
+                InfoPlist(defaultAvailability: [
+                    "ModuleName": defaultAvailability
+                ]),
+            ] + jsonFiles
+        )
```
d-ronnqvist: FYI: Using the result builder syntax is should be possible to express this like:


```suggestion
        let catalog = Folder(name: "unit-test.docc") {
            InfoPlist(defaultAvailability: [ "ModuleName": defaultAvailability])
           
            for (index, symbolGraphData) in symbolGraphs.indexed() {
                JSONFile(name: "ModuleName-\(index).symbols.json", content: makeSymbolGraph(
                    moduleName: "ModuleName",
                    platform: SymbolGraph.Platform(architecture: nil, vendor: nil, operatingSystem: SymbolGraph.OperatingSystem(name: symbolGraphData.operatingSystemPlatformName), environment: symbolGraphData.environmentName),
                    symbols: symbolGraphData.symbols,
                ))
            }
        }
```

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -17,6 +17,46 @@ import DocCCommon
 
 class SymbolAvailabilityTests: XCTestCase {
     
+    struct MockSymbolGraph {
```
d-ronnqvist: It feels a bit unexpected to me to call this a "mock". It isn't used as a replacement / stand-in for a symbol graph and it only represent a subset of the symbol graph's data. 

From how it's used in the tests below, this is a collection of data that's used to create the symbol graphs that are then tested. It's also only useful as a parameter to `renderNodeAvailability(...)` which is `private`, so this type could be `private` as well.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -17,6 +17,46 @@ import DocCCommon
...
+        let operatingSystemPlatformName: String
```
d-ronnqvist: minor: All the tests in this file pass `"ios"` for the platform name. We could simplify the tests (and this type) by removing this property. 

There's also only two values passed for the environment name (`"macabi"` and `nil`), so if we wanted, we could represent that with an `isCatalyst` Boolean instead.

Tests/SwiftDocCTests/Rendering/SymbolAvailabilityTests.swift
```diff
@@ -17,6 +17,46 @@ import DocCCommon
...
+            self.operatingSystemPlatformName = operatingSystemPlatformName
+            self.environmentName = environmentName
+            self.symbols = symbols
+        }
+    }
+    
+    private func renderNodeAvailability(
+        defaultAvailability: [DefaultAvailability.ModuleAvailability] = [],
+        symbolGraphs: [MockSymbolGraph],
+        symbolName: String
```
d-ronnqvist: minor: All tests pass the same value for this parameter (`"SymbolName"`)

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -173,6 +173,9 @@ public class DocumentationContext {
     /// The unsuccessful links are tracked so that the context doesn't attempt to re-resolve the unsuccessful links during rendering which runs concurrently for each page.
     var externallyResolvedLinks = [ValidatedURL: TopicReferenceResolutionResult]()
     
+    /// The set of symbol graph platforms registered for each module.
+    private(set) var registeredPlatformsPerModule: [String: Set<PlatformName>] = [:]
```
d-ronnqvist, line 177: minor: To me, storing and passing this information to the render node translator, feels like short-term workaround. Can we file an issue and add a FIXME comment to track a long-term solution?

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1376,15 +1376,24 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
                 }
                 addFallbackIfNeeded(named: PlatformName.iPadOS.displayName)
-                addFallbackIfNeeded(named: PlatformName.catalyst.displayName)
+                let symbolAvailabilityContainsMacCatalyst = inSourceAvailability.availability.contains(where: {
+                    $0.domain?.rawValue == SymbolGraph.Symbol.Availability.Domain.macCatalyst
+                })
+                let catalystSGFExists = context.registeredPlatformsPerModule[moduleName.symbolName]?.contains(.catalyst) ?? false
+                
+                // Only inherit iOS if the Catalyst SGF don't define availability
+                // or if there's no catalyst SGF at all.
```
d-ronnqvist: I don't understand this comment in relation to the code. The way I read this comment is that we should only add a fallback for Mac Catalyst if the symbol _doesn't_ have any in-source availability attributes for Mac Catalyst but the if-statement checks `symbolAvailabilityContainsMacCatalyst` which is is that the symbol _does_ have a Mac Catalyst in-source availability attribute.

d-ronnqvist (changes requested):
Please move these tests to AvailabilityTests.swift where we verify other availability behaviors.

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
sofiaromorales: I updated the variable names and code comment to better explain the inheritance logic in https://github.com/swiftlang/swift-docc/pull/1534/changes/5db91cb7ada4d4532b314b608f1c232b1be864b1

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
sofiaromorales: Added here https://github.com/swiftlang/swift-docc/pull/1534/changes/09c8160b60a26ea9c61ab797b536c768c93933e7

sofiaromorales:
@swift-ci please test

d-ronnqvist (approved):
Thanks for fixing this. This looks good to me.

d-ronnqvist:
@swift-ci please test

Tests/SwiftDocCTests/Model/AvailabilityTests.swift
```diff
@@ -350,6 +350,153 @@ struct AvailabilityTests {
         #expect(renderPlatforms.first(where: { $0.name == "Mac Catalyst" })?.introduced ==  "6.5")
     }
     
+    @Test
+    func catalystInheritsAvailabilityIfDefaulAvailabilityIsVersionless() async throws {
```
snprajwal, line 354: Nit


```suggestion
    func catalystInheritsAvailabilityIfDefaultAvailabilityIsVersionless() async throws {
```

Tests/SwiftDocCTests/Model/AvailabilityTests.swift
```diff
@@ -350,6 +350,153 @@ struct AvailabilityTests {
...
+        let renderPlatforms = try #require(renderNode.metadata.platforms)
+        #expect(renderPlatforms.compactMap(\.name) == ["iOS", "iPadOS", "Mac Catalyst"])
+        
+        #expect(renderPlatforms.first(where: { $0.name == "iOS"          })?.introduced == "12.0")
+        #expect(renderPlatforms.first(where: { $0.name == "iPadOS"       })?.introduced == "12.0")
+        #expect(renderPlatforms.first(where: { $0.name == "Mac Catalyst" })?.introduced ==  "12.0")
+    }
+    
+    @Test
+    func catalystDoesNotInheritsAvailabilityIfItDoesntExistsInCatalystSGF() async throws {
```
snprajwal, line 477: Nit


```suggestion
    func catalystDoesNotInheritAvailabilityIfItDoesntExistInCatalystSGF() async throws {
```

snprajwal (approved):
The changes LGTM. Since we squash and merge PRs in DocC, it would be good to update the PR title to be more descriptive, as that gets picked up for the commit title when we merge this PR. Thanks for working on this!

PR #1535, opened 2026-06-03
 Sort symbol graphs lexicographically for path hierarchy

snprajwal:
Bug/issue #, if applicable: rdar://170241366

## Summary

The symbol graph loader reads symbol graphs in parallel into a dictionary that maps from URL to SymbolGraph. The ordering of this dictionary is unstable, and the comparator used to sort this dictionary only performs a partitioned sort to place main symbol graphs ahead of extensions. It does not sort within these partitions to order the platforms and languages, leading to non-deterministic behaviour when building documentation for targets containing symbol identifiers that vary across platforms or languages.

This patch updates the symbol node construction to use the default symbol in the unified symbol graph, ensuring that symbol information like path components and the symbol name originate from the same symbol as the one currently used for the page title. The cascading sort from the previous commit is used for platform ordering hierarchy when identifying the default symbol. Other code paths that use this hierarchy have been updated to use the comparator introduced in the `PlatformName` type.

Previously, platforms were being sorted lexicographically in the `AvailabilitySortOrder.compare` method. This was not consistent with the platform ordering that is used by `PlatformName`. This patch also appropriately renames the static comparator method and moves it into `PlatformName`, and updates the logic to use the platform ordering to sort known platforms, and lexicographically sort unknown platforms. Code paths that use a similar duplicate comparator have been reconciled to use this comparator instead.

## Dependencies

N/A

## Testing

1. Create a catalog with two symbol graphs for different platforms, both containing a symbol with a different name in each.
2. `xcrun docc convert <your-catalog>` a few times and view the emitted render JSON file for each run.

Previously, across multiple runs, the filename and contents would change based on which symbol graph was processed first. With this patch, it remains consistent.

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -3366,6 +3366,47 @@ class PathHierarchyTests: XCTestCase {
         XCTAssertEqual(paths[defaultImplementationID], singlePlatformPaths[defaultImplementationID])
     }
     
+    func testSymbolWithDifferentNamesPerPlatformResolvesToDefaultSymbol() async throws {
```
d-ronnqvist: FYI: I've started writing new Swift Testing tests in `PathHierarchyTests_new` at the top of this file.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -136,7 +146,17 @@ struct PathHierarchy {
                 } else {
                     assert(!symbol.pathComponents.isEmpty, "A symbol should have at least its own name in its path components.")
 
-                    let node = Node(symbol: symbol, name: symbol.pathComponents.last!)
+                    // Use the default symbol to construct the node so that the symbol info doesn't depend on the order in which the symbol graphs were processed.
+                    let nodeSymbol: SymbolGraph.Symbol
+                    if let defaultSymbol = defaultSymbolsByUSR[id],
+                       defaultSymbol.identifier.interfaceLanguage == symbol.identifier.interfaceLanguage
+                    {
+                        nodeSymbol = defaultSymbol
```
d-ronnqvist: I need some time to consider the implications of this.

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
snprajwal: Missed that, thanks. Ported it over.

snprajwal:
@swift-ci please test

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,48 @@ struct PathHierarchyTests_new {
...
+        let iOSFirst   = Folder(name: "unit-test.docc", content: [iOSSymbolGraph, macOSSymbolGraph])
+        let macOSFirst = Folder(name: "unit-test.docc", content: [macOSSymbolGraph, iOSSymbolGraph])
+
+        for catalog in [iOSFirst, macOSFirst] {
+            let context = try await load(catalog: catalog)
+            let tree = context.linkResolver.localResolver.pathHierarchy
+            let paths = tree.caseInsensitiveDisambiguatedPaths()
+            // iOS has higher priority compared to macOS, and should be used for the URL
+            #expect(paths[symbolID] == "/ModuleName/SomeClass/iOSType")
+        }
```
d-ronnqvist: I don't think the order of the files actually makes a difference for two reasons:
- `SymbolGraphLoader.loadAll()` loads and decodes the symbol graph data in parallel, so the order that the context input lists them (which is based on the unstable order that the file system lists the files) doesn't necessarily impact the order that they are added to `loadedGraphs`.
- `SymbolGraphLoader.symbolGraphs` is a dictionary which doesn't have a stable order across different program executions so even if the symbol graphs were added in a deterministic order, the order that the `PathHierarchy` iterates over them wouldn't be.

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,48 @@ struct PathHierarchyTests_new {
...
+        // Use two catalogs with different ordering of symbol graphs to verify that the page URL
+        // for the symbol does not depend on the order in which the symbol graphs are processed.
+        let iOSFirst   = Folder(name: "unit-test.docc", content: [iOSSymbolGraph, macOSSymbolGraph])
+        let macOSFirst = Folder(name: "unit-test.docc", content: [macOSSymbolGraph, iOSSymbolGraph])
+
+        for catalog in [iOSFirst, macOSFirst] {
+            let context = try await load(catalog: catalog)
+            let tree = context.linkResolver.localResolver.pathHierarchy
+            let paths = tree.caseInsensitiveDisambiguatedPaths()
+            // iOS has higher priority compared to macOS, and should be used for the URL
```
d-ronnqvist: Question: where is order defined? To me this doesn't feel like it's knowledge / responsibility that belongs in the path hierarchy.

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
snprajwal: Both points are true. However, if we remove the second ordering (or make it a duplicate of the first ordering), it then requires multiple test invocations to have it fail without to the fix added in `PathHierarchy.swift`. You can try this with `git checkout HEAD~1` on this branch and running this test.

I found that by adding a second catalog with a different input order, I was able to get the test to fail much more frequently. Still not perfect, but it made it easier for me to verify my changes as I added them.

I can see how this can look misleading, since it might make the reader think that the ordering is used in the build process. I can update the test with a comment to include the info you shared above, so that it's obvious that the ordering doesn't affect the inputs used for the build. Or do you have a better strategy to test this? It's quite annoying since the behaviour is non-deterministic, so a single build the way most of our tests are written wouldn't reliably test it on every run.

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
snprajwal: This is defined in `PlatformName.sortedPlatforms`:

https://github.com/swiftlang/swift-docc/blob/d1a8c1656cafc2b9bb51599db61f68a6ede0f0ba/Sources/SwiftDocC/Semantics/Symbol/PlatformName.swift#L74-L84

This was already being used for the title, which is why I retained it to use for these changes too.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
d-ronnqvist: Ok. I've gotten to sleep on this and my primary question is what _specific_ problem this is aiming to solve.

If the problem is that this setup (which don't know how it can be represented in a project) experiences non-deterministic behavior in the path, then I feel that a much simpler solution to that non-determinism would be to sort the symbol graphs alphabetically by their path (while still sorting extension symbol graph files later). This would move the complexity out of the main hierarchy initialization flow.

If on the other hand the problem is that the path doesn't always match the symbol's title, then I feel that that should be solved outside of the path hierarchy. For example, we could change the path hierarchy to accept a `some Sequence<(URL, SymbolGraph)>` parameter instead `SymbolGraphLoader` parameter, making it the caller's responsibility to sort the symbol graphs. Alternatively, we could change `SymbolGraphLoader.symbolGraphs` to return a sorted sequence of `(URL, SymbolGraph)` and make the loader responsible for the sorting.

d-ronnqvist: Even though the path hierarchy is defined in the main SwiftDocC module it is rather isolated from the rest and I sometimes think of it as _if_ it was a separate module. Perhaps some day I will extract it into a separate module for real. Until then, I'd like to be careful about the boundary between it and the rest of the SwiftDocC module.

d-ronnqvist (changes requested):
I'm not sure I really understand the specific problem that this is solving but given what I understand so far I don't feel that this is the right solution to my understanding of that problem. 

Also, I don't know how I can test this myself. Specifically; can you elaborate on how I can accomplish this step without manually modifying the symbol graph json files (in potentially semantically invalid ways)?

> 1. Create a catalog with two symbol graphs for different platforms, both containing a symbol with a different name in each.

If I try to `#if`-define two symbols with different names for different platforms—in either Swift or C/Objective-C—those symbols also get different unique identifiers, so they don't reproduce this issue. Instead, that setup simply defines two _different_ platforms specific symbols where DocC creates separate pages for each symbol.

snprajwal:
This was discussed offline, a general summary of what was clarified:

- There are a number of cases that are fixed by this PR:
    - If a symbol is defined for one platform in Objective-C, and Swift symbols are synthesized for it via API notes, it can happen that the symbol is synthesized for a platform where its container is not available. In this case, the symbol has the same USR but different path components for different platforms. Currently, the URL is generated based on the first ingested symbol graph, which is not deterministic due to the symbol graphs in the loader being unordered. This case is technically an issue with the source code, but DocC should still behave deterministically with these inputs.
    - For the above case, if a documentation extension is used to provide the title and contents of the page, then the title can be different from the definition depending on which platform's definition was processed first.
    - If a C method has varying parameters for different platforms, it would still have the same USR. While the URL remains consistent in this case, the ordering of the different definitions in the page vary based on which platform was processed first.

- The path hierarchy is not responsible for determining the default symbol in the unified symbol graph and using that to register the symbol that corresponds to the one used for the page title. The simplest and least invasive solution to this problem that we've agreed upon is to use a simple alphabetical sort to maintain determinism in the path hierarchy logic. This way, if the user wants to control which symbol graphs are processed first, they can do so by renaming the files rather than having to modify the sorting logic if a more complex cascading sort were used (like the sort proposed in #1462).
- Due to the non-deterministic nature of the problem, there is no good way to test this reliably. An acceptable compromise is to use the existing test, and add a note stating that it tests non-deterministic behaviour and might need to be run multiple times to verify that the behaviour is correct.

snprajwal:
Note: dd447b4 is not related to this updated fix, but does fix a sort inconsistency. I'm happy to leave it in or split it into a separate PR.

snprajwal:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -79,7 +79,12 @@ struct PathHierarchy {
                 (url: url, graph: graph, language: graph.symbols.values.mapFirst(where: { SourceLanguage(id: $0.identifier.interfaceLanguage) }))
             }
             .sorted(by: { lhs, rhs in
-                return !lhs.url.lastPathComponent.contains("@")
+                let lhsIsExt = lhs.url.lastPathComponent.contains("@")
+                let rhsIsExt = rhs.url.lastPathComponent.contains("@")
+                // Sort extension files after main symbol graph files
+                if lhsIsExt != rhsIsExt { return !lhsIsExt }
+                // Sort lexicographically between symbol graph files of the same type
+                return lhs.url.lastPathComponent < rhs.url.lastPathComponent
```
d-ronnqvist: If we want the order to be fully stable I think that we need to sort by the full path. Otherwise it's possible that two symbol graph files in different subdirectories will have the same last path component—in which case I think that the sorted order would depend on the initial unordered order.

---

Here's a very silly example that tries to highlight what I'm talking about:
```swift
var urls = "ABCDEFGHIJKLMNOPQRSTUVXYZ".enumerated().map { offset, letter in
    URL(fileURLWithPath: "/\(letter)/\(offset.isMultiple(of: 2) ? "Even" : "Odd")")
}
// "/A/Even", "/B/Odd", "/C/Even", "/D/Odd", "/E/Even", "F/Odd", ...

urls.shuffle()

let byLast = urls.sorted(by: { $0.lastPathComponent < $1.lastPathComponent })
let byFull = urls.sorted(by: { $0.path < $1.path })
#expect(byLast == byFull) // This expectation fails

for _ in 1 ... 10 {
    urls.shuffle()
    
    let again = urls.sorted(by: { $0.lastPathComponent < $1.lastPathComponent })
    #expect(again == byLast) // This expectation also fails
}
```

Sources/SwiftDocC/Semantics/Symbol/PlatformName.swift
```diff
@@ -122,4 +135,27 @@ public struct PlatformName: Codable, Hashable, Equatable {
...
+        let rhsIndex = platformOrderIndex[rhs.lowercased()] ?? Int.max
+        if lhsIndex != rhsIndex { return lhsIndex < rhsIndex }
+        return lhs < rhs
+    }
+
+    /// An overload of ``isInOrder(_:_:)`` that accepts optional values. Nil values are sorted first.
+    static func isInOrder(_ lhs: String?, _ rhs: String?) -> Bool {
+        switch (lhs, rhs) {
+        case (nil, nil):
+            return false
```
d-ronnqvist: FYI: this case would be covered by `(_, nil)` (below) if `(_, nil)` was moved up to the top.

This can also be written more concisely by omitting the `return` keywords and moving the return values onto the same line as the cases
```swift
switch (lhs, rhs) {
    case (_,   nil): false
    case (nil, _  ): true
    case (let lhs?, let rhs?): isInOrder(lhs, rhs)
}
```

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,47 @@ struct PathHierarchyTests_new {
...
+                    makeSymbol(id: symbolID, kind: .enum, pathComponents: ["SomeMacOSType"]),
+                ]
+            ))
+        ])
+
+        let context = try await load(catalog: catalog)
+        let tree = context.linkResolver.localResolver.pathHierarchy
+        let paths = tree.caseInsensitiveDisambiguatedPaths()
+        // iOS sorts ahead of macOS lexicographically, so
+        // the path components of the iOS type should be used.
```
d-ronnqvist: minor: I feel like this would be more obvious (to the point where we wouldn't need this comment) if the symbol graph files were named "ModuleName-A" and "ModuleName-B" instead.

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,47 @@ struct PathHierarchyTests_new {
...
+            JSONFile(name: "ModuleName-ios.symbols.json", content: makeSymbolGraph(
+                moduleName: "ModuleName",
+                platform: .init(operatingSystem: .init(name: "ios")),
+                symbols: [
+                    makeSymbol(id: containerID, kind: .class, pathComponents: ["SomeClass"]),
+                    makeSymbol(id: symbolID, kind: .enum, pathComponents: ["SomeClass", "iOSType"]),
+                ],
+                relationships: [
+                    .init(source: symbolID, target: containerID, kind: .memberOf, targetFallback: nil),
+                ]
```
d-ronnqvist, line 168: minor: As far as understand; it's not necessary to have to symbols with a relationship to reproduce this issue (before theses changes)

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,47 @@ struct PathHierarchyTests_new {
...
         #expect(links["some-symbol-id"] == "/ModuleName/\(symbolName)", "Links allow any special characters")
     }
+
+    // Verifies the fix for rdar://170241366.
+    // This test verifies determinism in a previously non-deterministic code path
+    // that generates the URL of a symbol's page. If you suspect that your changes
+    // might affect this test, it would require multiple runs to verify that
+    // the behaviour has not changed undesirably.
+    @Test
+    func testSymbolWithDifferentNamesPerPlatformResolvesToDefaultSymbol() async throws {
```
d-ronnqvist: minor: This title doesn't match the actual behavior; the logic doesn't know about the default symbol.

Also, we don't use a `test` prefix for our Swift Testing tests.

```suggestion
    func symbolWithPlatformSpecificNamesHasDeterministicPath() async throws {
```

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,47 @@ struct PathHierarchyTests_new {
         #expect(paths["some-symbol-id"] == "/ModuleName/Symbol_Name__", "Each special character should be replaced with a '_'")
         #expect(links["some-symbol-id"] == "/ModuleName/\(symbolName)", "Links allow any special characters")
     }
+
+    // Verifies the fix for rdar://170241366.
+    // This test verifies determinism in a previously non-deterministic code path
+    // that generates the URL of a symbol's page. If you suspect that your changes
+    // might affect this test, it would require multiple runs to verify that
+    // the behaviour has not changed undesirably.
```
d-ronnqvist: nit: I feel that the downsides for wrapping code comments to any arbitrary character limit outweigh the upsides. Minor edits that on line will cascade throughout all remaining lines. I much prefer to wrap code comments at semantic locations (like periods or commas):

Also, we generally don't mention which issue a certain test verifies. If someone wants to know that they can find that information in the git history.

```suggestion
    // This verifies determinism in a previously non-deterministic behavior that cannot be reproduced reliably in a test.
    // If you suspect that your changes might affect this test, 
    // you need to run it repeatedly---and relaunch for each repetition---to verify that the behavior remains deterministic.
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
d-ronnqvist: The current test has 2 files with different last path components. Do you think that it would be worth to also verify the determinism when the files are named the same but exist in different sub directories?

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -152,6 +152,47 @@ struct PathHierarchyTests_new {
...
+                    makeSymbol(id: symbolID, kind: .enum, pathComponents: ["SomeClass", "iOSType"]),
+                ],
+                relationships: [
+                    .init(source: symbolID, target: containerID, kind: .memberOf, targetFallback: nil),
+                ]
+            )),
+            // A macOS symbol graph with the same enum as a top-level symbol
+            JSONFile(name: "ModuleName-macos.symbols.json", content: makeSymbolGraph(
+                moduleName: "ModuleName",
+                platform: .init(operatingSystem: .init(name: "macos")),
```
d-ronnqvist: nit: If I understand correctly, this platform is not needed to reproduce this issue (before the fix)
```suggestion
```

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
snprajwal: That's a good shout, I hadn't considered that case. Will update.

snprajwal:
@swift-ci please test

d-ronnqvist (approved):

PR #1537, opened 2026-06-05
Use GitHub Actions for CI workflows

snprajwal:
This patch introduces configuration for GitHub Actions, and adds a number of CI jobs that trigger on pull requests. Most of the information related to the workflow comes from the org-wide repository at https://github.com/swiftlang/github-workflows.

This is a subset of the changes proposed in #1414:

- The unacceptable language check has been disabled since it requires more discussion, and the workaround to make the check ignore words breaks the build.
- The trailing whitespace check has been omitted since it is unrelated to GHA support, and requires more discussion.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
--- a/.github/scripts/prebuild.ps1
+++ b/.github/scripts/prebuild.ps1
@@ -1,0 +1,38 @@
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2026 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
+# See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+#
+
+param (
+    [switch]$SkipAndroid,
+    [switch]$InstallCMake
+)
+
+# winget isn't easily made available in containers, so use chocolatey
+Set-ExecutionPolicy Bypass -Scope Process -Force; [System.Net.ServicePointManager]::SecurityProtocol = [System.Net.ServicePointManager]::SecurityProtocol -bor 3072; iex ((New-Object System.Net.WebClient).DownloadString('https://community.chocolatey.org/install.ps1'))
+
+if ($InstallCMake) {
+    choco install -y cmake --installargs 'ADD_CMAKE_TO_PATH=System' --apply-install-arguments-to-dependencies
+    choco install -y ninja
+
+    Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+    refreshenv
+
+    # Let swiftc find the path to link.exe in the CMake smoke test
+    $env:Path += ";$(Split-Path -Path "$(& "${env:ProgramFiles(x86)}\Microsoft Visual Studio\Installer\vswhere.exe" "-latest" -products Microsoft.VisualStudio.Product.BuildTools -find VC\Tools\MSVC\*\bin\HostX64\x64\link.exe)" -Parent)"
+}
+
+if (-not $SkipAndroid) {
+    choco install -y android-ndk
+
+    Import-Module $env:ChocolateyInstall\helpers\chocolateyProfile.psm1
+    refreshenv
+
+    # Work around a bug in the package causing the env var to be set incorrectly
+    $env:ANDROID_NDK_ROOT = $env:ANDROID_NDK_ROOT.replace('-windows.zip','')
+}
--- a/.github/scripts/prebuild.sh
+++ b/.github/scripts/prebuild.sh
@@ -1,0 +1,83 @@
+#!/bin/bash
+#
+# This source file is part of the Swift.org open source project
+#
+# Copyright (c) 2026 Apple Inc. and the Swift project authors
+# Licensed under Apache License v2.0 with Runtime Library Exception
+#
+# See https://swift.org/LICENSE.txt for license information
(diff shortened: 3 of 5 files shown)


snprajwal:
We would also have to disable the format check, since this repository does not require running swift-format on patches, and there are a number of places where we do not want the default swift-format behaviour in order to keep it more readable.

snprajwal:
- [macOS CI failures](https://github.com/swiftlang/swift-docc/actions/runs/27023821598/job/79758535005?pr=1537#step:7:7) are due to Xcode not being present on the runners. Only macOS + Xcode 16.4 is correctly configured
- [Linux CI failures](https://github.com/swiftlang/swift-docc/actions/runs/27023821598/job/79758534926?pr=1537#step:14:5225) are due to [NavigatorIndexTests.testNavigatorIndexAsReadOnlyFile](https://github.com/swiftlang/swift-docc/blob/main/Tests/SwiftDocCTests/Indexing/NavigatorIndexTests.swift#L1830) failing

snprajwal:
A number of the Windows CI failures are due to filename issues, potentially related to what #668 fixes

snprajwal:
I have additionally disabled:

- Docs check: Requires fixes in source files, I'll do that in a follow-up PR
- License header check: Requires fixes in source files, and is a source of noise due to the large number of changes. I'll do that in a follow-up PR
- Python lint check: I've updated the formatting in the single Python file we have to be correctly formatted, but the check cannot be enabled due to mutually conflicting checks [W503](https://www.flake8rules.com/rules/W503.html) and [W504](https://www.flake8rules.com/rules/W504.html) both being enabled. This is not usually an issue due to them being in the default ignore list for flake8, but we override it in the flake8 config. https://github.com/swiftlang/github-workflows/pull/279 fixes it, and we can re-enable it after that lands

snprajwal:
We now have Linux and macOS CI working a la the existing Swift CI workflow. @d-ronnqvist should we defer the Windows support to a different PR? I also see there's been a separate attempt at adding that in #1115.

d-ronnqvist:
> We now have Linux and macOS CI working a la the existing Swift CI workflow. @d-ronnqvist should we defer the Windows support to a different PR? I also see there's been a separate attempt at adding that in #1115.

That sounds like a good idea to me to achieve incremental benefits rather than blocking everything on the Windows CI.

snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
d-ronnqvist, line 1: Just so that I understand; this change to the bug report template is unrelated to GitHub actions, correct?

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
d-ronnqvist, line 1: Just so that I understand; this change to the feature request template is unrelated to GitHub actions, correct?

build-script-helper.py
d-ronnqvist, line 1: This looks like only formatting differences. Are they necessary or can we exclude this change?

.github/ISSUE_TEMPLATE/BUG_REPORT.yml
snprajwal: This is related to the YAML formatting check that's part of the soundness checks. This file was previously unformatted, and the CI check would fail without this update.

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
snprajwal: This is related to the YAML formatting check that's part of the soundness checks. This file was previously unformatted, and the CI check would fail without this update.

build-script-helper.py
snprajwal: This is related to the Python formatting check that's part of the soundness checks. This file was previously unformatted, and the CI check would fail without this update.

The Python formatting check is currently disabled due to [a bug in the configuration](https://github.com/swiftlang/github-workflows/pull/279), but I thought it would be good for us to get the formatting fix in so that we can simply enable the check once the fix lands. If you prefer, I can remove this now and include it when we enable the Python formatter.

d-ronnqvist: Yes. That would be good to cut the PR size in half.

.github/ISSUE_TEMPLATE/FEATURE_REQUEST.yml
d-ronnqvist: I tried 3 different YAML validators (the top 3 search results for "yaml validator") and neither of them reported any issues about either of these two files. If this specific CI check is too strict (warning about seemingly valid YAML), would it make sense to defer it to a later PR and reformat the YAML files then?

snprajwal: [The check](https://github.com/swiftlang/github-workflows/blob/c8aaa3d93d0750fc30d5e965dbdec5c21e6751dd/.github/workflows/soundness.yml#L417) uses [yamllint](https://yamllint.readthedocs.io/en/stable/). The YAML was structurally valid but unformatted, which is why the CI job was failing. I don't think it is too strict as [the configuration](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/configs/yamllint.yml) is pretty barebones. I don't have a strong preference, I'll move it to a separate PR so it's more obvious why we're making the change.

snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test

d-ronnqvist (approved):
Thanks for working through all these changes so that we can land this functionality in small incremental pieces. 

I'm not an expert in bash scripts—and I know almost nothing about PowerShell scripts—but because those scripts came from [Mishal's PR](https://github.com/swiftlang/swift-docc/pull/1414) originally I trust them to both be correct and to likely already be used in other repositories.

PR #1543, opened 2026-06-11
Improve handling of lists when exporting markdown

jrturton:
Bug/issue #, if applicable: 

179135030

## Summary

- Links in list items should get resolved to title + link (179135030)

Ordered and unordered lists were getting directly output using format() unless we were processing a linked list (like a topics section). This bypassed the link resolution / formatting steps that happen when visiting a link. 

Changed the processing of lists to find and convert any links or symbol links found in the list item. 

## Dependencies

N/A

## Testing

Create source documentation with article and symbol links inside ordered or unordered lists, and within standard paragraphs of prose. Run `convert` on the documentation with the `--enable-experimental-markdown-output` flag enabled. Examine the markdown produced in the docc archive. The links should be formatted the same in the list output as they are in the normal paragraph output. 

## Checklist

- [x] Added tests


d-ronnqvist (changes requested):
Hi, it appears that this is addressing two issues that aren't related to each other (and that seemingly doesn't modify the same lines of code). Can you please open one PR per issue so that it' easier to review the changes?

d-ronnqvist:
Regarding the "Testing" section of the PR description; 
This section is meant to contain instructions describing how a reviewer can manually try out and verify the changes. It is expected that all PRs add automated tests to cover their changes. Please fill out the PR template (testing excerpt below) in both PRs.

> Describe how a reviewer can test the functionality of your PR. Provide test content to test with if applicable.
>
>Steps:
>
> - Provide setup instructions.
> - Explain in detail how the functionality can be tested.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
d-ronnqvist, line 1: Please use Swift Testing for all new tests. See the ["Adding new tests" section](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md#adding-new-tests) of the contributing document for more details.

jrturton: I was planning to migrate all the tests to swift testing as a subsequent PR to this (now these?), to reduce noise - is that OK or would you prefer me to add another file with just the new tests in it?

d-ronnqvist: The contributing document offers some guidance on this; specifically that you can create a new test suite in (even in the same file) that contains only the Swift Testing test(s):

> #### Updating existing tests
>
>If you're updating an existing test case with additional logic, we appreciate if you also modernize that test while updating it, but we don't expect it. If the test case is part of a large file, you can create new test suite which contains just the test case that you're modernizing.
>
>If you modernize an existing test case, consider not only the syntactical differences between Swift Testing and XCTest, but also if there are any Swift Testing features or other changes that would make the test case easier to read, maintain, or debug.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+    func testNestedLists() async throws {
+        let catalog = catalog(files: [
+            TextFile(name: "NestedLists.md", utf8Content: """
+                # Nested Lists
+
+                Tests the appearance of nested lists
+
+                ## Overview
+
+                This is a list:
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion

```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+                - This is a top-level list item
+                  - This is a nested list item
+                  - This is another nested list item
+                - This is back to the top-level
+                
+                ## Topics
+
+                ### No more links
+                
+                Empty section
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -146,6 +240,11 @@ final class MarkdownOutputTests: XCTestCase {
                 This is an inline link: ``MarkdownSymbol``
                 
                 This is an unresolvable link: ``Unresolvable``
+                
+                This is a list of things that have links:
+                
+                - You can use ``MarkdownSymbol`` to do interesting things
+                - You can't use other things
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+                
+                1. This is an inline link: <doc:RowsAndColumns>
+                    1. This is a nested inline link with a heading: <doc:RowsAndColumns#Overview>
+                    2. Here is it again <doc:RowsAndColumns#Overview>
+                2. This is an inline link with a multi-word heading: <doc:RowsAndColumns#Multi-word-heading>
+                ## Topics
+
+                ### No more links
+                
+                Empty section
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+                """),
+            TextFile(name: "Links.md", utf8Content: """
+                # Links
+
+                Tests the appearance of inline and linked lists
+
+                ## Overview
+
+                These are some interesting links that are in list items: 
+
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+                ## Overview
+
+                These are some interesting links that are in list items: 
+                
+                - This is an inline link: <doc:RowsAndColumns>
+                  - This is a nested inline link with a heading: <doc:RowsAndColumns#Overview>
+                - This is an inline link with a multi-word heading: <doc:RowsAndColumns#Multi-word-heading>
+
+                These are some interesting links that are in ordered list items: 
+
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -133,7 +133,101 @@ final class MarkdownOutputTests: XCTestCase {
...
+                
+                Just here for the links
+                
+                ## Overview
+                
+                I am the overview
+                
+                ## Multi-word heading
+                
+                I am here to test the url readable fragment stuff
```
d-ronnqvist: minor: This markup is never verified in the output and can be removed.
```suggestion
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -17,6 +17,7 @@ import Foundation
 internal struct MarkdownOutputMarkupWalker: MarkupWalker {
     let context: DocumentationContext
     let identifier: ResolvedTopicReference
+    let formatOptions = MarkupFormatter.Options(unorderedListMarker: .dash, orderedListNumerals: .incrementing(start: 1))
```
d-ronnqvist: minor: This seems like an implementation details that callers aren't expected to inspect.
```suggestion
    private let formatOptions = MarkupFormatter.Options(unorderedListMarker: .dash, orderedListNumerals: .incrementing(start: 1))
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -93,36 +94,79 @@ extension MarkdownOutputMarkupWalker {
...
+                startNewParagraphIfRequired()
             }
+        } else {
+            let converted = convertList(unorderedList)
             startNewParagraphIfRequired()
+            defaultVisit(converted)
         }
     }
     
+    mutating func convertList<List: ListItemContainer>(_ list: List) -> List {
```
d-ronnqvist: minor: As far as I can tell, this is an implementation details that other types aren't expected to call.

Also, AFAICT this method doesn't mutate the walker.


```suggestion
    private func convertList<List: ListItemContainer>(_ list: List) -> List {
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -93,36 +94,79 @@ extension MarkdownOutputMarkupWalker {
...
+        var newItems: [ListItem] = []
+        for item in list.listItems {
+            newItems.append(convertListItem(item))
+        }
+        return List(newItems)
```
d-ronnqvist: minor: This is just a slower version of `map`—slow because it doesn't preemptively reserve space for the known number of elements.
```suggestion
        List(list.items.map { convertListItem($0) })
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -135,19 +179,19 @@ extension MarkdownOutputMarkupWalker {
...
     }
        
     mutating func visitCodeBlock(_ codeBlock: CodeBlock) {
         startNewParagraphIfRequired()
-        markdown.append(codeBlock.detachedFromParent.format())
+        markdown.append(codeBlock.detachedFromParent.format(options: formatOptions))
     }
     
-    mutating func visitSymbolLink(_ symbolLink: SymbolLink) {
+    mutating func convertSymbolLink(_ symbolLink: SymbolLink) -> (link: any InlineMarkup, abstract: (any Markup)?)? {
```
d-ronnqvist: minor: As far as I can tell, this is an implementation details that other types aren't expected to call.

Also, AFAICT this method doesn't mutate the walker.

```suggestion
    private func convertSymbolLink(_ symbolLink: SymbolLink) -> (link: any InlineMarkup, abstract: (any Markup)?)? {
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -180,21 +224,28 @@ extension MarkdownOutputMarkupWalker {
...
         withRenderingLinkList(value: false) {
             $0.visit(link)
-            $0.visit(linkListAbstract)
+            $0.visit(abstract)
         }
     }
     
-    mutating func visitLink(_ link: Link) {
-                
+    mutating func convertLink(_ link: Link) -> (link: Link, abstract: (any Markup)?) {
```
d-ronnqvist: minor: As far as I can tell, this is an implementation details that other types aren't expected to call.

Also, AFAICT this method doesn't mutate the walker.
```suggestion
    private func convertLink(_ link: Link) -> (link: Link, abstract: (any Markup)?) {
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -93,36 +94,79 @@ extension MarkdownOutputMarkupWalker {
...
     
+    mutating func convertList<List: ListItemContainer>(_ list: List) -> List {
+        var newItems: [ListItem] = []
+        for item in list.listItems {
+            newItems.append(convertListItem(item))
+        }
+        return List(newItems)
+    }
+    
+    mutating func convertListItem(_ item: ListItem) -> ListItem {
```
d-ronnqvist: minor: As far as I can tell, this is an implementation details that other types aren't expected to call.

Also, AFAICT this method doesn't mutate the walker.

```suggestion
    private func convertListItem(_ item: ListItem) -> ListItem {
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
d-ronnqvist: This method could also benefit from a brief comment explaining what it does and why. From the signature (and implementation) all I can tell is that it converts a type list to the same type of list.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
d-ronnqvist: This method could also benefit from a brief comment explaining what it does and why. From the signature all I can tell is that it converts a list item to the same type of list item.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -93,36 +94,79 @@ extension MarkdownOutputMarkupWalker {
...
+                        newComponents.append(converted)
+                    } else if let symbolLink = inlineChild as? SymbolLink {
+                        if let (converted, _) = convertSymbolLink(symbolLink) {
+                            newComponents.append(converted)
+                        }
+                    } else {
+                        newComponents.append(inlineChild)
+                    }
+                }
+                newChildren.append(Paragraph(newComponents))
```
d-ronnqvist: minor: It might make this implementation easier to read if this was extracted to a private function (or even an inner function. That would make it clearer at a glance that the outer loop is switching over the element type and calling different convert functions for each type.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -93,36 +94,79 @@ extension MarkdownOutputMarkupWalker {
...
+            } else {
+                newChildren.append(child)
+            }
+        }
```
d-ronnqvist: There are a couple patterns hiding in this implementation that I feel make it harder to read than necessary. 

Both loops behave like `map` (but is slower because it doesn't preemptively reserve capacity for the known number of elements) and both if-else expressions behave like a switch over a value. 

It might be easier to read this implementation if it leveraged those language constructs.

d-ronnqvist (changes requested):

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
jrturton: This is required as far as I can tell, because we want the heading content to be reflected in the link titles that are generated from the subsequent article. The paragraph text could maybe be removed but I feel it adds a certain amount of readability to the test rather than having a series of headings

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
jrturton: it mutates the walker by updating the outgoing links that form part of the manifest, though this is nested a few levels in. I've added some comments for reasoning.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
jrturton: I can't use map because convertListItem is a mutating function.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
jrturton: See other comments about mutating. I have updated it to be private

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
jrturton: See other comments about mutating. I have updated it to be private.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
jrturton: Can't use map, as mentioned above, updated to switch

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
d-ronnqvist: If that's the case; please update the paragraph text to explain that these headings exist so that other pages can link to them in this test.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
d-ronnqvist: I'm surprised to hear that the various `convert...` methods are mutating since their code reads as not performing any mutations and since they return the transformed result. I had thought that the reason for extracting that code was to move some work into non-mutating functions. 

Can you explain what mutations is happening in those new `convert...` methods?

jrturton: Eventually they may get down to an article or symbol link, and then that calls `add(source: ResolvedTopicReference, type: MarkdownOutputManifest.RelationshipType, subtype: RelationshipsGroup.Kind?)`, which mutates the walker by adding to the `outgoingReferences` set. The reason the code was extracted was to avoid duplication of functionality from `visitLink`.

I could refactor the convert methods to return a set of `MarkdownOutputManifest.Relationship` values or have an inout argument for the building set, if you prefer?

jrturton: Thinking on this an `inout` parameter makes more sense, we can then pass in the `outgoingReferences` set from the visit methods. Will update and convert to maps 👍

d-ronnqvist (approved):

franklinsch:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

PR #1553, opened 2026-06-16
Add diagnostic for miscased section headings

snprajwal:
Bug/issue #, if applicable: rdar://179824051

## Summary

Section headings in DocC are matched exactly in order to be treated in a special way, e.g. "See Also", "Topics", "Overview", "Discussion". If these headings are present but use an incorrect casing, then they are rendered as a normal section, and, in some cases, an autogenerated section is also presented alongside. This can lead to confusing output, and requires thorough checking in order to identify the casing error. This patch adds a diagnostic that warns the user if any of these sections are present but use a different casing than the one DocC expects.

## Dependencies

N/A

## Testing

1. Create an article with a miscased section like `## See also`.
2. Build documentation
3. A warning should be emitted stating that the "See Also" section is miscased.

## Checklist

- [x] Added tests


snprajwal:
@swift-ci please test

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
```diff
@@ -0,0 +1,73 @@
...
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+public import Foundation
+public import Markdown
+
+/// A checker that warns when a level-2 section heading differs only in casing from a recognized DocC section heading.
+public struct SectionHeadingCaseChecker: Checker {
```
d-ronnqvist: minor: I don't think this checker (or any other checker) needs to be public because there's no way for a caller outside the repo to configure which checkers the context runs.

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
```diff
@@ -0,0 +1,73 @@
...
+        return miscasedHeadings.compactMap { miscased -> Diagnostic? in
+            guard let headingRange = miscased.heading.range else { return nil }
+            let actual = miscased.heading.plainText
+            let expected = miscased.expectedHeading
+
+            return Diagnostic(
+                source: sourceFile,
+                severity: .warning,
+                range: headingRange,
+                identifier: "org.swift.docc.SectionHeadingCase",
```
d-ronnqvist: DocC's diagnostics don't use a `"org.swift.docc."` prefix anymore. Any diagnostic with that prefix is an indication that we haven't modernized that warning yet. 

It's also worth noting that these identifiers are user-facing, so it's good to make them clear and descriptive. 

See the [Adding user-facing diagnostics](https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/Adding-Diagnostics.md) contributor documentation for more information.

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
```diff
@@ -0,0 +1,73 @@
...
+                // - Words beginning with a vowel sound use "an"
+                // - Words beginning with a consonant sound use "a"
+                //
+                // This cannot always be determined from the first letter alone. For example,
+                // "honour" begins with the letter 'h', but the 'h' is silent,
+                // so it is pronounced with an initial vowel sound and uses "an":
+                //
+                // "It is an honour to meet you."
+                //
+                // To avoid applying potentially incorrect logic, the one case where "an" must be used is hardcoded.
```
d-ronnqvist: minor: I find it a bit ironic that this long comment describes how the article cannot be determined by the first letter alone and then the implementation just below is checking only the first character. 

Considering that there are only 4 values that this needs to check against I feel like the comment could convey the necessary information more succinctly. For example

```suggestion
                // The article for a word depends on the word's pronunciation, not its spelling; 
                // using "an" before a vowel sound and "a" before a consonant sound. 
                // For these 4 expected headings, only "overview" starts with a vowel sound.
```

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
snprajwal: Ah okay, I took a look at the existing ones, and some of them had the prefix and some others didn't, I wasn't sure what the convention is. Is this worth adding to the contributor docs as a note?

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
snprajwal: Haha fair enough, I meant to indicate that the reason for the first letter check is because the case is hardcoded, and not because the article can be determined with the first letter. I'll shorten it.

d-ronnqvist:
As an alternative solution; do you think that it'd be better or worse if DocC matched these headings case <I><b>in</b></i>sensitively?

snprajwal:
I don't think it's desirable to have casing variations across a catalog. If there is a catalog that multiple authors contribute to, it could happen that each one uses a different casing (See Also, see also, See also), and that inconsistency can add friction to other operations. For instance, if I want to find all "See Also" sections in a catalog at the moment, I can do a regular case-sensitive grep for `See Also`, and reasonably expect only the headings to show up. If I do a case-insensitive search, I might also get search results from places in prose where the phrase is present. If the headings are made case-insensitive, then I would need to either use a case-insensitive search and filter out the irrelevant results, or use a more specific regex for the search query.

It's not a dealbreaker, but I think it's a good idea to enforce consistency where possible. I feel warning the user to change the casing to what DocC expects is better than being more permissive with what we accept for these headings.

d-ronnqvist:
> I don't think it's desirable to have casing variations across a catalog. If there is a catalog that multiple authors contribute to, it could happen that each one uses a different casing (See Also, see also, See also), and that inconsistency can add friction to other operations. For instance, if I want to find all "See Also" sections in a catalog at the moment, I can do a regular case-sensitive grep for `See Also`, and reasonably expect only the headings to show up. If I do a case-insensitive search, I might also get search results from places in prose where the phrase is present. If the headings are made case-insensitive, then I would need to either use a case-insensitive search and filter out the irrelevant results, or use a more specific regex for the search query.

This might only be a DocC problem but if I search for "See Also" (case sensitively) I get 179 results but if I search for "## See Also" I only get 63 results. 

> It's not a dealbreaker, but I think it's a good idea to enforce consistency where possible. I feel warning the user to change the casing to what DocC expects is better than being more permissive with what we accept for these headings.

It's a matter of preference for sure. There's some merit to the simplicity (both conceptually and implementation-wise) of having a strict syntax but there is also some value in having a syntax that tries to accommodate what the user likely intended.

I was just floating this as a possible idea. I would be fine either way. This is also something that we could change in the future if we wanted to.

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
d-ronnqvist: We could add something short to the "Organize diagnostics into groups" section of the documentation about adding new diagnostics that mentions that the identifier is user-facing. We only have 10 or so updated diagnostic so far but that should be enough to find one or two identifiers to refer to as good examples.

snprajwal:
> This might only be a DocC problem but if I search for "See Also" (case sensitively) I get 179 results but if I search for "## See Also" I only get 63 results.

Ah yeah that should be exclusive to this project, since we have a number of occurences of the case sensitive variant in the source code.

> It's a matter of preference for sure. There's some merit to the simplicity (both conceptually and implementation-wise) of having a strict syntax but there is also some value in having a syntax that tries to accommodate what the user likely intended.
> I was just floating this as a possible idea. I would be fine either way. This is also something that we could change in the future if we wanted to.

Of course, we need to strike a balance 😄

snprajwal:
@swift-ci please test

Sources/SwiftDocC/SwiftDocC.docc/SwiftDocC/StaticAnalysis.md
```diff
@@ -10,14 +10,4 @@ Run static analysis checks on markup files.
...
-### Predefined Checks
-
-- ``DuplicateTopicsSections``
-- ``InvalidAdditionalTitle``
-- ``MissingAbstract``
-- ``NonInclusiveLanguageChecker``
-- ``NonOverviewHeadingChecker``
-- ``SectionHeadingCaseChecker``
-- ``SeeAlsoInTopicsHeadingChecker``
-
```
snprajwal, line 21: It's unfortunate to lose this, but internal types wouldn't show up here. I'm not sure if we want to make the types package-wide just to be able to document them here.

d-ronnqvist (changes requested):
It's a source breaking change to lower the access level of the existing `public` checker type. Because these types were public before it's _possible_—but highly unlikely—that an API client would create one of those checkers and run it themselves outside of the documentation context.

If we want to remove these from the public API safely we'd need to deprecate them for a full minor release first. That gets a bit more complicated because we still use these types internally. We'd probably need to internally rename them with an underscore prefix and leave the old name as a deprecated type alias. However, that change might be best to do in a later PR—and focus this PR on only the new checker.

snprajwal:
That's a good point, I'll split that out into a separate PR. I'll just add this new checker as a non-public one in this PR for now.

d-ronnqvist (approved):

Sources/SwiftDocC/Checker/Checkers/SectionHeadingCaseChecker.swift
d-ronnqvist: FYI: I opened [this PR](https://github.com/swiftlang/swift-docc/pull/1559) to add a bit of contributor documentation about naming diagnostic IDs and to clarify what happens for not-yet-updated diagnostics.

snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test

PR #1555, opened 2026-06-17
Disallow cross-language curation of internal links

snprajwal:
Bug/issue #, if applicable: rdar://164204630

## Summary

https://github.com/swiftlang/swift-docc/pull/283 added a feature where, if a child page in one language is curated under a parent page in a different language and they have no languages in common, the child page is retained to prevent dropping the curation entirely. However, this is relevant only when an external link is being curated. For internal links, a variant of the parent page must be created in the language of the link being curated. This patch tightens the original implementation to only allow this form of curation for external links. It also deletes the now-unneeded test bundle that contained the case with internal links.

## Dependencies

N/A

## Testing

1. Create a bundle containing a mixed-language framework.
2. Curate an ObjC symbol in a Swift-only article or symbol page.
3. Prior to this patch, the Topics section would contain the reference, but the navigator wouldn't.
4. With this patch, the Topics section does not contain the reference, and a warning is emitted.

## Checklist

- [x] Added tests


snprajwal:
@swift-ci please test

Sources/SwiftDocC/Infrastructure/DocumentationCurator.swift
d-ronnqvist, line 1: minor: As far as I know the term "curation" is jargon that we generally don't display to the end-user. Instead [DocC's user-facing documentation](https://www.swift.org/documentation/docc/adding-structure-to-your-documentation-pages) refers to the idea of curation as organizing the documentation hierarchy.

If you look at [other](https://github.com/swiftlang/swift-docc/pull/1555/changes#diff-06d437f9d5d634ca184eaa8d26a782027adb9daa998de42ff4350dd91856fd86R300-R312) [similar](https://github.com/swiftlang/swift-docc/pull/1555/changes#diff-06d437f9d5d634ca184eaa8d26a782027adb9daa998de42ff4350dd91856fd86R271-R275) [diagnostics](https://github.com/swiftlang/swift-docc/pull/1555/changes#diff-06d437f9d5d634ca184eaa8d26a782027adb9daa998de42ff4350dd91856fd86R282-R287) in the same file they phrase those issues in terms of the impact it would have on t "the documentation hierarchy", saying things like:

> If this link contributed to the documentation hierarchy it would introduce this cycle:
> ASCII_DIAGRAM_OF_CYCLE

The warnings about incorrect use of the `TechnologyRoot` is [another example](https://github.com/swiftlang/swift-docc/blob/495575b287b40ae674fce86220952eae4d84844e/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2625-L2647) of warnings that phrase the issue in terms of the impact it would have on t "the documentation hierarchy", saying things like:

> If SYMBOL_DESCRIPTION became a root page it would move out of MODULE_NAME, creating a disjoint documentation hierarchy with two possible starting points, resulting in undefined behavior for core DocC features that rely on a consistent and well defined documentation hierarchy.

Sources/SwiftDocC/Infrastructure/DocumentationCurator.swift
```diff
@@ -313,6 +313,28 @@ struct DocumentationCurator {
...
+                        \(topicSectionBaseExplanation). The curated page and the parent page have no source languages in common:
+
+                        - \(describeForDiagnostic(childReference).singleQuoted) (curated): \(childLanguages)
+                        - \(describeForDiagnostic(nodeReference).singleQuoted) (parent): \(nodeLanguages)
+
+                        Due to this, the curated page is not reachable from any variant of the parent page.
+                        """,
+                        solutions: removeListItemSolutions
+                    ))
+                    continue
```
d-ronnqvist, line 334: I may be misremembering this but from what I can recall of the warnings about cyclic curation, calling `continue` in the curator, like this, doesn't prevent the node from either
- being listed in the topic section of the curating page 
- appearing under the curating page in the navigator hierarchy 

If we want to do more than warn about this issue it might be good to have tests that verify that the mitigations work for both those behaviors.

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -719,4 +719,48 @@ struct DocumentationCuratorTests_New {
             ["doc://unit-test/documentation/unit-test/API-Collection"],
         ], "Unexpected rendered topic sections on the module page \(assertionMessageDescription)")
     }
+
+    // If a symbol is curated in a page that is not available in the symbol's language,
+    // a warning should be emitted stating that it is not reachable from the parent page.
+    // rdar://164204630
+    @Test
+    func raisesDiagnosticAboutUnreachableCrossLanguageCuration() async throws {
```
d-ronnqvist: nit: If there's information that we want to convey about a purpose of a test, I find that it's better described in a descriptive test name.
```suggestion
    @Test
    func warnsAboutCuratingSymbolWithoutAnyCommonLanguages() async throws {
```

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -719,4 +719,48 @@ struct DocumentationCuratorTests_New {
...
+
+                - ``ObjcOnlyClass``
+                """),
+            ])
+        )
+
+        let crossLanguageDiagnostics = context.diagnostics
+            .filter { $0.identifier == "UnreachableCrossLanguageCuration" }
+        let diagnostic = try #require(crossLanguageDiagnostics.first)
+        #expect(crossLanguageDiagnostics.count == 1)
```
d-ronnqvist: minor: If we're expecting only one specific diagnostic, I find it better to expect exactly that diagnostic. That way we know that the test setup contains no other issues that could have an unexpected impact on the results.

```suggestion
        #expect(diagnostics.map(\.identifier) == ["UnreachableCrossLanguageCuration"],
                "Encountered unexpected problems: \(context.diagnostics.map(\.summary))")
        let diagnostic = try #require(crossLanguageDiagnostics.first)
```

Sources/SwiftDocC/Infrastructure/DocumentationCurator.swift
snprajwal: I've added a step in the test to verify that the reference is not present in the rendered parent page.

snprajwal:
@swift-ci please test

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

PR #1557, opened 2026-06-18
Use the prose field from symbol names for inline link text

timdream:
## Summary

`SymbolGraph.Symbol.Names.prose` is documented as the name to use for inline links and documentation prose, falling back to `title` when not set. DocC was decoding it but never using it — all inline link text came from the `title` field instead. This change threads the prose value through the rendering pipeline so inline links display the prose form when available.

~~Adds `proseVariants`/`prose` to `Symbol`, populates it from the symbol graph in both DocumentationNode init paths, and also copies it in `ReferenceResolver.visitSymbol` which rebuilds the Symbol during link resolution. Applies to all three output formats: JSON archive, Markdown, and static HTML (Symbol.makeNames).~~

Adds `proseVariants` (stored, populated from the symbol graph per language variant) and `proseTitleVariants` (computed, merges prose into title with per-language fallback) to `Symbol`, with wiring in `ReferenceResolver.visitSymbol`. Applies to all three output formats — JSON archive (`LinkTitleResolver`), static HTML (`Symbol.makeNames`), and Markdown — as well as `LinkDestinationSummary` for cross-bundle linking.

## Dependencies

None.

## Testing

Added a test case. You can also test the minimal test case provided here.

Steps:
1. Unzip the test case provided and note the path to `ProseTest.docc`
2. Use `docc preview` to preview the bundle:
```
docc preview ProseTest.docc
```
Then open `http://localhost:8080` in a browser and navigate to **ProseTest → ProseTest → Canopy**
3. In the **Overview** section, the sentence reads:
```
Call calculateLightPenetration to measure light levels.
```
Verify the inline link text is `calculateLightPenetration` — the short prose form from the `prose` field in the symbol names.

Before this fix, the link text would have been `calculateLightPenetration(sunAngle)` — the full `title` field — because the `prose` field was decoded but never used. The symbol graph in the test case deliberately sets `"prose": "calculateLightPenetration"` and `"title": "calculateLightPenetration(sunAngle)"` to make the difference visible.

## Checklist

- [x] Added tests
- [ ] Updated documentation if necessary — N/A


Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -585,9 +592,13 @@ extension Symbol {
     /// The kind of the first variant of this symbol, such as protocol or variable.
     public var kind: SymbolGraph.Symbol.Kind { kindVariants.firstValue! }
     
+    /// The title of the first variant of this symbol, usually a simplified version of the declaration.
     /// The title of the first variant of this symbol, usually a simplified version of the declaration.
     public var title: String { titleVariants.firstValue! }
-    
+
+    /// The prose form of the first variant of this symbol's title, for use in inline links and documentation prose.
+    public var prose: String { proseVariants.firstValue ?? title }
```
d-ronnqvist: Please don't add any more computed properties that ignore language specific symbol information. There's a comment a few lines above this that says:
> New code should be written to work with the variants so that it supports language specific content.

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -585,9 +592,13 @@ extension Symbol {
     /// The kind of the first variant of this symbol, such as protocol or variable.
     public var kind: SymbolGraph.Symbol.Kind { kindVariants.firstValue! }
     
+    /// The title of the first variant of this symbol, usually a simplified version of the declaration.
```
d-ronnqvist: minor: This looks like the documentation is duplicated.

Sources/SwiftDocC/Model/Rendering/HTML/HTMLRenderer.swift
```diff
@@ -509,14 +509,15 @@ private extension DocumentationNode {
 
 private extension Symbol {
     func makeNames(goal: RenderGoal, fallbackTitle: String) -> LinkedElement.Names {
-        switch titleVariants.values(goal: goal) {
+        let variants = proseVariants.isEmpty ? titleVariants : proseVariants
```
d-ronnqvist: I can't find this behavior being verified anywhere in a test. 

Also, I don't _think_ this will behave correctly if the symbol has a prose name for one language but not the other. In that case I would expect the symbol to display the prose name in the language where it's available and fallback to the title for the other language. My understanding is that this implementation will fall back to the other language's prose name, which is not correct.

It would be great to verify that links to language specific symbols display the expected text both when the symbol has a prose name for all languages, a prose name for only some languages, and doesn't have a prose name at all.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -951,7 +951,8 @@ public class DocumentationContext {
         } else {
             source = .external
         }
-        let graphNode = TopicGraph.Node(reference: reference, kind: documentation.kind, source: source, title: symbol.defaultSymbol!.names.title, isVirtual: module.isVirtual)
+        let title = symbol.defaultSymbol!.names.prose ?? symbol.defaultSymbol!.names.title
+        let graphNode = TopicGraph.Node(reference: reference, kind: documentation.kind, source: source, title: title, isVirtual: module.isVirtual)
```
d-ronnqvist: Question: does this change make a difference anywhere? I didn't _think_ that we used the title of a topic graph node to compute the text that's displayed on the page anywhere (either as a title or as a link text)

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary.swift
```diff
@@ -411,7 +411,7 @@ extension LinkDestinationSummary {
         
         // Multi-language symbols need to access the default content via the variant accessors (rdar://86580516)
         let kind = DocumentationNode.kind(forKind: (symbol.kindVariants[summaryTrait] ?? symbol.kind).identifier)
-        let title = symbol.titleVariants[summaryTrait] ?? symbol.title
+        let title = symbol.proseVariants[summaryTrait] ?? symbol.prose
```
d-ronnqvist: I can't find this behavior being verified anywhere in a test.

Also, as pointed out in [this other comment](https://github.com/swiftlang/swift-docc/pull/1557/changes#r3434249794), I don't _think_ this will behave correctly if the symbol has a prose name for one language but not the other. Regardless, it would be great to verify that behavior for a few combinations of symbols with and without language-specific prose names.

Sources/SwiftDocC/Model/Rendering/LinkTitleResolver.swift
```diff
@@ -49,11 +49,11 @@ struct LinkTitleResolver {
         }
         
         if let symbol = (page.semantic as? Symbol) {
-            return symbol.titleVariants
+            return symbol.proseVariants.isEmpty ? symbol.titleVariants : symbol.proseVariants
```
d-ronnqvist: I can't find this behavior being verified anywhere in a test.

Also, as pointed out in [this other comment](https://github.com/swiftlang/swift-docc/pull/1557/changes#r3434249794), I don't _think_ this will behave correctly if the symbol has a prose name for one language but not the other. Regardless, it would be great to verify that behavior for a few combinations of symbols with and without language-specific prose names.

Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
```diff
@@ -23,9 +25,65 @@ class LinkTitleResolverTests: XCTestCase {
...
         let myProtocolNode = try context.entity(with: reference)
-        
+
         // Tests title resolving for symbols
         let title = resolver.title(for: myProtocolNode)
         XCTAssertEqual("MyProtocol", title?.allValues.first?.variant)
     }
+
+    func testSymbolProseTitleResolving() async throws {
+        let (_, context) = try await testBundleAndContext(named: "LegacyBundle_DoNotUseInNewTests")
```
d-ronnqvist: Please don't add new tests using the fixture named "LegacyBundle_DoNotUseInNewTests".

Also, please write new tests using Swift Testing. We have more guidance for writing tests in the ["Adding new tests" section](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md#adding-new-tests) of our contributing guidelines. If you're adding a new test to a file that's currently using XCTest you can create a separate test suite using Swift Testing and only add the new test there.

Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
```diff
@@ -23,9 +25,65 @@ class LinkTitleResolverTests: XCTestCase {
...
+                parametersSectionVariants: .init(swiftVariant: nil),
+                dictionaryKeysSection: nil,
+                possibleValuesSection: nil,
+                httpEndpointSection: nil,
+                httpBodySection: nil,
+                httpParametersSection: nil,
+                httpResponsesSection: nil,
+                redirects: nil
+            )
+        }
```
d-ronnqvist: minor: I don't recognize this pattern of explicitly creating a node and a symbol semantic from other tests. 

We usually write tests that load a catalog and a minimal symbol graph from an in-memory file system. You can find a handful of symbol graph helpers in [DocCTestUtilities/SymbolGraphCreation.swift](https://github.com/swiftlang/swift-docc/blob/main/Sources/DocCTestUtilities/SymbolGraphCreation.swift). If you search for code that calls them you should find a variety of examples of how tests use that API.

d-ronnqvist (changes requested):
Thanks for opening this PR. 

From what I can tell these changes are largely untested and I have some concerns that the current implementation doesn't behave correctly when a symbol has one language representation with a prose name and another language representation without a prose name.

timdream:
[prose-test-case.zip](https://github.com/user-attachments/files/29224153/prose-test-case.zip)

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -174,7 +174,7 @@ extension MarkdownOutputMarkupWalker {
                     .map { $0.spelling }
                     .joined(separator: " ")
             } else {
-                linkTitle = symbol.title
+                linkTitle = symbol.proseVariants.firstValue ?? symbol.title
```
timdream, line 241: I don't want to add `symbol.prose` in legacy extension so this is using mix syntaxes.

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -99,6 +101,20 @@ public final class Symbol: Semantic, Abstracted, Redirected, AutomaticTaskGroups
...
+    internal(set) public var proseVariants: DocumentationDataVariants<String> = .init()
+
+    /// The prose title for inline links, merged per-language: uses prose where available, falls back to title otherwise.
+    public var proseTitleVariants: DocumentationDataVariants<String> {
+        var result = titleVariants
+        for (trait, prose) in proseVariants.allValues {
+            result[trait] = prose
+        }
+        return result
+    }
```
d-ronnqvist, line 115: I don't think that this is API that we'd want to commit to. As such; it would be best to _not_ make it public.

Tests/SwiftDocCTests/DocumentationService/ConvertService/ConvertServiceTests.swift
```diff
@@ -96,6 +96,35 @@ class ConvertServiceTests: XCTestCase {
             )
         }
     }
+
+    func testConvertPageUsesProseNameForInlineLinks() throws {
```
d-ronnqvist: Like [previously remarked](https://github.com/swiftlang/swift-docc/pull/1557#discussion_r3434308819) please write new tests using Swift Testing. 

> [Please] write new tests using Swift Testing. We have more guidance for writing tests in the ["Adding new tests" section](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md#adding-new-tests) of our contributing guidelines. If you're adding a new test to a file that's currently using XCTest you can create a separate test suite using Swift Testing and only add the new test there.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
+import SymbolKit
+import DocCTestUtilities
+@testable import SwiftDocC
+
+@Suite struct ProseSymbolNameTests {
```
d-ronnqvist: AFAIK there's no need to explicitly define a `@Suite` here. Other tests don't (unless they need to configure something about the suite's behavior). Please follow the style and patterns used in the rest of the project.

```suggestion
struct ProseSymbolNameTests {
```

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+            language: language,
+            kind: .func,
+            pathComponents: ["Canopy", title]
+        )
+        symbol.names = SymbolGraph.Symbol.Names(
+            title: title,
+            navigator: symbol.names.navigator,
+            subHeading: symbol.names.subHeading,
+            prose: prose
+        )
```
d-ronnqvist: Is it not possible to assign the prose to `symbol.names.prose`? Recreating the entire Names value like this feels overly verbose unless necessary.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+        return symbol
+    }
```
d-ronnqvist: Why does this method exist? It seems like almost an exact copy of `makeMethod(title:prose:language:)` above.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+
+    private func loadedDataContext(prose: String?) async throws -> DocumentationContext {
+        let method = makeDataMethod(title: "calculateLightPenetration(sunAngle)", prose: prose)
+        return try await load(catalog: Folder(name: "unit-test.docc") {
+            JSONFile(name: "unit-test.data.symbols.json",
+                     content: makeSymbolGraph(moduleName: "unit-test", symbols: [method]))
+        })
+    }
+
+    @Test func dataSymbolProseNameIsUsedForInlineLinkTitle() async throws {
```
d-ronnqvist: From what I understand of this test, it's not specifically relevant that the symbol has a data language but rather that the symbol only has a single language representation. 

Also, please follow the style and patterns used in the rest of the project and place `@Test` and other attributes on their own line.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+        )
+        let resolver = LinkTitleResolver(context: context, source: nil)
+
+        let reference = try #require(context.documentationCache.reference(symbolID: "method-id"))
+        let node = try context.entity(with: reference)
+        let variants = try #require(resolver.title(for: node))
+
+        #expect(variants[.swift] == "calculateLightPenetration")
+        #expect(variants[.objectiveC] == "calculateLightPenetration")
+    }
```
d-ronnqvist: It would be better if these tests verified that the title and prose information made its way into the right place in the page's RenderNode output.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+
+        let reference = try #require(context.documentationCache.reference(symbolID: "method-id"))
+        let node = try context.entity(with: reference)
+        let renderNode = DocumentationNodeConverter(context: context).convert(node)
+        let summary = try #require(
+            node.externallyLinkableElementSummaries(context: context, renderNode: renderNode).first
+        )
+
+        // The summary title should use the prose form, not the full title.
+        #expect(summary.title == "calculateLightPenetration")
```
d-ronnqvist: This test doesn't verify the variant behavior for the Objective-C spelling of the title.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+        let rendered = renderer.renderSymbol(canopySemantic)
+
+        // HTMLRenderer inserts <wbr> word-break elements inside long identifiers.
+        // Strip them before checking so we can match the plain text of the link.
+        let html = rendered.content.xmlString.replacingOccurrences(of: "<wbr></wbr>", with: "")
+
+        // The inline link in the prose paragraph should use the prose form (no parameter label).
+        // It appears as: <code>calculateLightPenetration</code> inside an <a> tag.
+        #expect(html.contains(">calculateLightPenetration<"))
+    }
```
d-ronnqvist: This test doesn't verify the Objective-C spelling of the title/prose.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+        let canopyNode = try context.entity(with: canopyReference)
+        let canopySemantic = try #require(canopyNode.semantic as? Symbol)
+
+        var renderer = HTMLRenderer(reference: canopyReference, context: context, goal: .richness,
+                                    featureFlags: .init())
+        let rendered = renderer.renderSymbol(canopySemantic)
+
+        // HTMLRenderer inserts <wbr> word-break elements inside long identifiers.
+        // Strip them before checking so we can match the plain text of the link.
+        let html = rendered.content.xmlString.replacingOccurrences(of: "<wbr></wbr>", with: "")
```
d-ronnqvist: It would be good to verify that the word breaks are inserted int he expected locations regardless if the link text uses the symbol's title or its prose name.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,240 @@
...
+    private func makeMethod(
+        title: String,
+        prose: String?,
+        language: SourceLanguage = .swift
+    ) -> SymbolGraph.Symbol {
+        var symbol = makeSymbol(
+            id: "method-id",
+            language: language,
+            kind: .func,
+            pathComponents: ["Canopy", title]
```
d-ronnqvist: There are no `memberOf` relationships in the tests that make this symbol a member of the "Canopy" class.

d-ronnqvist (changes requested):

timdream:
@d-ronnqvist Thanks for the review. Test cases are re-formatted to cover all symbol combinations with and without prose (single, swift-only, objC-only, both, neither) and on all rendering paths (RenderNode, static HTML, Markdown). There is a known issue w/ Markdown given it accesses `firstValue` of multi-language presentation symbol.

Tests/SwiftDocCTests/DocumentationService/ConvertService/ConvertServiceTests.swift
```diff
@@ -2480,3 +2481,25 @@ private extension ConvertServiceTests {
...
+    func renderNodeReferenceUsesProseNameForInlineLink() async throws {
+        var method = makeSymbol(id: "method-id", kind: .func, pathComponents: ["myMethod(_:)"],
+                                docComment: "See ``myMethod(_:)`` for details.")
+        method.names.prose = "myMethod"
+
+        let context = try await load(catalog: Folder(name: "ModuleName.docc", content: [
+            JSONFile(name: "ModuleName.symbols.json", content: makeSymbolGraph(
+                moduleName: "ModuleName", symbols: [method]
+            )),
+        ]))
```
d-ronnqvist: If you search for `load(catalog:` elsewhere in the project you'll find that we tend to follow loading any catalog with ensuring that it doesn't report any issues with the content:

```swift
#expect(context.diagnostics.isEmpty, "Unexpected problems: \(context.diagnostics.map(\.summary))")
```
This both ensures that the test setup is sufficiently complete and can helps other developers who work on your code after you identify possible reasons why their changes is may break a test that you wrote.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,317 @@
...
+    func singleLanguageSymbolProseNameIsUsedForInlineLinkTitle() async throws {
+        let ref = try await inlineLinkReferenceForSingleLanguage(prose: "doSomething")
+        #expect(ref.titleVariants.defaultValue == "doSomething")
+    }
+
+    @Test
+    func singleLanguageSymbolTitleIsUsedWhenProseIsNotSet() async throws {
+        let ref = try await inlineLinkReferenceForSingleLanguage(prose: nil)
+        #expect(ref.titleVariants.defaultValue == "doSomething(with:)")
+    }
```
d-ronnqvist: This and a few other tests in this file could benefit from being parameterized instead of manually repeating the same tests with minor variations in inputs and expected outputs. 

If you search for `@Test(arguments` in the project you'll find several examples of how we use parameterized tests elsewhere. We also have some general guidance on writing tests in the ["Adding new tests" section](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md#adding-new-tests) of our contributor documentation.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
```diff
@@ -0,0 +1,317 @@
...
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
+import SymbolKit
+import DocCTestUtilities
+@testable import SwiftDocC
+
+struct ProseSymbolNameTests {
```
d-ronnqvist, line 16: With the way that these tests are structured I find it hard to compare the expected behavior across different output formats to see that the same behaviors are covered across all output formats. 

Another way to think about this would be: if we add another output format in the future how would it know which tests it needs to repeat?

---

I think that it would be beneficial to structure these tests so that each test verifies all of the output formats for that setup. That makes it easier for the reader, and future developer, of the test code to see that the logic is consistently applied for each setup.

Tests/SwiftDocCTests/Rendering/ProseSymbolNameTests.swift
d-ronnqvist: Note: if the test are restructured to verify all output formats for each test setup then there may be different repeated patterns that may or may not be suitable for parameterized tests anymore.

Tests/SwiftDocCTests/DocumentationService/ConvertService/ConvertServiceTests.swift
```diff
@@ -2480,3 +2481,25 @@ private extension ConvertServiceTests {
         return temporaryDirectory
     }
 }
+
+struct ConvertServiceProseNameTests {
+    @Test
+    func renderNodeReferenceUsesProseNameForInlineLink() async throws {
+        var method = makeSymbol(id: "method-id", kind: .func, pathComponents: ["myMethod(_:)"],
+                                docComment: "See ``myMethod(_:)`` for details.")
+        method.names.prose = "myMethod"
```
d-ronnqvist: minor: It may help future developers understand this test setup if the names were more distinct and perhaps even self-identifying. For example `symbolTitle` and `proseName`.

Also: this test never verifies that link (to itself) in the symbol's documentation comment, so that could be removed to avoid confusing the reader (wondering where that content is checked).

d-ronnqvist (changes requested):
With the way that `ProseSymbolNameTests` separates each output format into its own list of test it's hard as a reader to know that the same setups are covered across all output formats and its harder for future developers adding new test to know that they're covering all output formats. 

Please update the tests so that each test setup verifies the same behavior for all output formats together.

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

PR #1561, opened 2026-06-18
Replace legacy bundle in link render context and chapter tests

franklinsch:

## Summary

Replaces uses of the `LegacyBundle_DoNotUseInNewTests` fixture in render context and link resolver tests. There's no strong relationship between those tests, I just bundled them together since they were easier to migrate.

## Dependencies

None.

## Testing

No functional changes.

## Checklist

- [x] Added tests

### DIFF
--- a/Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
+++ b/Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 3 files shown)


Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
```diff
@@ -11,17 +11,20 @@
...
+import DocCTestUtilities
 
 class LinkTitleResolverTests: XCTestCase {
     func testSymbolTitleResolving() async throws {
-        let (_, context) = try await testBundleAndContext(named: "LegacyBundle_DoNotUseInNewTests")
+        let catalog = Folder(name: "unit-test.docc", content: [
+            JSONFile(name: "MyKit.symbols.json", content: makeSymbolGraph(moduleName: "MyKit", symbols: [
+                makeSymbol(id: "s:6MyKit10MyProtocolP", kind: .protocol, pathComponents: ["MyProtocol"]),
+            ])),
+        ])
```
d-ronnqvist: FYI (non-blocking): This can be expressed slightly more concisely with some recent-ish updates to our test helpers (that only a handful of tests are utilizing so far):

```suggestion
        let catalog = Folder(name: "unit-test.docc") {
            JSONFile(symbolGraph: makeSymbolGraph(moduleName: "MyKit", symbols: [
                makeSymbol(id: "s:6MyKit10MyProtocolP", kind: .protocol, pathComponents: ["MyProtocol"]),
            ]))
        }
```

(The same also applies to the other updated tests)

d-ronnqvist (approved):
Thanks for updating these test and removing even more places that use the legacy fixture.

Tests/SwiftDocCTests/Rendering/LinkTitleResolverTests.swift
franklinsch: Nice okay, thanks for the pointer!

franklinsch:
@swift-ci please test

PR #1558, opened 2026-06-18
Add slightly more contributor guidance about writing and naming tests

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This updates the contributing guidelines with slightly more information about writing and naming tests.

## Dependencies

None.

## Testing

Nothing in particular. This is a documentation-only change.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- ~[ ] Added tests~ This is a documentation-only change.
- ~[ ] Ran the `./bin/test` script and it succeeded~ This is a documentation-only change.
- [x] Updated documentation if necessary

PR #1562, opened 2026-06-18
fix(static-hosting): preserve custom templates

### LINKED ISSUE #1532: docc process-archive transform-for-static-hosting doesn't preserve custom header/footer HTML from --experimental-enable-custom-templates
### Description

While working with creating the combined documentation for Swift, I'm explicitly doing a post-processing command the change the hosting base path, and found that it's regenerating a stripped stub for every symbol where the hosting base path changes, loosing any custom injected HTML header or footer data.

### Checklist

### Expected Behavior

I was anticipating that the first archive in the merge set would set the stage for the HTML included, and then the hosting base path could be adjusted without impacting what injected HTML content in each of the stubs.

### Actual behavior

After docc convert (or docc merge), the resulting archive's index.html correctly contains custom-header/custom-footer templates. I expected process-archive transform-for-static-hosting to either preserve those templates in the per-route stubs it generates, or copy them through verbatim. Instead, every generated stub is a bare shell.

### Steps To Reproduce

1. create archive with customer header/footer
2. apply a hosting base path
3. view using python -m http.server with the relevant location reflected
4. note the missing HTML

6.3.2

```shell
Apple Swift version 6.3.2 (swift-6.3.2-RELEASE)
Target: arm64-apple-macosx26.0

(also applicable on `main`)
```

agisilaos:
## Summary

Fixes #1532.

This preserves custom header and footer templates when `docc process-archive transform-for-static-hosting` regenerates route stub pages.

When an archive already contains custom templates in its root `index.html`, transforming it with a hosting base path previously generated bare per-route `index.html` files and dropped the custom template HTML. This now carries the archive's `custom-header` and `custom-footer` templates into the generated route pages while still applying the requested hosting base path.

## Changes

- Reads the archive root `index.html` before in-place template copying can replace it.
- Preserves `custom-header` and `custom-footer` template blocks in generated static-hosting route pages.
- Uses the injected file manager consistently when building transformed index HTML.
- Adds regression coverage for both external-output and in-place transform modes.

## Alternatives considered

- Leaving route stubs bare was rejected because it loses archive customizations.
- Copying the archive root `index.html` verbatim was rejected because route stubs still need the newly transformed hosting base path.

## Notes

The regression tests remain in the existing XCTest-based static-hosting test file to match the surrounding test suite and reuse its temporary-directory helpers.

## Screenshots

Not applicable; no UI changes.

## Testing

- `swift test --filter TransformForStaticHostingActionTests`
- `swift test --filter StaticHostableTransformerTests`

### DIFF
--- a/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
+++ b/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
@@ -95,7 +95,8 @@ extension StaticHostableTransformer {
     static func indexHTMLData(
         in htmlTemplateDirectory: URL,
         with hostingBasePath: String?,
-        fileManager: any FileManagerProtocol
+        fileManager: any FileManagerProtocol,
+        preservingCustomTemplatesFrom archiveIndexHTML: URL? = nil
     ) throws -> Data {
         let customHostingBasePathProvided = !(hostingBasePath?.isEmpty ?? true)
         
@@ -127,6 +128,30 @@ extension StaticHostableTransformer {
             indexHTML = indexHTML.replacingOccurrences(of: HTMLTemplate.tag.rawValue, with: replacementString)
         }
         
+        if let archiveIndexHTML,
+           let archiveIndexData = try? fileManager.contents(of: archiveIndexHTML),
+           let archiveIndexContents = String(data: archiveIndexData, encoding: .utf8)
+        {
+            let customTemplates = Self.customTemplateHTML(in: archiveIndexContents)
+            if !customTemplates.isEmpty,
+               let bodyTagRange = indexHTML.range(of: "<body[^>]*>", options: .regularExpression)
+            {
+                indexHTML.replaceSubrange(bodyTagRange, with: indexHTML[bodyTagRange] + customTemplates)
+            }
+        }
+
         return Data(indexHTML.utf8)
+    }
+
+    private static func customTemplateHTML(in indexHTML: String) -> String {
+        let pattern = #"<template\s+id="custom-(?:header|footer)"[^>]*>[\s\S]*?</template>"#
+        guard let regex = try? NSRegularExpression(pattern: pattern) else {
+            return ""
+        }
+
+        let range = NSRange(indexHTML.startIndex..<indexHTML.endIndex, in: indexHTML)
+        return regex.matches(in: indexHTML, range: range).compactMap { match in
+            Range(match.range, in: indexHTML).map { String(indexHTML[$0]) }
+        }.joined()
     }
 }
--- a/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
+++ b/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
@@ -145,9 +145,7 @@ extension StaticHostableTransformer {
 
     private static func customTemplateHTML(in indexHTML: String) -> String {
         let pattern = #"<template\s+id="custom-(?:header|footer)"[^>]*>[\s\S]*?</template>"#
-        guard let regex = try? NSRegularExpression(pattern: pattern) else {
-            return ""
-        }
+        let regex = try! NSRegularExpression(pattern: pattern)
 
         let range = NSRange(indexHTML.startIndex..<indexHTML.endIndex, in: indexHTML)
         return regex.matches(in: indexHTML, range: range).compactMap { match in
--- a/Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
+++ b/Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
@@ -8,6 +8,7 @@
  See https://swift.org/CONTRIBUTORS.txt for Swift project authors
 */
 
+import Testing
 import XCTest
 import Foundation
 @testable import SwiftDocC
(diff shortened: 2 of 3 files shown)


Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
```diff
@@ -173,5 +173,122 @@ class TransformForStaticHostingActionTests: StaticHostingBaseTests {
...
 
+    func testTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePagesForExternalOutput() async throws {
+        try await assertTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePages(outputIsExternal: true)
+    }
+
+    func testTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePagesForInPlaceOutput() async throws {
+        try await assertTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePages(outputIsExternal: false)
+    }
+
+    private func assertTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePages(outputIsExternal: Bool) async throws {
```
d-ronnqvist: Please use Swift Testing for new tests. The ["Updating existing tests" section](https://github.com/swiftlang/swift-docc/blob/main/CONTRIBUTING.md#updating-existing-tests) of the  contributing guidelines contains some additional information (with some more information being [added in this PR](https://github.com/swiftlang/swift-docc/pull/1558/changes#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R318-R328)).

Because this is literally running the same code both a `true` and `false` literal as an argument, it would be great to utilize a parameterized test instead.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
```diff
@@ -173,5 +173,122 @@ class TransformForStaticHostingActionTests: StaticHostingBaseTests {
...
+            documentationBundleURL: bundleURL,
+            outOfProcessResolver: nil,
+            analyze: false,
+            targetDirectory: archiveURL,
+            htmlTemplateDirectory: convertTemplateURL,
+            emitDigest: false,
+            currentPlatforms: nil,
+            temporaryDirectory: try createTemporaryDirectory(),
+            experimentalEnableCustomTemplates: true
+        )
```
d-ronnqvist: Please avoid modifying the real file system where possible. Instead look for a test that uses a `TestFileSystem` and passes that to a `ConvertAction`.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
```diff
@@ -173,5 +173,122 @@ class TransformForStaticHostingActionTests: StaticHostingBaseTests {
...
+
+    private func assertTransformForStaticHostingPreservesCustomTemplatesInGeneratedRoutePages(outputIsExternal: Bool) async throws {
+        let displayName = "CustomTemplateStaticHosting"
+        let identifier = "com.test.custom-template-static-hosting"
+        let headerContent = "<header>custom header</header>"
+        let footerContent = "<footer>custom footer</footer>"
+        let hostingBasePath = "custom/base/path"
+
+        let bundle = Folder(name: "\(displayName).docc", content: [
+            InfoPlist(displayName: displayName, identifier: identifier),
```
d-ronnqvist: minor: This file (and its referenced data) isn't verified by the test and can be removed.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
```diff
@@ -173,5 +173,122 @@ class TransformForStaticHostingActionTests: StaticHostingBaseTests {
...
+
+            @Metadata {
+                @TechnologyRoot
+            }
+
+            An abstract.
+
+            ## Overview
+
+            Some discussion.
```
d-ronnqvist: minor: It's not necessary to specify that the only page is the root. 

Also, none of this article content is verified in the test. In such cases we prefer to only use a brief abstract that describes what role the article has in the test and if there's anything specific about its content. In this case, because the content isn't verified, the abstract could be something general like "Some abstract for this root page" or "Any abstract for this root page".

Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
```diff
@@ -127,6 +128,30 @@ extension StaticHostableTransformer {
...
+        }
+
         return Data(indexHTML.utf8)
     }
+
+    private static func customTemplateHTML(in indexHTML: String) -> String {
+        let pattern = #"<template\s+id="custom-(?:header|footer)"[^>]*>[\s\S]*?</template>"#
+        guard let regex = try? NSRegularExpression(pattern: pattern) else {
+            return ""
+        }
```
d-ronnqvist: minor: If this pattern is known at compile time and known to be valid, it can force try here to avoid handling an error that will never happen.

Also, I _think_ that DocC can use Swift's build in regular expression support now, which could avoid some of this error handling and avoid some of the conversations between `Range` and `NSRange` below.

d-ronnqvist (changes requested):

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
agisilaos: Done in 8852f9f05. I converted this regression to a Swift Testing parameterized test with `@Test(arguments: [true, false])` for external and in-place output.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
agisilaos: Done in 8852f9f05. The regression now builds the catalog/templates in `TestFileSystem` and passes that file manager to both `ConvertAction` and `TransformForStaticHostingAction`.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
agisilaos: Done in 8852f9f05. I removed the unverified fixture data and kept only the files needed for the regression.

Tests/DocCCommandLineTests/TransformForStaticHostingActionTests.swift
agisilaos: Done in 8852f9f05. I simplified the page content to a minimal abstract since the article body itself isn’t what the test verifies.

Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
agisilaos: Done in 8852f9f05. I switched the static regex construction to `try!` since the pattern is known at compile time.

@d-ronnqvist Thank you so much for taking the time to review! My next PR is gonna be even better, will keep everything in mind. Hopefully everything has been adressed now!

agisilaos:
@d-ronnqvist any updates on this? is this good to go?

PR #1564, opened 2026-06-19
Replace legacy bundle in server tests

franklinsch:

## Summary

Replaces uses of the `LegacyBundle_DoNotUseInNewTests` fixture in the file server tests with an in-memory file system. To enable this, the file server providers were made injectable with a `FileManagerProtocol`.

## Dependencies

None.

## Testing

No functional changes.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Servers/FileServer.swift
+++ b/Sources/SwiftDocC/Servers/FileServer.swift
@@ -182,17 +182,23 @@ public protocol FileServerProvider {
 public class FileSystemServerProvider: FileServerProvider {
     
     private(set) var directoryURL: URL
+    private let fileManager: any FileManagerProtocol
     
-    public init?(directoryPath: String) {
-        guard FileManager.default.directoryExists(atPath: directoryPath) else {
+    public convenience init?(directoryPath: String) {
+        self.init(directoryPath: directoryPath, fileManager: FileManager.default)
+    }
+    
+    package init?(directoryPath: String, fileManager: any FileManagerProtocol) {
+        guard fileManager.directoryExists(atPath: directoryPath) else {
             return nil
         }
         self.directoryURL = URL(fileURLWithPath: directoryPath)
+        self.fileManager = fileManager
     }
     
     public func data(for path: String) -> Data? {
         let finalURL = directoryURL.appendingPathComponent(path)
-        return try? Data(contentsOf: finalURL)
+        return try? fileManager.contents(of: finalURL)
     }
     
 }
@@ -202,7 +208,19 @@ public class MemoryFileServerProvider: FileServerProvider {
     /// Files to serve based on relative path.
     private var files = [String: Data]()
     
-    public init() {}
+    private let fileManager: any FileManagerProtocol
+
+    /// Creates a memory file server provider with the default file manager.
+    public convenience init() {
+        self.init(fileManager: FileManager.default)
+    }
+
+    /// Creates a memory file server provider with the given file manager.
+    ///
+    /// - Parameter fileManager: The file manager that the provider uses to read files added with ``addFiles(inFolder:inSubPath:recursive:)``.
+    package init(fileManager: any FileManagerProtocol) {
+        self.fileManager = fileManager
+    }
     
     
     /// Add a file to the file server.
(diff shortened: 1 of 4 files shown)


Sources/SwiftDocC/Servers/FileServer.swift
```diff
@@ -241,18 +247,20 @@ public class MemoryFileServerProvider: FileServerProvider {
     ///   - destination: The destination directory in the file server to add the files to.
     ///   - recursive: Whether or not to recursively add files from the source directory.
     public func addFiles(inFolder source: String, inSubPath destination: String = "", recursive: Bool = true) {
-        var isDirectory: ObjCBool = false
-        guard FileManager.default.fileExists(atPath: source, isDirectory: &isDirectory) else { return }
-        guard isDirectory.boolValue else { return }
+        addFiles(inFolder: source, inSubPath: destination, recursive: recursive, fileManager: FileManager.default)
```
d-ronnqvist: Why is this passing a new `FileManager.default` instead of using the provider's `fileManager` property?

Sources/SwiftDocC/Servers/FileServer.swift
```diff
@@ -241,18 +247,20 @@ public class MemoryFileServerProvider: FileServerProvider {
...
+    }
+    
+    package func addFiles(inFolder source: String, inSubPath destination: String = "", recursive: Bool = true, fileManager: any FileManagerProtocol) {
```
d-ronnqvist: Why does this have a `fileManager` parameter when the provider also has a `fileManager` property?

Tests/SwiftDocCTests/Servers/DocumentationSchemeHandlerTests.swift
```diff
@@ -15,17 +15,31 @@ import FoundationNetworking
...
+    /// Contains `images/figure1.jpg` at `/template`.
+    private func makeTemplate() throws -> (fileManager: TestFileSystem, url: URL) {
+        let fileManager = try TestFileSystem(folders: [
+            Folder(name: "template") {
+                Folder(name: "images") {
+                    DataFile(name: "figure1.jpg", data: Data())
+                }
+            }
+        ])
+        return (fileManager, URL(fileURLWithPath: "/template"))
```
d-ronnqvist: Would it make sense to leverage `makeTestFileSystemWith(_:)` for this?

Sources/SwiftDocC/Servers/FileServer.swift
```diff
@@ -241,18 +247,20 @@ public class MemoryFileServerProvider: FileServerProvider {
...
-        for file in enumerator {
-            guard let file = file as? String else { fatalError("Enumerator returned an unexpected type.") }
-            guard let data = try? Data(contentsOf: URL(fileURLWithPath: source).appendingPathComponent(file)) else { continue }
-            if recursive == false && file.contains("/") { continue } // skip if subfolder and recursive is disabled
-            addFile(path: "/\(trimmedSubPath)/\(file)", data: data)
+        for fileURL in fileManager.recursiveFiles(startingPoint: sourceURL, options: []) {
+            let relativePath = fileURL.path.removingPrefix(sourceURL.path).removingPrefix("/")
+            if recursive == false && relativePath.contains("/") { continue }
+            guard let data = try? fileManager.contents(of: fileURL) else { continue }
+            addFile(path: "/\(trimmedSubPath)/\(relativePath)", data: data)
```
d-ronnqvist, line 271: With the current implementation that has a dedicated `fileManager` parameter; this doesn't forward that value to the inner files, meaning that for inner folders the file manager changes to `FileManager.default`

Sources/SwiftDocC/Servers/FileServer.swift
franklinsch: `MemoryFileServerProvider` doesn't have a `fileManager` property, but I think it's worth adding one, so I will

Tests/SwiftDocCTests/Servers/DocumentationSchemeHandlerTests.swift
franklinsch: Nice thanks, I didn't realise that existed.

Sources/SwiftDocC/Servers/FileServer.swift
d-ronnqvist: I see. That was truncated in the diff so I thought that this was a method on `FileSystemServerProvider` (above) where you added a `fileManager` property in this PR.

franklinsch: I'm a bit confused by `MemoryFileServerProvider` not inheriting `FileSystemServerProvider` but I will leave them as is for now.

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

d-ronnqvist (approved):

PR #1565, opened 2026-06-19
Replace legacy bundle in benchmark hash tests

franklinsch:

## Summary

Replaces uses of the `LegacyBundle_DoNotUseInNewTests` fixture in `TopicGraphHashTests` and `ExternalTopicsGraphHashTests` with synthetic in-memory catalogs.

## Dependencies

None.

## Testing

No functional changes.

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Benchmark/TopicGraphHashTests.swift
```diff
@@ -11,11 +11,27 @@
...
+            JSONFile(name: "MyKit.symbols.json", content: makeSymbolGraph(moduleName: "MyKit", symbols: [
+                makeSymbol(id: "s:6MyKit5MyClassC", kind: .class, pathComponents: ["MyClass"]),
+                makeSymbol(id: "s:6MyKit11myFunctionFyyF", kind: .func, pathComponents: ["myFunction"]),
+            ])),
+        ])
+    }
+
     func testTopicGraphSameHash() async throws {
+        let catalog = minimalCatalog
         func computeTopicHash(file: StaticString = #filePath, line: UInt = #line) async throws -> String {
```
d-ronnqvist: nit (non-blocking): This doesn't use the `file` or `line` values.
```suggestion
        func computeTopicHash() async throws -> String {
```

Tests/SwiftDocCTests/Benchmark/ExternalTopicsHashTests.swift
```diff
@@ -35,9 +36,28 @@ class ExternalTopicsGraphHashTests: XCTestCase {
...
+    /// `sideclass.md` can curate just the external links and let the rest be
+    /// curated automatically.
+    private func makeSideKitCatalog(sideclassMarkdown: String) -> Folder {
+        Folder(name: "unit-test.docc", content: [
+            JSONFile(name: "SideKit.symbols.json", content: makeSymbolGraph(
+                moduleName: "SideKit",
+                symbols: [
+                    makeSymbol(id: "s:7SideKit0A5ClassC", kind: .class, pathComponents: ["SideClass"]),
+                    makeSymbol(id: "s:7SideKit0A5ClassC10myFunctionyyF", kind: .method, pathComponents: ["SideClass", "myFunction()"]),
+                    makeSymbol(id: "s:7SideKit0A5ClassC4pathSSvp", kind: .property, pathComponents: ["SideClass", "path"]),
```
d-ronnqvist: minor: This symbol graph doesn't have any `memberOf` relationships for these two symbols.

d-ronnqvist: Actually, I can't quite see if these symbols are ever referenced in the tests below. Would the test still pass if we removed them?

Tests/SwiftDocCTests/Benchmark/ExternalTopicsHashTests.swift
```diff
@@ -100,8 +122,13 @@ class ExternalTopicsGraphHashTests: XCTestCase {
...
             - <doc://\(externalResolver.bundleID)/path/to/external/symbol2>
-            """.write(to: url.appendingPathComponent("documentation/sideclass.md"), atomically: true, encoding: .utf8)
-            }
+            """)
+            let (_, context) = try await self.loadBundle(catalog: catalog, configuration: {
+                var configuration = DocumentationContext.Configuration()
+                configuration.externalDocumentationConfiguration.sources = [externalResolver.bundleID: externalResolver]
+                configuration.externalDocumentationConfiguration.globalSymbolResolver = self.externalSymbolResolver
+                return configuration
+            }())
```
d-ronnqvist: nit (non-blocking): Stylistically most other tests create a configuration value up-front and pass it as a an argument directly (instead of returning the configuration from an inline closure)

Tests/SwiftDocCTests/Benchmark/TopicGraphHashTests.swift
```diff
@@ -11,11 +11,27 @@
...
+    /// deterministic and `testTopicGraphChangedHash` can find an existing node
+    /// to add an edge from.
+    private var minimalCatalog: Folder {
+        Folder(name: "unit-test.docc", content: [
+            JSONFile(name: "MyKit.symbols.json", content: makeSymbolGraph(moduleName: "MyKit", symbols: [
+                makeSymbol(id: "s:6MyKit5MyClassC", kind: .class, pathComponents: ["MyClass"]),
+                makeSymbol(id: "s:6MyKit11myFunctionFyyF", kind: .func, pathComponents: ["myFunction"]),
+            ])),
+        ])
+    }
```
d-ronnqvist: minor: From what I understand of this comment, this test only really needs there to be two pages but it doesn't matter what the module name is, what the symbol names are, what kinds of symbols they are, or what the symbol identifiers are. If so, most of this could be communicated by using less specific values in the test content.
```suggestion
    /// We need at least two pages to verify that the topic graph hash is computed deterministically.
    private var catalogWithTwoSymbols: Folder {
        Folder(name: "unit-test.docc") {
            JSONFile(symbolGraph: makeSymbolGraph(moduleName: "ModuleName", symbols: ["First, "Second"].map { name in 
                makeSymbol(id: "\(name.lowercased())-symbol-id", kind: .class, pathComponents: [name])
            ]))
        }
    }
```

Tests/SwiftDocCTests/Benchmark/ExternalTopicsHashTests.swift
```diff
@@ -35,9 +36,28 @@ class ExternalTopicsGraphHashTests: XCTestCase {
...
+                moduleName: "SideKit",
+                symbols: [
+                    makeSymbol(id: "s:7SideKit0A5ClassC", kind: .class, pathComponents: ["SideClass"]),
+                    makeSymbol(id: "s:7SideKit0A5ClassC10myFunctionyyF", kind: .method, pathComponents: ["SideClass", "myFunction()"]),
+                    makeSymbol(id: "s:7SideKit0A5ClassC4pathSSvp", kind: .property, pathComponents: ["SideClass", "path"]),
+                ]
+            )),
+            TextFile(name: "sideclass.md", utf8Content: sideclassMarkdown),
+        ])
+    }
```
d-ronnqvist: minor: From what I understand of this comment and the tests that use it, this test only really needs there to be a page that some markup can be associated with. If it wasn't that each test assumed that it was extending `SideKit/SideClass`, this catalog could either be:
- An empty module, where the provided markdown value is used as a documentation extension file for that module (the helper could provide the level-1 header so that the tests doesn't need to repeat that level-1 header symbol link)
```suggestion
    private func makeExampleCatalog(markdown: String) -> Folder {
        Folder(name: "unit-test.docc") {
            JSONFile(symbolGraph: makeSymbolGraph(moduleName: "Something", symbols: [])
            TextFile(name: "something", utf8Content: """
            # ``Something``
            
            " + markdown)
        }
    }
```

- A module with any symbol where the provided markdown value is used as an in-source documentation comment.
```suggestion
    private func makeExampleCatalog(markdown: String) -> Folder {
        Folder(name: "unit-test.docc") {
            JSONFile(symbolGraph: makeSymbolGraph(moduleName: "Something", symbols: [
                makeSymbol(id: "some-class-id", kind: .class, pathComponents: ["SomeClass"], docComment: markdown)
            ])
        }
    }
```

Tests/SwiftDocCTests/Benchmark/ExternalTopicsHashTests.swift
franklinsch: Ah yes, thanks for spotting that and agreed, it'd be better for the catalogs to be as small as possible.

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

PR #1568, opened 2026-06-25
Render the `@Small` directive in the static HTML output

YooGyeongMo:
Bug/issue #, if applicable: rdar://165755944

## Summary

The static HTML renderer (`DocCHTML`) previously discarded all block directives, producing empty output for them. This PR adds rendering for the `@Small` directive, wrapping its content in an HTML `<small>` element.

### Before
  ```swift
  func visit(_: BlockDirective) -> XMLNode {
      .text("") // all block directives produced empty output
  }
```
### After
```swift
  func visit(_ directive: BlockDirective) -> XMLNode {
      switch directive.name {
      case "Small":
          return .element(named: "small", children: visit(directive.children))
      default:
          return .text("") // other block directives still TODO (rdar://165755944)
      }
  }
```
This is a first step toward supporting block directives in the static HTML output; other directives (`@Row`, `@Image`, etc.) remain as a follow-up.

### Open question
`@Small` content is wrapped in a paragraph, so the output is `<small><p>…</p></small>`. I kept the paragraph for now - happy to unwrap it to inline content if reviewers prefer.

## Dependencies

None.

## Testing

Steps:
1. Build a documentation catalog that uses the `@Small` directive.
2. Run docc convert with --output-format experimental-html-for-development.
3. The `@Small` content now appears inside a <small> element in the generated HTML.

Unit tests covering both `@Small` rendering and the unchanged behavior of other directives are included.
I added a `parseOptions` parameter (defaulting to the existing `.parseSymbolLinks`) to the test helper so directive rendering can be tested; existing tests are unaffected.

## Checklist

- [x] Added tests

### DIFF
--- a/Tests/DocCHTMLTests/MarkdownRendererTests.swift
+++ b/Tests/DocCHTMLTests/MarkdownRendererTests.swift
@@ -292,24 +292,14 @@ struct MarkdownRendererTests {
     }
     
     @Test
-    func renderingSmallDirectives() async throws {
+    func renderingSmallDirective() async throws {
         assert(
             rendering: """
             @Small {
                 Licensed under Apache License v2.0
             }
             """,
-            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
-            matches: "<small><p>Licensed under Apache License v2.0</p></small>")
-        
-        assert(
-            rendering: """
-            @Row {
-                empty
-            }
-            """,
-            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
-            matches: "")
+            matches: "<p><small>Licensed under Apache License v2.0</small></p>")
     }
     
     @Test
@@ -620,7 +610,6 @@ struct MarkdownRendererTests {
     
     private func assert(
         rendering markdownContent: String,
-        parseOptions: ParseOptions = .parseSymbolLinks,
         elementToReturn: LinkedElement? = nil,
         assetToReturn: LinkedAsset? = nil,
         fallbackLinkTextToReturn: String? = nil,
@@ -637,7 +626,7 @@ struct MarkdownRendererTests {
                 fallbackLinkTextToReturn: fallbackLinkTextToReturn
             )
         )
-        let htmlNodes = Document(parsing: markdownContent, options: parseOptions).children.map { renderer.visit($0) }
+        let htmlNodes = Document(parsing: markdownContent, options: [.parseSymbolLinks, .parseBlockDirectives]).children.map { renderer.visit($0) }
         htmlNodes.assertMatches(prettyFormatted: prettyFormatted, expectedXMLString: expectedHTML, sourceLocation: sourceLocation)
     }
     
--- a/Tests/DocCHTMLTests/MarkdownRendererTests.swift
+++ b/Tests/DocCHTMLTests/MarkdownRendererTests.swift
@@ -292,6 +292,27 @@ struct MarkdownRendererTests {
     }
     
     @Test
+    func renderingSmallDirectives() async throws {
+        assert(
+            rendering: """
+            @Small {
+                Licensed under Apache License v2.0
+            }
+            """,
+            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
+            matches: "<small><p>Licensed under Apache License v2.0</p></small>")
+        
+        assert(
+            rendering: """
+            @Row {
+                empty
+            }
+            """,
+            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
+            matches: "")
+    }
+    
+    @Test
     func renderingCodeBlocks() async throws {
         assert(
             rendering: """
@@ -599,6 +620,7 @@ struct MarkdownRendererTests {
     
     private func assert(
         rendering markdownContent: String,
+        parseOptions: ParseOptions = .parseSymbolLinks,
         elementToReturn: LinkedElement? = nil,
         assetToReturn: LinkedAsset? = nil,
         fallbackLinkTextToReturn: String? = nil,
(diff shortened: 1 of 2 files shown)


Tests/DocCHTMLTests/MarkdownRendererTests.swift
```diff
@@ -599,6 +620,7 @@ struct MarkdownRendererTests {
     
     private func assert(
         rendering markdownContent: String,
+        parseOptions: ParseOptions = .parseSymbolLinks,
```
YooGyeongMo: Added this parameter so directive rendering can be tested; defaults to the existing behavior so other tests are unaffected.

Tests/DocCHTMLTests/MarkdownRendererTests.swift
```diff
@@ -291,6 +291,27 @@ struct MarkdownRendererTests {
...
+            matches: "<small><p>Licensed under Apache License v2.0</p></small>")
+        
+        assert(
+            rendering: """
+            @Row {
+                empty
+            }
+            """,
+            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
+            matches: "")
```
d-ronnqvist: minor: I don't think it's necessary to test the not-yet-implemented support for other directives. 
```suggestion
```

Tests/DocCHTMLTests/MarkdownRendererTests.swift
d-ronnqvist: Question: would any of the other tests be impacted if the `Document` always used both options? If not, I would find it preferable that individual tests don't have control over the parsing options.

Tests/DocCHTMLTests/MarkdownRendererTests.swift
```diff
@@ -291,6 +291,27 @@ struct MarkdownRendererTests {
...
+    @Test
+    func renderingSmallDirectives() async throws {
+        assert(
+            rendering: """
+            @Small {
+                Licensed under Apache License v2.0
+            }
+            """,
+            parseOptions: [.parseSymbolLinks, .parseBlockDirectives],
+            matches: "<small><p>Licensed under Apache License v2.0</p></small>")
```
d-ronnqvist: From what I understand; this is semantically the other way around. 

According to the HTML specification a [`<small>` element](https://html.spec.whatwg.org/#the-small-element) can be used where "phrasing content" is expected and accepts "phasing contents" as its own content. However, a [`<p>` element](https://html.spec.whatwg.org/#the-p-element) is "flow content" (that also accepts phrasing contents as its content). 

If I understand correctly—which seems to be confirmed by the examples in the HTML specification—it would instead be expected to have the `<p>` element wrap the `<small>` element.

d-ronnqvist (commented):
Thanks for adding support for `@Small` in the static HTML output. 

Did you also want to add CSS that styles the `<small>` element the pages created by DocC-Render or do you prefer to leave that to another PR (potentially by someone else)?

Tests/DocCHTMLTests/MarkdownRendererTests.swift
YooGyeongMo: @d-ronnqvist Oh Thanks for refs ! 👍 i will fix this :)

Tests/DocCHTMLTests/MarkdownRendererTests.swift
YooGyeongMo: @d-ronnqvist oh i see  I'd used `@Row` as a concrete upcoming directive, but you're right that testing not-yet-implemented behavior isn't needed. Removed :)

Tests/DocCHTMLTests/MarkdownRendererTests.swift
YooGyeongMo: I'd added the parameter to keep the existing tests unaffected, but moving both options into the helper sounds cleaner. I'll verify there's no impact and switch 👍

YooGyeongMo:
> Did you also want to add CSS that styles the `<small>` element the pages created by DocC-Render or do you prefer to leave that to another PR (potentially by someone else)?

I’d prefer to keep this PR focused on adding `@Small` support to the static HTML output.

Once I’ve had a chance to look into how DocC-Render handles styling, I’d be happy to follow up with a separate PR for the `<small>` CSS if that sounds good.

Tests/DocCHTMLTests/MarkdownRendererTests.swift
YooGyeongMo: Ran the DocCHTML tests and they all pass with both options enabled, so I moved both into the helper and removed the per-test parameter.

YooGyeongMo:
Pushed an update addressing the feedback. Once this PR is merged, I'll follow up with the `<small>` CSS in a separate PR.

YooGyeongMo:
Hi @d-ronnqvist :)

I’ve been following the [1/n]–[4/n] series (#1571, #1578, #1583, #1584), and it looks like the HTML rendering direction is moving away from XMLNode.element(named:) toward the new formatter/accessor/result builder API.

To avoid adding another .element(named:) call site that would need to be migrated later, I’m happy to wait until #1583/#1584 lands and then rebase this PR so `@Small` uses the new .small(...) accessor / result builder path.

After that, I’d also be happy to help port the block-directive rendering to the new infrastructure and continue with `@Row` and `@Image` as follow-ups, if that would be useful.

I’m fine with either keeping this PR open and rebasing later, or closing it and reopening against the new API — whichever you prefer.

d-ronnqvist (approved):

d-ronnqvist:
I don't think there's any reason to hold these changes. It won't be until the 5th PR in that series (I think) that we'll move away from creating XMLNode elements and I expect it to take a month or more before we get to that point.

d-ronnqvist:
@swift-ci please test

PR #1571, opened 2026-06-28
[1/n] Add formatter that preserves whitespace in `pre` tags and formats void tags according to spec

d-ronnqvist:

## Summary

This adds types to represent a hierarchy of HTML and a formatter that can format "void" elements (like `hr`, `<meta>`, or `<wbr>`) according to the specification and that preserves significant whitespace when "pretty" formatting the HTML output of a `<pre>` element (which semantically contains pre-formatted content, where whitespace is significant). The formatter's output is also consistent across macOS and other platforms.

The current `XMLNode` based solution for representing HTML has issues with all of these:
- it formats void elements like `<wbr />` instead of `<wbr>`
- it inserts newlines and indentation when "pretty" formatting `<pre>` elements.
- there are minor formatting differences in the output across macOS and other platforms. This has the implication that our tests have to preprocess the output to normalize these platform differences before they can be compared to the expected values.

So far there's nothing that uses this new hierarchy or the new formatter, except for the tests. I plan to integrate that over a few follow-up PRs. The formatter is less than 500 lines of code. 

Apart from the correctness improvements, the new types to represent HTML and the new formatter is between 7 to 13 times faster than the current solution at various microbenchmarks (creating a hierarchy, compactly formatting the output, pretty formatting the output, etc.) for various small and large test pages. However, this wasn't the primary goal for making these changes. The task of "rendering" the documentation as HTML is largely I/O-bound and isn't meaningfully impacted by these speedups.

I have several follow up PRs planned after this one to, amongst other things:
- Add factory methods for the remaining HTML elements and attributes (even the ones that we don't need yet) (https://github.com/swiftlang/swift-docc/pull/1578)
- Add a result builder to allow for a nicer syntax for defining HTML structures (https://github.com/swiftlang/swift-docc/pull/1584)
- Transition the existing XMLNode hierarchy to this HTMLNode hierarchy

## Dependencies

None.

## Testing

At this point, because this code isn't integrated yet, it can only be verified in isolation by formatting some HTML hierarchy (like in the added 500 lines of formatting tests).

## Checklist

- [x] Added tests


d-ronnqvist:
Hm. Something in this code is resulting in a compiler crash on Linux. 
> error: compile command failed due to signal 6 (use -v to see invocation)
> swift-frontend: /home/build-user/swift/lib/SILOptimizer/Mandatory/MoveOnlyAddressCheckerUtils.cpp:3107: bool (anonymous namespace)::GlobalLivenessChecker::compute(): Assertion `hadAnyErrorUsers == emittedDiagnostic' failed.

d-ronnqvist:
@swift-ci please test

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+    /// <dl><dt>range</dt><dd><p>The range in which to create a random value.</p></dd></dl>
+    /// ```
+    ///
+    /// - Parameters:
+    ///   - element: The HTML element to format compactly.
+    ///   - nextElementTag: The tag of the next element in the element's container, or `nil` if there are no more elements in the element's container.
+    private mutating func _compactFormat(_ element: consuming HTMLNode, nextElementTag: HTMLNode._Tag?) {
+        switch element._storage {
+        case .text(let text):
+            _format(text: consume text)
```
patshaughnessy: Similar question to the `&+ 1` operator: Is the performance advantage of using `consume` here (and `consuming` above) significant? These keywords make the code somewhat harder to read, and I wonder if the additional mental overhead is worth it?

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+            guard !shouldSelfClose else {
+                // Don't create an end tag if the start tag was already self-closing.
+                return
+            }
+            
+            // Contents
+            for index in contents.indices {
+                let node = contents[index]
+                
+                let nextIndex = index &+ 1
```
patshaughnessy: Why are we using this and not a simple `+ 1`? Is the performance advantage significant enough to warrant making the code harder to read? (In the PR description you mentioned that rendering HTML is I/O bound anyway.)

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+            //    ```
+            //    <p>
+            //      Some <b>bold</b> text before a
+            //      <a href="something">link</a>
+            //      and some <i>italicized</i> text after.
+            //    </p>
+            //    ```
+            
+            let hasAttributeReasonToPresentOnSeparateLine = switch tag {
+                case .a, .td, .th: attributes.count > 1
```
patshaughnessy, line 268: How did you decide which tags should behave this way? Is indenting handled by the HTML spec? Or just personal preference based on formatting existing HTML?

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+            guard !value.isEmpty else { continue }
+            
+            value.withUTF8 {
+                var remaining = $0[...]
+                
+                // If the formatter is configured to omit optional quotes around attribute values; check if this value _needs_ to be quoted.
+                guard !options.contains(.omitOptionalQuotesAroundAttributeValues) || remaining.contains(where: \.needsQuotingInHTMLAttribute) else {
+                    buffer.append(.init(ascii: "="))
+                    // If the value doesn't need quoting, the only escapable character is `&`.
+                    while let index = remaining.firstIndex(of: .init(ascii: "&")) {
```
patshaughnessy, line 400: Can you express this logic by calling the functions below? `needsEscapingInHTMLText` etc.?

If those functions ever change then this assertion might no longer be true:

> If the value doesn't need quoting, the only escapable character is `&`.

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+}
+
+private extension UTF8.CodeUnit {
+    /// A Boolean value that determines whether this UTF-8 code unit needs to be escaped if it appears in the textual contents of an HTML element.
+    var needsEscapingInHTMLText: Bool {
+        return self == .init(ascii: "&")
+            || self == .init(ascii: "<")
+    }
+    
+    /// A Boolean value that determines whether this UTF-8 code unit needs to be escaped if it appears in the attribute value of an HTML element.
```
patshaughnessy, line 444: Maybe add some comments here explaining that if this ever changes, the code above has some assumptions we should revisit.

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+    /// ```
+    var canPrettyPrintInline: Bool {
+        // Our pretty printer can include any "text-level semantic" tag inline
+        Self.a.rawValue <= self.rawValue && self.rawValue <= Self.wbr.rawValue
+    }
+    
+    /// Determines whether or not an element of this tag can omit its end tag when followed by the given `next` element in the same container.
+    func canOmitEndTag(whenFollowedBy next: Self?) -> Bool {
+        switch self {
+        case .p:
```
patshaughnessy, line 489: Are these behaviors specified in the HTML spec? Maybe include a link to that?

Sources/DocCHTML/HTMLNode+Tags.swift
```diff
@@ -0,0 +1,458 @@
...
+        case canvas
+        case noscript
+        case script
+        case slot
+        case template
+    }
+}
+
+extension HTMLNode._Tag {
+    var name: StaticString {
```
patshaughnessy: Reading this all the static strings are identical to each case's spelling. Would it make more sense use `enum _Tag: String` and return the `rawValue` here? You could have a different property to return a UInt8.

Or would Swift macros help here?

Sources/DocCHTML/CMakeLists.txt
```diff
@@ -8,6 +8,10 @@ See https://swift.org/LICENSE.txt for license information
 #]]
 
 add_library(DocCHTML STATIC
+  HTMLFormatter.swift
+  HTMLNode.swift
+  HTMLNode+Attributes.swift
+  HTMLNode+Tags.swift.swift
```
patshaughnessy:
```suggestion
  HTMLNode+Tags.swift
```

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+            true
+
+        case .thead:
+            next == .tbody || next == .tfoot
+            
+        case .tfoot:
+            next == nil
+            
+        case .tr:
+            next == .td || next == nil
```
patshaughnessy:
```suggestion
            next == .tr || next == nil
```

`td` can never be a sibling of `tr`. We want to avoid emitting `</tr>` between consecutive rows, right?

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -0,0 +1,530 @@
...
+    
+    /// Common indentation data, allocated once.
+    private static let _indentationData: [UInt8] = {
+        // Stop indenting further after 64 levels of indentation
+        var data = [UInt8](repeating: .init(ascii: " "), count: 128)
+        data[0] = .init(ascii: "\n")
+        return data
+    }()
+    
+    /// Private state that the
```
patshaughnessy: Missing the rest of the sentence here?

Sources/DocCHTML/HTMLNode.swift
```diff
@@ -0,0 +1,69 @@
...
+    var _storage: _Storage
+    
+    /// Creates a new HTML text node.
+    package static func text(_ text: consuming String) -> HTMLNode {
+        .init(_storage: .text(text))
+    }
+
+    static func _element(_ tag: _Tag, attributes: [Attribute] = [], contents: [HTMLNode]) -> HTMLNode {
+        assert(!tag.isVoid, "Cannot create an element using void tag '\(tag)'. Use `.voidElement(...)` instead.")
+        assert(attributes.count == Set(attributes.map { $0.nameForFormatting.description.lowercased() }).count, "All attribute names has to be case insensitively unique. This wasn't true for ...")
```
patshaughnessy: > This wasn't true for ...

... for what?

Sources/DocCHTML/HTMLNode+Tags.swift
```diff
@@ -0,0 +1,458 @@
...
+    ._element(.dd, attributes: attributes, contents: contents)
+}
+
+/// Creates a new `<dl>` element with the given attributes and contents.
+///
+/// The `<dl>` element semantically represents a description list consisting of zero or more name-value groups defined using pairs of `<dt>` and `<dd>` elements.
+/// A conforming `<dl>` element can only contain `<dt>` and `<dd>` elements.
+package func dl(attributes: [HTMLNode.Attribute] = [], contents: [HTMLNode]) -> HTMLNode {
+    assert(contents.allSatisfy { $0._tag == .dt || $0._tag == .dd }, "<dl> tags can only contain <dt> and <dd> tags")
+    return ._element(.dl, contents: contents)
```
patshaughnessy:
```suggestion
    return ._element(.dl, attributes: attributes, contents: contents)
```

patshaughnessy (changes requested):
Amazing work 👍 Very well designed and executed. I'm looking forward to reading the follow on PRs.

I just have a few questions and a few typos/tweaks for you.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Yes, thanks for catching this. This was supposed to check if the next element is a `.tr` tag.

Here's what the HTML specification says about omitting the end tag for the `tr` element:

> **Tag omission in text/html:**
> A `tr` element's end tag can be omitted if the `tr` element is immediately followed by another `tr` element, or if there is no more content in the parent element.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Yes, this is all based on the HTML spec. 

This information in the specification is represented for each tag so there's no specific page or section of the specification to link to—only to the specification itself.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: I can add that comment but it's practically very very unlikely to ever change.

This logic is based on the information in the HTML specification regarding which characters need to be escaped in a HTML attribute. The specification can't easily change this without having to consider backwards compatibility, so it's very unlikely that this logic will need to change.

Sources/DocCHTML/HTMLNode+Tags.swift
d-ronnqvist: I had thought that the compiler would make quicker comparisons with an integer value and that it would have easier access to the string's bytes if they were known static but I experimented a bit in Compiler Explorer and it turns out that Swift implements string enumerations kind of like this; with an integer tag and a static value.

The only thing that the integer raw value is used for is to determine if a certain tag is a text-level semantic but that can probably switch over all those values instead.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Like the long comment above this describes; there are no specific rules or even recommendations for how to pretty-format HTML output. It's basically all personal preference.

For each behavior of this formatter I've provided an example that hopefully highlights why we'd want to pretty-format the HTML like this. 

For example, a more naive implementation that adds a line break for each tag would format this paragraph of formatted text
```html
<p>
  Some <b>bold</b> and <i>italicized</i> text.
</p>
```
across 11 lines like this
```html
<p>
  Some
  <b>
    bold
  </b>
  and
  <i>
    italicized
  </i>
  text.
</p>
```
which I personally find to mostly get in the way and being rather hard to read and comprehend.

-----

I've made some tweaks to the formatter in the later PRs as I've encountered more content (mostly in tests) and have felt that I the current formatting looked slightly unexpected in the expected test string.

I expect that we'll continue to tweak these behaviors as we work more on features that output formatted HTML (and write tests expecting what that HTML output is).

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Yes. I meant for it to say "Private state that the formatter uses to pretty-print the HTML element hierarchy"

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: No single code expression has a significant performance advantage on its own but taken together many very small optimizations can start to add up.

Compared to the DocC's JSON decoder I've spent _very_ little time working on the performance of this formatter. I did a few comparisons of accumulating bytes into either a `String` value, a `Data` value, or a `[UInt8]` value and I benchmarked a couple variations of the indentation (for pretty printing) but other than that I mostly tried to avoid doing unnecessary work and tried to do a few smaller tricks that previously made a difference in the JSON decoder. 

This is one of those things. I haven't measured it here, but I did measure similar changes in the JSON decoder where I thought that it wouldn't make any difference but did (presumably because the extra annotations—in the case of the JSON decoder—helped the compiler avoid some copies)

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: I would answer this the same as I [answered](https://github.com/swiftlang/swift-docc/pull/1571/changes/BASE..d97ec69b435edb29c1b44f5f2522c2325815f6cd#r3546149275) the [question above](https://github.com/swiftlang/swift-docc/pull/1571/changes/BASE..d97ec69b435edb29c1b44f5f2522c2325815f6cd#r3544911726) (I read the feedback top-down and replied to the question above first)

For this line specifically; the `index` will exceed `contents.endIndex` long before it overflows a signed 64-bit value so the overflow check is unnecessary. Still, it's only saving _one_ branch instruction per element in the hierarchy but is saving unnecessary "work" and it's also only a 1 character difference in the source code 🤷‍♂️

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Like I mentioned in [the answer below](https://github.com/swiftlang/swift-docc/pull/1571/changes/BASE..d97ec69b435edb29c1b44f5f2522c2325815f6cd#r3545640047) those functions are very unlikely to change because the logic they represent is based on requirements from the HTML specification. 

If I were to update this to call `needsEscapingInHTMLText` I would probably structure it more like the while-loop below to avoid iterating over the value twice. The main reason why I didn't do that before was that I thought that this implementation was easier to follow (because it doesn't need to backtrack to insert the missing quotation mark if it finds a `"` after it has found a `&`).

Sources/DocCHTML/HTMLNode.swift
d-ronnqvist: Oh. I meant to print the repeated attributes here. I added this assertion in an earlier implementation when the attributes were `[String: String]`. 

It makes slightly less sense now that the attributes are structures with predefined names but I can still add the missing portion of the assertion message in case it ever catches a bug.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: In a formatting micro benchmark I couldn't measure any improvement from this, so I'll remove all the `consume` and `consuming` keywords in the formatter's implementation to make the code more readable.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Note that https://github.com/swiftlang/swift-docc/pull/1583 refines which tags this behavior applies to after finding more cases—from updating the expected output in tests—where I felt that this behavior was more legible. That said, I'd like to avoid the overhead of moving commits between open PRs so I prefer to leave that change in that PR. In the end the code will be the same regardless.

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: In a formatting micro benchmark I could measure a roughly 0.5% improvement from using `&+` instead of `+` in the 4 places it is used in this formatter. 

Because the formatter code also does a few buffer pointer operations I don't think the use of the overflow addition operator has a major impact on the overall understandability of the code. I'll keep them in for now and we can revisit it that becomes a readability problem in the future.

Sources/DocCHTML/HTMLNode.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1571/commits/80cfeb3de2b1495908e24b94831f4953ba046f88

Sources/DocCHTML/HTMLNode+Tags.swift
d-ronnqvist: Updated to use a String enum in https://github.com/swiftlang/swift-docc/pull/1571/commits/8ad4db34e9e0bba07e0fe8102da2dd450c2dc4a0

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Removed in https://github.com/swiftlang/swift-docc/pull/1571/commits/71427b4a6cb92daa1672e1bb0784cbc380515c49

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Added in https://github.com/swiftlang/swift-docc/pull/1571/commits/a6d9a1db8344288fa3173d801331c3508ad333d4

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Added more comments and a link in https://github.com/swiftlang/swift-docc/pull/1571/commits/a6d9a1db8344288fa3173d801331c3508ad333d4

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: Added more comments and links to the specification in https://github.com/swiftlang/swift-docc/pull/1571/commits/a6d9a1db8344288fa3173d801331c3508ad333d4

Sources/DocCHTML/HTMLFormatter.swift
d-ronnqvist: I added a couple comments about the overflowing in https://github.com/swiftlang/swift-docc/pull/1571/commits/a6d9a1db8344288fa3173d801331c3508ad333d4

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -319,19 +318,24 @@ package struct HTMLFormatter {
...
         }
     }
     
-    /// Append the static string to the formatter's buffer.
+    /// Appends the static string to the formatter's buffer.
     private mutating func _append(_ string: StaticString) {
         string.withUTF8Buffer { buffer.append(contentsOf: $0) }
     }
+    /// Appends the tag's string representation to the formatter's buffer.
+    private mutating func _append(_ tag: HTMLNode._Tag) {
```
patshaughnessy, line 330: Nice - thanks for switching to the string enum type, and avoiding that long enum statement. 👍

patshaughnessy (approved):
Thanks for all the changes and fixes!

d-ronnqvist:
@swift-ci please test

PR #1572, opened 2026-06-29
Fix Python formatting

snprajwal:
GitHub CI runs a Python formatting check with flake8 as part of the soundness check suite. The check was buggy due to flake8's default ignore list being overridden by the config in the workflow definition. This has been fixed in v0.0.12 [1], and the check can now be enabled. This patch removes the configuration value that disabled the check, and formats `build-script-helper.py` as per the flake8 configuration in swiftlang/github-workflows [2].

[1]: https://github.com/swiftlang/github-workflows/pull/279
[2]: https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/configs/.flake8

### DIFF
--- a/build-script-helper.py
+++ b/build-script-helper.py
@@ -3,7 +3,7 @@
 """
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2022 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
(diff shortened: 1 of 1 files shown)


snprajwal:
@swift-ci please test

snprajwal:
@swift-ci please test

d-ronnqvist (commented):
Question: is it possible to configure this Python linter to allow 4 space indentation and longer lines so that we don't need to reformat all this code?

snprajwal:
I'm guessing you meant to say 2-space indentation, since this PR updates it to use 4 spaces instead. We can use our own flake8 configuration, but the PEP8 style guide [recommends using 4 spaces](https://peps.python.org/pep-0008/#indentation), and we'd be going against that.

It also [recommends limiting lines to 79 characters](https://peps.python.org/pep-0008/#maximum-line-length), but [the configuration used by swiftlang](https://github.com/swiftlang/github-workflows/blob/main/.github/workflows/configs/.flake8#L10) adds an extra 10% to allow up to 88 characters.

I would be in favour of just using the swiftlang configuration. If you feel that the reformat adds noise, I can add the commit to `.git-blame-ignore-revs` to exclude it when viewing the commit history for the script.

anferbui (approved):
This looks good to me, the Python lint check is passing and the reasoning about not configuring our own rules makes sense to me.

d-ronnqvist:
It's a bit unfortunate if we lose the ability to use Show Last Change for Line or similar features in this script. If we can do something to preserve that then that would be good. That said, this script is rarely modified (at most a few times per year) so a person can still read the entire history if they need to understand why a certain change was made in the first place (and not why it was reformatted).

snprajwal:
We can exclude the formatting change from the blame, no problems there. Here's what I'll do:

- This PR has proven that the Python lint check passes, so I'm going to remove the change that enables the check, and merge this PR with only the formatting changes. This is so we don't lose the changes in `pull_request.yml` when we do the step below.
- Once we merge, I can add the squashed commit SHA to `.git-blame-ignore-revs`, and open a separate PR with that + the change to `pull_request.yml` that enables the check. This way, the formatting gets excluded from the blame, but we correctly retain the change to the GHA manifest.

snprajwal:
@swift-ci test

PR #1573, opened 2026-06-30
Drop unnamed symbols when loading symbol graph

snprajwal:
Bug/issue #, if applicable: rdar://180902381

## Summary

Clang can sometimes erroneously emit anonymous structs and unions without a title in the symbol graph. This typically occurs for anonymous types nested within public types, making them technically public but functionally private since they cannot be referenced in any way by any consumers of that header. This patch updates the symbol graph loader to drop these symbols to avoid a variety of issues caused by registering them.

## Dependencies

N/A

## Testing

1. Create a symbol graph containing an unnamed struct or union. Optionally, add a relationship making this type a child of a named container type.
2. Build documentation with this symbol graph in an archive.
3. Notice that the root page (or the container page if the type was nested) breaks.
4. With this patch, the page renders correctly.

## Checklist

- [x] Added tests


snprajwal:
@swift-ci please test

d-ronnqvist (approved):

PR #1574, opened 2026-06-30
Enable unacceptable language check

snprajwal:
The soundness check suite provides an unacceptable language check. This was previously disabled due to the default list containing a number of words that DocC uses in various places in a non-offensive and contextually appropriate manner. This patch enables this check and restricts the list of words to the ones checked by DocC's own unacceptable language check. It also updates DocC's check to identify usage of the word "master".

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -35,9 +35,10 @@ jobs:
       license_header_check_project_name: "Swift"
       # The license header check is disabled as it requires a large number of changes.
       license_header_check_enabled: false
-      # The unacceptable language check is disabled due to conflict with
-      # DocC's own unacceptable language check tests.
-      unacceptable_language_check_enabled: false
+      # There are some words in typical programming jargon that are a part of the default list.
+      # The list of words checked by this job is restricted below to prevent raising issues for
+      # words that are contextually acceptable to use in this repository.
+      unacceptable_language_check_word_list: "blacklist whitelist slave master"  # ignore-unacceptable-language
       # The format check is disabled as this project does not conform to swift-format.
       # More discussion is needed before enabling this check.
       format_check_enabled: false
--- a/.unacceptablelanguageignore
+++ b/.unacceptablelanguageignore
@@ -1,0 +1,2 @@
+Sources/SwiftDocC/Checker/Checkers/NonInclusiveLanguageChecker.swift
+Tests/SwiftDocCTests/Checker/Checkers/NonInclusiveLanguageCheckerTests.swift
--- a/Dockerfile
+++ b/Dockerfile
@@ -9,7 +9,7 @@
 FROM swift:5.9-focal
 
 # Set up the current build user in the same way done in the Swift.org CI system: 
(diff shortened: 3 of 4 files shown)


.github/workflows/pull_request.yml
```diff
@@ -35,9 +35,10 @@ jobs:
       license_header_check_project_name: "Swift"
       # The license header check is disabled as it requires a large number of changes.
       license_header_check_enabled: false
-      # The unacceptable language check is disabled due to conflict with
-      # DocC's own unacceptable language check tests.
-      unacceptable_language_check_enabled: false
+      # There are some words in typical programming jargon that are a part of the default list.
+      # The list of words checked by this job is restricted below to prevent raising issues for
+      # words that are contextually acceptable to use in this repository.
+      unacceptable_language_check_word_list: "blacklist whitelist slave master"  # ignore-unacceptable-language
```
d-ronnqvist, line 41: We currently check for this unacceptable non-inclusive language in [the `bin/check-source` script](https://github.com/swiftlang/swift-docc/blob/main/bin/check-source#L24-L38). 

Is the plan for this check to replace that in the long run?

d-ronnqvist: Do you know if this is adding to the default list or overriding the default list of unacceptable language?

Also, out of curiosity; do you know what the default list of unacceptable language is? (I'm wondering if it has more terms that we could also check for)

QuietMisdreavus: The default list is set [in the `swiftlang/github-workflows` configuration](https://github.com/swiftlang/github-workflows/blob/07936268efedce0a6650c75a00150b820f53f384/.github/workflows/soundness.yml#L77):

```yaml
unacceptable_language_check_word_list:
  type: string
  description: "List of unacceptable words. Defaults to a sensible list of words."
  default: "blacklist whitelist slave master sane sanity insane insanity kill killed killing hang hung hanged hanging"  # ignore-unacceptable-language

```

Going from how i understand the workflow definitions to operate, setting this configuration value replaces the default word list instead of adding to it.

As for whether it should replace the non-inclusive language check in `bin/check-source`, i think it serves the same purpose as far as CI is concerned, but the script itself is hosted in another repo. I feel like if we want people to be able to check for it locally we should have an equivalent check.

snprajwal: As Vera mentioned, it overrides the default. Our script performs an unacceptable language check and header validation, both of which are present in the soundness check on CI. It would be nice to move to the org-wide checks instead of our own script, but I haven't thought about how the local workflow is impacted. AFAIK, it's mostly trivial to spin up the CI job locally via a container, so maybe we can repurpose the existing script to do that instead and remove any custom logic there once we start using the CI checks properly (the header checks are still disabled at the moment)

d-ronnqvist: I see. In that case I think it makes sense to keep the non-inclusive language check in `bin/check-source` to support local workflows, at least for this PR. That said, we haven't had many issues with non-inclusive language so it might be fine for such issues to not be surfaced before opening the PR. Still, that's something we can revisit for a later PR.

d-ronnqvist (approved):

snprajwal:
@swift-ci test

PR #1576, opened 2026-07-02
Split input and output `FileManagerProtocol`s

al45tair:
In order to output compressed archives, we need first to split `FileManagerProtocol` into `ReadOnlyFileManagerProtocol` and (read-write) `FileManagerProtocol`, then we need to add `outputFileSystem` arguments to various things in the codebase, to reflect the fact that we are now reading from one place and writing to another.

rdar://181216501

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
@@ -43,6 +43,8 @@ public struct ConvertAction: AsyncAction {
     let sourceRepository: SourceRepository?
     
     private var fileManager: any FileManagerProtocol
+    private var outputFileManager: any FileManagerProtocol
+
     private let temporaryDirectory: URL
     
     private let diagnosticWriterOptions: (formatting: DiagnosticFormattingOptions, baseURL: URL)
@@ -94,6 +96,7 @@ public struct ConvertAction: AsyncAction {
         fileManager: any FileManagerProtocol = FileManager.default,
         temporaryDirectory: URL,
         outputFormat: Docc.Convert.OutputFormat = .json,
+        outputFileManager: (any FileManagerProtocol)? = nil,
         documentationCoverageOptions: DocumentationCoverageOptions = .noCoverage,
         bundleDiscoveryOptions: BundleDiscoveryOptions = .init(),
         diagnosticLevel: String? = nil,
@@ -121,6 +124,7 @@ public struct ConvertAction: AsyncAction {
         self.outputFormat = outputFormat
         self.buildLMDBIndex = buildIndex
         self.fileManager = fileManager
+        self.outputFileManager = outputFileManager ?? fileManager
         self.temporaryDirectory = temporaryDirectory
         self.documentationCoverageOptions = documentationCoverageOptions
         self.transformForStaticHosting = transformForStaticHosting
@@ -213,11 +217,6 @@ public struct ConvertAction: AsyncAction {
     /// A block of extra work that tests perform to affect the time it takes to convert documentation
     var _extraTestWork: (() async -> Void)?
 
-    /// The `Indexer` type doesn't work with virtual file systems.
-    ///
-    /// Tests that don't verify the contents of the navigator index can set this to `true` so that they can use a virtual, in-memory, file system.
-    var _completelySkipBuildingIndex: Bool = false
-    
     /// Converts each eligible file from the source documentation bundle,
     /// saves the results in the given output alongside the template files.
     public func perform(logHandle: inout LogHandle) async throws -> ActionResult {
@@ -226,25 +225,21 @@ public struct ConvertAction: AsyncAction {
     
     func perform(logHandle: inout LogHandle) async throws -> (ActionResult, DocumentationContext) {
         // FIXME: Use `defer` again when the asynchronous defer-statement miscompilation (rdar://137774949) is fixed.
-        let temporaryFolder: URL
-        switch outputFormat {
-        case .json:
-            temporaryFolder = try createTempFolder(with: htmlTemplateDirectory)
-        case .experimentalHTML:
-            temporaryFolder = try createTempFolder(with: nil)
-            for file in DocCHTML.StaticResources.allFiles {
-                try fileManager.createFile(at: temporaryFolder.appendingPathComponent(file.filename), contents: file.data)
-            }
-        }
+        let temporaryFolder = try Self.createUniqueDirectory(
+            inside: temporaryDirectory,
+            template: nil,
+            fileManager: outputFileManager
+        )
         
         do {
             let result = try await _perform(logHandle: &logHandle, temporaryFolder: temporaryFolder)
+
             diagnosticEngine.flush()
-            try? fileManager.removeItem(at: temporaryFolder)
+            try? outputFileManager.removeItem(at: temporaryFolder)
             return result
         } catch {
             diagnosticEngine.flush()
-            try? fileManager.removeItem(at: temporaryFolder)
+            try? outputFileManager.removeItem(at: temporaryFolder)
             throw error
         }
     }
@@ -255,6 +250,25 @@ public struct ConvertAction: AsyncAction {
             signposter.endInterval("Convert", convertSignpostHandle)
         }
         
+        let generateInFolder: URL = temporaryFolder
+        let generateInFileManager = outputFileManager
+
+        // Populate the directory with the initial files
+        switch outputFormat {
+            case .json:
+                if let htmlTemplateDirectory {
+                    try fileManager.copyItem(
+                        at: htmlTemplateDirectory,
+                        to: generateInFolder,
+                        on: generateInFileManager
+                    )
+                }
+            case .experimentalHTML:
+                for file in DocCHTML.StaticResources.allFiles {
+                    try generateInFileManager.createFile(at: generateInFolder.appendingPathComponent(file.filename), contents: file.data)
+                }
+        }
+
         // Add the default diagnostic console writer now that we know what log handle it should write to.
         if !diagnosticEngine.hasConsumer(matching: { $0 is DiagnosticConsoleWriter }) {
             diagnosticEngine.add(
@@ -283,12 +297,12 @@ public struct ConvertAction: AsyncAction {
         // FIXME: Use `defer` here again when the miscompilation of this asynchronous defer-statement (rdar://137774949) is fixed.
 //        let temporaryFolder = try createTempFolder(with: htmlTemplateDirectory)
 //        defer {
-//            try? fileManager.removeItem(at: temporaryFolder)
+//            try? generateInFileManager.removeItem(at: temporaryFolder)
 //        }
 
         let indexHTML: URL?
         if let htmlTemplateDirectory, outputFormat == .json {
-            let indexHTMLUrl = temporaryFolder.appendingPathComponent(
+            let indexHTMLUrl = generateInFolder.appendingPathComponent(
                 HTMLTemplate.indexFileName.rawValue,
                 isDirectory: false
             )
@@ -304,10 +318,10 @@ public struct ConvertAction: AsyncAction {
                 
                 // A hosting base path was provided which means we need to replace the standard
                 // 'index.html' file with the transformed one.
-                try fileManager.createFile(at: indexHTMLUrl, contents: data)
+                try generateInFileManager.createFile(at: indexHTMLUrl, contents: data)
             }
             
-            let indexHTMLTemplateURL = temporaryFolder.appendingPathComponent(
+            let indexHTMLTemplateURL = generateInFolder.appendingPathComponent(
                 HTMLTemplate.templateFileName.rawValue,
                 isDirectory: false
             )
@@ -315,26 +329,27 @@ public struct ConvertAction: AsyncAction {
             // Delete any existing 'index-template.html' file that
             // was copied into the temporary output directory with the
             // HTML template.
-            try? fileManager.removeItem(at: indexHTMLTemplateURL)
+            try? generateInFileManager.removeItem(at: indexHTMLTemplateURL)
         } else {
             indexHTML = nil
         }
         
         let coverageAction = CoverageAction(
             documentationCoverageOptions: documentationCoverageOptions,
-            workingDirectory: temporaryFolder,
-            fileManager: fileManager)
+            workingDirectory: generateInFolder,
+            fileManager: generateInFileManager)
 
-        let indexer = _completelySkipBuildingIndex ? nil : try Indexer(outputURL: temporaryFolder, bundleID: inputs.id)
+        let indexer = try Indexer(outputURL: generateInFolder, fileManager: generateInFileManager, bundleID: inputs.id)
 
         let registerInterval = signposter.beginInterval("Register", id: signposter.makeSignpostID())
         let context = try await DocumentationContext(bundle: inputs, dataProvider: dataProvider, diagnosticEngine: diagnosticEngine, configuration: configuration)
         signposter.endInterval("Register", registerInterval)
         
         let outputConsumer = ConvertFileWritingConsumer(
-            targetFolder: temporaryFolder,
+            targetFolder: generateInFolder,
             bundleRootFolder: rootURL,
             fileManager: fileManager,
+            outputFileManager: generateInFileManager,
             context: context,
             indexer: indexer,
             enableCustomTemplates: experimentalEnableCustomTemplates,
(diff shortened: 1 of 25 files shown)


Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -213,11 +217,6 @@ public struct ConvertAction: AsyncAction {
     /// A block of extra work that tests perform to affect the time it takes to convert documentation
     var _extraTestWork: (() async -> Void)?
 
-    /// The `Indexer` type doesn't work with virtual file systems.
-    ///
-    /// Tests that don't verify the contents of the navigator index can set this to `true` so that they can use a virtual, in-memory, file system.
-    var _completelySkipBuildingIndex: Bool = false
```
d-ronnqvist, line 219: Does this mean that the `Indexer` can write to a virtual file system now? 👀

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -255,6 +250,25 @@ public struct ConvertAction: AsyncAction {
             signposter.endInterval("Convert", convertSignpostHandle)
         }
         
+        let generateInFolder: URL = temporaryFolder
+        let generateInFileManager = outputFileManager
```
d-ronnqvist, line 254: Question: Does this need to capture a local variable to the output file manager? Otherwise I find the original names easier to understand.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -453,35 +464,47 @@ public struct ConvertAction: AsyncAction {
...
         // FIXME: We no longer output a diagnostics file in the output. We can remove the `emitDigest` check below.
         if !didEncounterError || emitDigest {
-            try moveOutput(from: temporaryFolder, to: targetDirectory)
+            try signposter.withIntervalSignpost("Move output") {
+                try Self.moveOutput(from: generateInFolder, to: targetDirectory, fileManager: generateInFileManager)
+            }
         }
 
+        // Stop the "total time" metric here. While moveOutput isn't interesting, compression is.
+        benchmark(end: totalTimeMetric)
```
d-ronnqvist: Moving this metric creates a breaking point where past builds can no longer be compared to newer builds. 

If we want to measure the compression then we may need to add a new metric for it.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
```diff
@@ -18,17 +18,19 @@ class JSONEncodingRenderNodeWriter {
...
     private let transformForStaticHostingIndexHTML: URL?
     private let fileManager: any FileManagerProtocol
+    private let outputFileManager: any FileManagerProtocol
     private let renderReferenceCache = RenderReferenceCache([:])
     
     /// Creates a writer object that write render node JSON into a given folder.
     ///
     /// - Parameters:
     ///   - targetFolder: The folder to which the writer object writes the files.
     ///   - fileManager: The file manager with which the writer object writes data to files.
```
d-ronnqvist, line 27: minor: This documentation is no longer correct.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
```diff
@@ -18,17 +18,19 @@ class JSONEncodingRenderNodeWriter {
...
-    init(targetFolder: URL, fileManager: any FileManagerProtocol, transformForStaticHostingIndexHTML: URL?) {
+    init(targetFolder: URL, fileManager: any FileManagerProtocol, outputFileManager: any FileManagerProtocol, transformForStaticHostingIndexHTML: URL?) {
```
d-ronnqvist: It's a bit hard to tell from the diff below. Does the writer ever use the `fileManager` property? If it doesn't then we only need one file manager and we could keep the existing parameter.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
```diff
@@ -70,23 +72,16 @@ class JSONEncodingRenderNodeWriter {
...
-        do {
-            try fileManager._copyItem(at: indexHTML, to: htmlTargetFileURL)
-        } catch let error as NSError where error.code == NSFileWriteFileExistsError {
-            // We already have an 'index.html' file at this path. This could be because
-            // we're writing to an output directory that already contains built documentation
-            // or because we we're given bad input such that multiple documentation pages
-            // have the same path on the filesystem. Either way, we don't want this to error out
-            // so just remove the destination item and try the copy operation again.
-            try fileManager.removeItem(at: htmlTargetFileURL)
-            try fileManager._copyItem(at: indexHTML, to: htmlTargetFileURL)
```
d-ronnqvist, line 88: Question: Where did this workaround move? I can't find it in the diff. As far as I know, we still need this workaround somewhere.

Sources/DocCCommandLine/Action/Actions/Merge/MergeAction+SynthesizedLandingPage.swift
```diff
@@ -34,7 +34,7 @@ extension MergeAction {
         }
     }
     
-    func readRootNodeRenderReferencesIn(dataDirectory: URL) throws -> RootRenderReferences {
+    func readRootNodeRenderReferencesIn(dataDirectory: URL, on fileManager: any ReadOnlyFileManagerProtocol) throws -> RootRenderReferences {
```
d-ronnqvist: minor: This may just be me not being a native speaker but I find the argument label "on" to be a bit surprising and ungrammatical for a file manager parameter. I would find that something like "using" would read better at the call site.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
```diff
@@ -495,6 +502,9 @@ extension NavigatorIndex {
         /// The output URL.
         public let outputURL: URL
         
+        /// The filesystem in which to work.
+        public let fileManager: any FileManagerProtocol
```
d-ronnqvist, line 506: I don't think we want this to be public API considering that this file manager isn't used for the LMDB files.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
```diff
@@ -22,30 +22,78 @@ package import Foundation
...
+/// a file-system, it performs actions like creating new files, copying, renaming
+/// and deleting files, and organizing them in directories.
+///
+/// The Cocoa `FileManager` type is the default implementation of that protocol
+/// that manages files stored on disk.
+///
+/// Should you need a file system with a different storage, create your own
+/// protocol implementations to manage files in memory,
+/// on a network, in a database, or elsewhere.
+public protocol FileManagerProtocol: ReadOnlyFileManagerProtocol {
```
d-ronnqvist, line 87: I don't think we want the file manager abstraction to be public API. At least that's something that we need to have a more careful review about.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
```diff
@@ -88,38 +125,94 @@ package protocol FileManagerProtocol: DataProvider {
...
+
     /// Returns a Boolean value that indicates whether a directory exists at a specified path.
-    package func directoryExists(atPath path: String) -> Bool {
+    public func directoryExists(atPath path: String) -> Bool {
         var isDirectory = ObjCBool(booleanLiteral: false)
         let fileExistsAtPath = fileExists(atPath: path, isDirectory: &isDirectory)
         return fileExistsAtPath && isDirectory.boolValue
     }
+
+    public func copyItem(at source: URL, to destination: URL, on otherFileManager: any FileManagerProtocol) throws {
```
d-ronnqvist, line 158: Like mentioned above: it may just be me not being a native speaker but I find the argument label "using" reads better than the label "on" for a file manager parameter.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
```diff
@@ -88,38 +125,94 @@ package protocol FileManagerProtocol: DataProvider {
...
+            files:       Array( allContents[..<partitionIndex] ),
+            directories: Array( allContents[partitionIndex...] )
+        )
+    }
+
+}
+
+extension FileManagerProtocol {
+
+    public func moveItem(at source: URL, to destination: URL, on otherFileManager: any FileManagerProtocol) throws {
```
d-ronnqvist, line 195: Like mentioned above: it may just be me not being a native speaker but I find the argument label "using" reads better than the label "on" for a file manager parameter.

d-ronnqvist (changes requested):
I'm generally in favor of splitting the file manager protocol into reading and writing pieces. 

That said: if I understand these changes correctly it both breaks file copying behavior by removing the previous workaround and it redefines a metric such that past versions can't be compared to new versions. 

I also think that we want to keep the file manager protocol(s) as a package-level implementation detail so that we are free to change them without worrying about breaking clients.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
al45tair: Yes. See the NavigatorIndex.swift changes later on.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
al45tair: It will in a later PR; I'm trying to break these changes up into chunks we can easily review as separate PRs and didn't want to rename all of the uses of these variable names only to move them back again in a subsequent PR.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
al45tair: Not sure how big a deal this really is. We are now including the time taken to move the output to the final location. Is that really a problem?

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
al45tair: It hasn't moved; it's just been rewritten so that it works by checking for the existence of the file and explicitly removing it, instead of relying on catching an error.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
al45tair: I forget why it ended up having to be `public`. I think because it's being used from another package within the DocC source tree. I'll have to take a look and see what the reason was.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
al45tair: OK, having thought about it, we don't need to move this in this PR anyway, and maybe you're right that we should have a separate metric for compression time. I'll undo this change.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
al45tair: I think we can actually remove the `outputFileManager`, as you point out below.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
al45tair: Let's remove `outputFileManager`; it doesn't seem to be necessary here.

Sources/DocCCommandLine/Action/Actions/Merge/MergeAction+SynthesizedLandingPage.swift
al45tair: Let's use `using` instead.

Sources/SwiftDocC/Indexing/Navigator/NavigatorIndex.swift
al45tair: That's a fair point. I didn't do anything about using LMDB files from non-default file managers.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
al45tair: I used `on` here because it's copying from `source` on the filesystem pointed at by `self` to `destination` on the filesystem pointed by `otherFileManager`. I agree that it's not entirely satisfactory because the parameter is a "file manager", and "on" a file manager seems to make less sense than "using" a file manager, _but_ "using" makes it seem like that's the only file manager being used here, which isn't true.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
al45tair: The same logic here applies as applies to `copyItem`. `on` I think properly conveys the fact that the _destination_ is `on` the filesystem addressed by the `otherFileManager`, whereas `using` would make it seem like that was the only file manager responsible for both URLs, I think.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
al45tair: OK, it needs to be `public` so that it can be used in:

- `Benchmark/Metrics/OutputSize.swift`, in the public `init`
- `Indexing/Navigator/NavigatorIndex.swift`, again in the public `init`
- `Infrastructure/Diagnostics/DiagnosticConsoleWriter.swift`, in the `formattedDescription` method.
- `Infrastructure/Input Discovery/DocumentationInputsProvider.swift` in the package `init`.
- `Model/Rendering/HTML/HTMLRenderer+FullPage.swift` in `prepareForFullPage`.

Making it `package` instead solves the latter three, but still creates a problem in `Benchmark` and `Indexing`.

d-ronnqvist: In that case we need to figure out a different solution for the first two. In other places we introduce a second `package` access initializer have have the `public` initializer pass `FileManager.default` to the `package` access initializer.

Sources/DocCCommandLine/Action/Actions/Convert/JSONEncodingRenderNodeWriter.swift
d-ronnqvist: If I understand correctly that seems like the close description of this anti-pattern from the note callout in [the `fileExists(atPath:)` documentation](https://developer.apple.com/documentation/foundation/filemanager/fileexists(atpath:)):

> Attempting to predicate behavior based on the current state of the file system or a particular file on the file system is not recommended. Doing so can cause odd behavior or race conditions. It’s far better to attempt an operation (such as loading a file or creating a directory), check for errors, and handle those errors gracefully than it is to try to figure out ahead of time whether the operation will succeed. For more information on file-system race conditions, see [Race Conditions and Secure File Operations](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Articles/RaceConditions.html#//apple_ref/doc/uid/TP40002585) in [Secure Coding Guide](https://developer.apple.com/library/archive/documentation/Security/Conceptual/SecureCodingGuide/Introduction.html#//apple_ref/doc/uid/TP40002415).

al45tair: There's no security-relevant race condition here, so that admonishment is a bit pointless.

Both

```swift
do {
  try fileManager._copyItem(at: indexHTML, to: htmlTargetFileURL)
} catch let error as NSError where error.code == NSFileWriteFileExistsError {
  try fileManager.removeItem(at: indexHTML)
  try fileManager._copyItem(at: indexHTML, to: htmlTargetFileURL)
}
```

and

```swift
if fileManager.fileExists(at: indexHTML) {
  try fileManager.removeItem(at: indexHTML)
}
try fileManager._copyItem(at: indexHTML, to: htmlTargetFileURL)
```

are racy, and there's nothing you can do about that. For instance, in the `do... catch` version, if some malicious party puts a file at `indexHTML`, then waits a tiny amount and does it again, the _second_ `_copyItem` will fail. That's _exactly_ the same kind of failure as someone doing it once, which is what trips up the simpler `fileExists`-based code.

The reason I changed it is that it relies on the file manager specifically throwing an `NSError` with `error.code` set to `NSFileWriteFileExistsError`. That's what Foundation's `FileManager` does, or at least _might_ do (assuming there's no other error), but in general I don't think we should force `FileManagerProtocol` implementations to throw `NSError`s specifically.

Sources/SwiftDocC/Utility/FileManagerProtocol.swift
al45tair: Hmmm. That might work.

al45tair:
Looks like my rebase has gone wrong somehow :-(

PR #1578, opened 2026-07-02
[2/n] Define accessors for the remaining HTML tags and attributes

d-ronnqvist:
Bug/issue #, if applicable: rdar://165755530

## Summary

This is a follow up to https://github.com/swiftlang/swift-docc/pull/1571 ~(still open)~ that defines factory methods for all remaining HTML tags and _almost_ all remaining HTML attributes.

Compared to https://github.com/swiftlang/swift-docc/pull/1571, this can be thought of as adding all the repetitive boilerplate that both:
- avoids relying on string-based API for the HTML hierarchy
- provides brief documentation of the semantic meaning of each HTML tag and HTML attribute.

## Dependencies

This depends on ~and includes all the changes of~:

- https://github.com/swiftlang/swift-docc/pull/1571

## Testing

At this point, because this code isn't integrated yet, it can only be verified in isolation by formatting some HTML hierarchy (like in the formatting tests from https://github.com/swiftlang/swift-docc/pull/1571).

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- ~[ ] Added tests~
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1583, opened 2026-07-07
[3/n] Minimally integrate custom HTML formatter 

d-ronnqvist:

## Summary

This is a follow up to https://github.com/swiftlang/swift-docc/pull/1571 ~(still open)~ and https://github.com/swiftlang/swift-docc/pull/1578 ~(still open)~ that minimally integrated the custom HTML formatter by traversing the XMLNode hierarchy and creating an HTMLNode hierarchy that it later formats.

This is clearly not the intended long-term solution but rather a step towards replacing the XMLNode hierarchy with a HTMLNode hierarchy. The reason for this short-term integration solution is to try and break up the work into smaller pieces that can be reviewed and iterated incrementally.

## Dependencies

This depends on and includes all the changes of:

- https://github.com/swiftlang/swift-docc/pull/1571
- https://github.com/swiftlang/swift-docc/pull/1578

## Testing

1. Build some documentation and pass `--output-format experimental-html-for-development` to DocC. 
2. Inspect the HTML files in DocC's output. It should be formatted using the custom formatter. 
  - An easy way to verify this is that the various `meta` elements end with `>` instead of `/>`
3. Build the same documentation with the same flags but with `DOCC_JSON_PRETTYPRINT="YES"` defined
4. Inspect the HTML files in DocC's output. It should be pretty printed (this wasn't the case before because the XMLNode formatter broke `pre` tags (where whitespace is significant).

## Checklist

- [x] ~Added~ Updated tests
- ~[] Updated documentation if necessary~

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
@@ -348,6 +348,7 @@ public struct ConvertAction: AsyncAction {
             htmlConsumer = try FullPageHTMLContentConsumer(
                 targetFolder: temporaryFolder,
                 fileManager: fileManager,
+                prettyPrint: shouldPrettyPrintOutputJSON, // Use the same configuration as the JSON output to make it convenient for the developer.
                 customHeader: experimentalEnableCustomTemplates ? inputs.customHeader : nil,
                 customFooter: experimentalEnableCustomTemplates ? inputs.customFooter : nil
             )
--- a/Sources/DocCCommandLine/Action/Actions/Convert/FileWritingHTMLContentConsumer.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/FileWritingHTMLContentConsumer.swift
@@ -38,7 +38,7 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
             
             // Ensure that the index.html file has at least a `<head>` and a `<body>`.
             guard var beforeEndOfHead  = content.utf8.firstRange(of: "</head>".utf8)?.lowerBound,
-                  var afterStartOfBody = content.range(of: "<body[^>]*>", options: .regularExpression)?.upperBound
+                  var afterStartOfBody = content.range(of: "<body[^>]*>\n?", options: .regularExpression)?.upperBound
             else {
                 struct MissingRequiredTagsError: DescribedError {
                     let errorDescription = "Missing required `<head>` and `<body>` elements in \"index.html\" file."
@@ -55,10 +55,10 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
             {
                 titleReplacementRange = titleStart ..< titleEnd
             } else {
-                content.insert(contentsOf: "<title></title>", at: beforeEndOfHead)
-                content.utf8.formIndex(&beforeEndOfHead,  offsetBy: "<title></title>".utf8.count)
-                content.utf8.formIndex(&afterStartOfBody, offsetBy: "<title></title>".utf8.count)
-                let titleInside = content.utf8.index(beforeEndOfHead, offsetBy: -"</title>".utf8.count)
+                content.insert(contentsOf: "<title></title>\n", at: beforeEndOfHead)
+                content.utf8.formIndex(&beforeEndOfHead,  offsetBy: "<title></title>\n".utf8.count)
+                content.utf8.formIndex(&afterStartOfBody, offsetBy: "<title></title>\n".utf8.count)
+                let titleInside = content.utf8.index(beforeEndOfHead, offsetBy: -"</title>\n".utf8.count)
                 titleReplacementRange = titleInside ..< titleInside
             }
             
@@ -67,7 +67,7 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
             {
                 contentReplacementRange = noScriptStart ..< noScriptEnd
             } else {
-                content.insert(contentsOf: "<noscript></noscript>", at: afterStartOfBody)
+                content.insert(contentsOf: "<noscript></noscript>\n", at: afterStartOfBody)
                 let noScriptInside = content.utf8.index(afterStartOfBody, offsetBy: "<noscript>".utf8.count)
                 contentReplacementRange = noScriptInside ..< noScriptInside
             }
@@ -92,7 +92,7 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
                 let metaDescription = XMLNode.element(named: "meta", attributes: ["name": "description", "content": plainDescription])
                 copy.replaceSubrange(descriptionReplacementRange, with: metaDescription.rendered(prettyPrinted: prettyPrint))
             }
-            copy.replaceSubrange(titleReplacementRange,   with: title)
+            copy.replaceSubrange(titleReplacementRange, with: title)
             
             return copy
         }
@@ -153,7 +153,15 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
 
 private extension XMLNode {
     func rendered(prettyPrinted: Bool) -> String {
-        if prettyPrinted {
+        if let htmlNode = HTMLNode(from: self) {
+            let data = HTMLFormatter.format(htmlNode, options: prettyPrinted ? .prettyPrint : [])
+            return String(decoding: data, as: UTF8.self)
+        }
+         
+        assertionFailure("Failed to convert XMLNode \(name ?? "<no tag>") to an HTMLNode")
+        
+        // Fallback to the XMLNode string formatting for now.
+        return if prettyPrinted {
             xmlString(options: [.nodePrettyPrint, .nodeCompactEmptyElement])
         } else {
             xmlString(options: .nodeCompactEmptyElement)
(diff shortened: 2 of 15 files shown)


Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -348,6 +348,7 @@ public struct ConvertAction: AsyncAction {
             htmlConsumer = try FullPageHTMLContentConsumer(
                 targetFolder: temporaryFolder,
                 fileManager: fileManager,
+                prettyPrint: shouldPrettyPrintOutputJSON, // Use the same configuration as the JSON output to make it convenient for the developer.
```
snprajwal, line 351: The boolean is no longer representative of what it is used for, should we consider renaming this to `shouldPrettyPrint`/`shouldPrettyPrintOutput`?

Relatedly, using `DOCC_JSON_PRETTYPRINT` to pretty print HTML sounds like an antipattern to me.

d-ronnqvist:
@swift-ci please test

Sources/DocCCommandLine/Action/Actions/Convert/FileWritingHTMLContentConsumer.swift
```diff
@@ -38,7 +38,7 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
             
             // Ensure that the index.html file has at least a `<head>` and a `<body>`.
             guard var beforeEndOfHead  = content.utf8.firstRange(of: "</head>".utf8)?.lowerBound,
-                  var afterStartOfBody = content.range(of: "<body[^>]*>", options: .regularExpression)?.upperBound
+                  var afterStartOfBody = content.range(of: "<body[^>]*>\n?", options: .regularExpression)?.upperBound
```
snprajwal: Minor: we can support files that were edited on Windows

```suggestion
                  var afterStartOfBody = content.range(of: "<body[^>]*>\r?\n?", options: .regularExpression)?.upperBound
```

Sources/DocCHTML/HTMLNode+fromXMLNode.swift
```diff
@@ -0,0 +1,58 @@
...
+            }
+            
+            let attributes = element.attributes?.compactMap {
+                HTMLNode.Attribute($0)
+            } ?? []
+            
+            if tag.isVoid {
+                self = ._voidElement(tag, attributes: attributes)
+            } else {
+                let contents = xmlNode.children?.compactMap {
```
snprajwal, line 35: Any child whose tag is not in the list of tags gets silently dropped here. It might be better to drop it loudly.

Sources/SwiftDocC/Model/Rendering/HTML/HTMLRenderer+FullPage.swift
```diff
@@ -106,14 +106,9 @@ package extension HTMLRenderer {
...
-        // Specify the "html" doctype for the document
-        let documentTypeDefinition = XMLDTD()
-        documentTypeDefinition.name = "html"
-        let page = XMLDocument(rootElement: .element(named: "html", children: [head, body], attributes: ["lang": "en-US"]))
-        page.documentContentKind = .html
-        page.dtd = documentTypeDefinition
+        let root = XMLNode.element(named: "html", children: [head, body], attributes: ["lang": "en-US"])
         
-        return page
+        return HTMLNode(from: root) ?? html(contents: [])
```
snprajwal, line 111: It might be better to have an assertion failure instead of silently returning an empty file.

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -265,8 +265,10 @@ package struct HTMLFormatter {
             //    ```
             
             let hasAttributeReasonToPresentOnSeparateLine = switch tag {
-                case .a, .td, .th: attributes.count > 1
-                default:           !attributes.isEmpty
+                case .a, .td, .th, .dt, .span:
```
snprajwal, line 269: `.dt` and `.span` can be added to the detailed comment above

Sources/DocCHTML/HTMLFormatter.swift
```diff
@@ -293,10 +295,13 @@ package struct HTMLFormatter {
                 // It's necessary to know what element comes next in the container (if any) to determine when it's allowed to omit the end tag.
                 childState.nextElementTag = nextIndex < contents.endIndex ? contents[nextIndex]._tag : nil
                 
-                // If the previous element presented on its own line and the current line is text, add a line break before the text as well.
-                if childState.presentOnCurrentLine || !child._isText {
+                let presentOnCurrentLine = shouldPresentInline(for: child)
+                if presentOnCurrentLine && !childState.presentOnCurrentLine {
+                    // If the previous element presented on its own line but this element presents on the same line, add a "trailing" line break after the previous element.
+                    appendLineBreakAndIndentation(depth: childState.depth)
                     childState.presentOnCurrentLine = shouldPresentInline(for: child)
```
snprajwal: This write can be removed since it is overwritten right below

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
```diff
@@ -322,61 +319,67 @@ final class FileWritingHTMLContentConsumerTests: XCTestCase {
...
+              <p>
+                Description of the <code>second</code> parameter.
+              </p>
+            </dd>
+          </dl>
+          <h2>Return Value</h2>
+          <p>Description of the return value.</p>
+          <h2>Discussion</h2>
+          <p>Further description of this method and how to use it.</p>
+
```
snprajwal, line 369: Is this newline expected? It contains the same number of spaces as the surrounding lines, looks like a bug where an empty element is being formatted

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
```diff
@@ -389,44 +392,53 @@ final class FileWritingHTMLContentConsumerTests: XCTestCase {
...
+                Deprecated
+              </p>
+              <p>
+                Description of why this <i>article</i> is deprecated.
+              </p>
+            </aside>
+          </section>
+          <h2>Custom discussion</h2>
+          <p>
+            It explains how a developer can perform some task using
```
snprajwal, line 423: There's a trailing newline here 👀 do we need to trim somewhere before adding a newline?

Sources/SwiftDocC/Model/Rendering/HTML/HTMLRenderer+FullPage.swift
d-ronnqvist: I don't think I'd want to trap in release builds for this and I also don't think that a debug-only assertion would be of any real help here.

Like I said in the PR description, this is a short-term stepping stone solution to try and break down the work into smaller pieces. The plan is to update all the code that creates XMLNode values to create HTMLNode values directly, which will also remove this code.

Sources/DocCHTML/HTMLNode+fromXMLNode.swift
d-ronnqvist: I consider this to be temporary code that will change once we add support for "foreign elements" (inline SVG or MathML) which I plan on adding in a later PR. 

The known lists of tags is all supported HTML tags (except for "foreign elements")—unless I missed a tag by mistake—so I don't think it failing to create a tag will be an issue in practice. Same as [below](https://github.com/swiftlang/swift-docc/pull/1583#discussion_r3665950169) I don't think I'd want to trap in release builds for this and I also don't think that a debug-only assertion would be of any real help here.

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
d-ronnqvist: Are you asking about the trailing _space_ before the newline? If so, that space is ignored by parsers so it doesn't matter if we keep it or trim it. It's there because the markdown `Text` element includes it in its textual contents.

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
d-ronnqvist: Kind of. It has to do with there being indentation in the template and the contents being added before that whitespace; immediately after the newline.

Similar to other comments; I consider the string manipulation to be a temporary solution until the page is represented as a hierarchy of HTMLNode values. In practice the template isn't pretty printed so real pages wouldn't have a whitespace "problem" like this—it's only occurring in these tests. Because the template isn't pretty printed in practice, this also means that if a developer specifies that the page content HTML should be pretty printed, that pretty printed content will be inserted into an otherwise compactly formatted file.  

Because I plan to address this later, and properly format the entire page, I didn't bother updating the regular expressions for the string replacement to catch this whitespace in the template.

Sources/DocCCommandLine/Action/Actions/Convert/FileWritingHTMLContentConsumer.swift
d-ronnqvist: Like I tried to explain in [this comment below](https://github.com/swiftlang/swift-docc/pull/1583/changes#r3666082080), I only made this change for the test output and I consider this string replacement code to be temporary. So, rather than discussing the likelihood of a custom template that's been edited on Windows I'll just remove this and have the "expected" test output look ever so slightly more unexpected instead.

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
d-ronnqvist: The full-page HTML output isn't even an officially "experimental" feature yet. It's merged pre-experimental so that other's can contribute to its early development. 

Because this feature is only intended to be used to continue developing it and trying to keep these changes _somewhat_ focused I felt that reusing the pretty print environmental variable was sufficiently in line with its purpose, which I consider to be to format the output to make it easier for a human to read. If DocC had supported other output formats when this environmental variable was added I don't know if we would have named it JSON-specific. 

Regardless, I feel that deciding how to control the pretty formatting of the HTML output is something that can be addressed separately at some point before this feature is officially pitched as an experimental feature.

d-ronnqvist:
@swift-ci please test

snprajwal (approved):
Thanks for the clarifications David, this looks good!

PR #1584, opened 2026-07-08
[4/n] Add result builder for constructing HTML element hierarchies

d-ronnqvist:
Bug/issue #, if applicable: rdar://165755530

This is a follow up to https://github.com/swiftlang/swift-docc/pull/1571 ~(still open)~, https://github.com/swiftlang/swift-docc/pull/1578 ~(still open)~, and https://github.com/swiftlang/swift-docc/pull/1583 ~(still open)~ that adds a result builder that makes the syntax nicer for creating `HTMLNode` element hierarchies.

As mentioned in https://github.com/swiftlang/swift-docc/pull/1583; this is not the intended final state but rather a _step_ towards replacing the `XMLNode` hierarchy with an `HTMLNode` hierarchy. Future PRs will update the code that creates `XMLNode` values to create `HTMLNode` values instead.

## Dependencies

This depends on and includes all the changes of:

- https://github.com/swiftlang/swift-docc/pull/1571
- https://github.com/swiftlang/swift-docc/pull/1578
- https://github.com/swiftlang/swift-docc/pull/1583

## Testing

Compared to https://github.com/swiftlang/swift-docc/pull/1583 this is a non-functional change.

## Checklist

Make sure you check off the following items. If they cannot be completed, provide a reason.

- [x] ~Added~ Updated tests
- [x] Ran the `./bin/test` script and it succeeded
- [x] Updated documentation if necessary

PR #1585, opened 2026-07-08
Draw separators between certain sections using CSS instead of `hr` tags

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This is a slight refinement to DocC's static HTML output to no longer emit `<hr>` tags between certain sections but to instead draw those visible separators/dividers using CSS.

Semantically, an `<hr>` tag represents a thematic break between paragraphs _within_ a section. 

This change also makes it easier to update the node creation code to use the result builders introduced in https://github.com/swiftlang/swift-docc/pull/1584 because it removes the code that checked what the previous section was to determine if it needed to insert an `<hr>` tag before the next section. This would have been very difficult to translate to a result builder. 

## Dependencies

None.

## Testing

1. Build some documentation and pass `--output-format experimental-html-for-development` to DocC.
2. Preview the output HTML pages. The thematic break looking separators should still be displayed in the same places as before (before the 2nd section for symbol pages and before any Topics section or See Also section (not currently present in the static HTML output))

## Checklist

- [ ] Added tests


snprajwal:
@swift-ci please test

snprajwal (approved):
Verified visually, LGTM

PR #1590, opened 2026-07-16
Deterministically register assets

snprajwal:
Bug/issue #, if applicable: rdar://182512073

## Summary

DocC keys assets by filename, so if two files with the same name in different directories collide, the winner is chosen by iterating an unordered set. Variants are also serialised by the winning source path instead of the rendered URL. This changes the process to sort assets prior to registration, and sort encoded variants by the rendered URL using traits as a tie-breaker.

This PR also modernises tests associated with this code path.

## Dependencies

N/A

## Testing

1. Create a catalog with two image assets sharing a filename in different subdirs, and referenced in an article.
2. Run `docc convert` several times in separate processes, each to its own output dir.
3. Previously, the output would differ across runs. With this patch, it should be identical.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Infrastructure/Bundle Assets/DataAssetManager.swift
+++ b/Sources/SwiftDocC/Infrastructure/Bundle Assets/DataAssetManager.swift
@@ -107,7 +107,7 @@ struct DataAssetManager {
     ///
     /// Data objects which have a file name ending with '~dark' are associated to their light variant.
     mutating func register(data datas: some Collection<URL>) throws {
-        for dataURL in datas {
+        for dataURL in datas.sorted(by: \.path) {
             let meta = try referenceMetaInformationForDataURL(dataURL)
 
             let referenceURL = URL(fileURLWithPath: meta.reference, isDirectory: false)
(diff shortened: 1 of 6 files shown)


Tests/SwiftDocCTests/Infrastructure/DataAssetManagerTests.swift
```diff
@@ -223,6 +226,26 @@ class DataAssetManagerTests: XCTestCase {
...
+
+        var forward = DataAssetManager()
+        try forward.register(data: urls)
+
+        var reversed = DataAssetManager()
+        try reversed.register(data: urls.reversed())
+
+        #expect(
+            forward.data(named: "overview.png", bestMatching: trait)?.url
+                == reversed.data(named: "overview.png", bestMatching: trait)?.url
```
franklinsch, line 252: Is it worth also asserting that we picked a specific one, so that if this logic ever changes, this test will fail?

Tests/SwiftDocCTests/Rendering/URLReferenceTests.swift
```diff
@@ -1,91 +1,134 @@
...
         encoder = RenderJSONEncoder.makeEncoder(assetPrefixComponent: "com.example.bundle2")
-        jsonData = try XCTUnwrap(try encoder.encode(decodedReference))
+        jsonData = try encoder.encode(decodedReference)
         decodedReference = try RenderJSONDecoder.makeDecoder().decode(VideoReference.self, from: jsonData)
-        XCTAssertEqual(Array(decodedReference.asset.metadata.keys), [URL(string: "/videos/com.example.bundle1/video.mov")!])
+        #expect(Array(decodedReference.asset.metadata.keys) == [URL(string: "/videos/com.example.bundle1/video.mov")!])
+    }
+
+    @Test
+    func imageReferenceVariantsAreSorted() throws {
```
franklinsch, line 88: Should we also have a test that covers the case where we exercise the path where we call `traits.joined()`?

Same for `videoReferenceVariantsAreSorted`

Sources/SwiftDocC/Infrastructure/Bundle Assets/DataAssetManager.swift
franklinsch: Is this a source of non-determinism that also needs addressing?

snprajwal: Nice catch, yes -- I don't want to address the FIXME in this PR and increase the scope, but we can make this minimally deterministic as part of this PR.

snprajwal: Never mind, it made more sense to properly fix this here.

Sources/SwiftDocC/Model/Rendering/References/ImageReference.swift
```diff
@@ -72,11 +72,17 @@ public struct ImageReference: MediaReference, URLReference, Equatable {
...
             let url = renderURL(for: value, prefixComponent: encoder.assetPrefixComponent)
             result.append(VariantProxy(url: url, traits: key, svgID: asset.metadata[value]?.svgID))
         }
+        // Sort by the rendered URL, and by traits for variants with identical URLs
+        result.sort { lhs, rhs in
+            if lhs.url.path != rhs.url.path {
+                return lhs.url.path < rhs.url.path
+            }
+            return lhs.traits.joined() < rhs.traits.joined()
+        }
```
franklinsch: nit: it would be good to refactor this with the one in `VideoReference.swift`.

franklinsch (approved):
Thanks!

snprajwal:
@swift-ci test

PR #1593, opened 2026-07-17
Deterministically render conditional conformances

snprajwal:
Bug/issue #, if applicable: rdar://182511900

## Summary

Conditional conformance constraints (e.g. "Available when Self is X") are not consistently present across multiple builds. This is due to two independent sources of non-determinism:

- When checking whether a `Self is <Parent>` constraint is trivially true and can be hidden, the renderer picks the symbol's parent from an unordered collection. However, symbols can sometimes be reached through multiple parent nodes in the topic graph, causing the constraint to be filtered out in some builds and retained in others.
- A reference can be present in a page both directly and as a dependency of another reference (e.g. via a symbol link in an abstract). Only the direct path attaches conformance constraints, so whichever write happened last decides whether the reference includes conformances.

To resolve these issues:

- When checking a `Self is <Parent>` constraint, the renderer now picks the parent with the shortest finite path in the topic graph, which is guaranteed to remain consistent for a given set of inputs.
- The code path that adds dependency references now also includes conformance constraints, and the information is present in the final list of references irrespective of the write order.

This PR also modernises tests for the documentation content renderer.

## Dependencies

N/A

## Testing

1. Build a catalog with both offending cases:
    - a member reachable from more than one parent that carries a conditional conformance (e.g. a protocol default implementation surfaced under both the protocol and a conforming type)
    - a symbol that is only reached through another reference's abstract (a symbol link), rather than being directly curated.
2. Run `docc convert` several times in separate processes, each to its own output dir.
3. Previously, the conformance list would change across runs. With these fixes, the output is identical across runs.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
+++ b/Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
@@ -194,7 +194,9 @@ public class DocumentationContentRenderer {
         }
         
         let isLeaf = SymbolReference.isLeaf(symbol)
-        let parentName = context.parents(of: reference).first
+        // A symbol can be reached in the topic graph through one or more parent nodes.
+        // To ensure stable ordering, use the parent with the shortest path to the symbol.
+        let parentName = context.shortestFinitePath(to: reference)?.last
             .flatMap { try? context.entity(with: $0).symbol?.names.title }
         
         let options = ConformanceSection.ConstraintRenderOptions(
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -468,6 +468,11 @@ public struct RenderNodeTranslator: SemanticVisitor {
             
             
             for dependencyReference in dependencies.topicReferences {
+                // If this reference is also a direct reference of the page,
+                // do not redundantly process it as a dependency reference.
+                guard renderReferences[dependencyReference.absoluteString] == nil else {
+                    continue
+                }
                 var dependencyRenderReference: TopicRenderReference
                 if let renderContext, let prerendered = renderContext.store.content(for: dependencyReference)?.renderReference as? TopicRenderReference {
                     dependencyRenderReference = prerendered
@@ -478,7 +483,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                 renderReferences[dependencyReference.absoluteString] = dependencyRenderReference
             }
             
-            // Add any conformance constraints to the reference, if any are present.
+            // Add conformance constraints to the reference, if any are present.
             if let conformanceSection = renderer.conformanceSectionFor(reference, collectedConstraints: collectedConstraints) {
                 renderReference.conformance = conformanceSection
             }
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -447,17 +447,9 @@ public struct RenderNodeTranslator: SemanticVisitor {
         
         for reference in collectedTopicReferences {
             var renderReference: TopicRenderReference
-            var dependencies: RenderReferenceDependencies
-            
-            if let renderContext, let prerendered = renderContext.store.content(for: reference)?.renderReference as? TopicRenderReference,
-                let renderReferenceDependencies = renderContext.store.content(for: reference)?.renderReferenceDependencies {
-                renderReference = prerendered
-                dependencies = renderReferenceDependencies
-            } else {
-                dependencies = RenderReferenceDependencies()
-                renderReference = renderer.renderReference(for: reference, dependencies: &dependencies)
-            }
-            
+            let dependencies: RenderReferenceDependencies
+            (renderReference, dependencies) = makeRenderReference(for: reference, with: renderer)
+
             for link in dependencies.linkReferences {
                 linkReferences[link.identifier.identifier] = link
             }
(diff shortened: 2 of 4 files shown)


Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
```diff
@@ -194,7 +194,9 @@ public class DocumentationContentRenderer {
         }
         
         let isLeaf = SymbolReference.isLeaf(symbol)
-        let parentName = context.parents(of: reference).first
+        // A symbol can be reached in the topic graph through one or more parent nodes.
+        // To ensure stable ordering, use the parent with the shortest path to the symbol.
+        let parentName = context.shortestFinitePath(to: reference)?.last
```
franklinsch, line 199: It would be good to run the performance testing script for this change. Specifically, I'd be curious if recomputing `reverseEdgesGraph` here is a problem.

Tests/SwiftDocCTests/Rendering/RenderNodeTranslatorTests.swift
```diff
@@ -1587,3 +1588,40 @@ class RenderNodeTranslatorTests: XCTestCase {
...
+            typeKind: .struct,
+            constraints: [.init(kind: .sameType, leftTypeName: "Self", rightTypeName: "Bar")]
+        )
+
+        let catalog = Folder(name: "unit-test.docc", content: [
+            JSONFile(name: "SomeModule.symbols.json", content: makeSymbolGraph(
+                moduleName: "SomeModule",
+                symbols: [
+                    makeSymbol(id: "s:Foo", kind: .struct, pathComponents: ["Foo"]),
+                    makeSymbol(id: "s:Bar", kind: .struct, pathComponents: ["Bar"]),
```
franklinsch, line 1606: Nit: technically, I don't think this symbol is needed for this test, but I can see why it's included in this test to be consistent with the `selfIsBar` mixin.

snprajwal: Yeah, we'd have a dangling constraint otherwise.

Sources/SwiftDocC/Model/Rendering/DocumentationContentRenderer.swift
snprajwal: No meaningful impact:

```
│ Metric                                   │ Change          │ 9417c16ac            │ current              │
│ Duration for 'convert-total-time'        │ no change¹,²    │ 10.33 secs           │ 9.509 secs           │
│ Duration for 'documentation-processing'  │ no change³,⁴    │ 3.732 secs           │ 3.076 secs           │
│ Duration for 'finalize-navigation-index' │ no change⁵,⁶    │ 0.072 secs           │ 0.038 secs           │
│ Peak memory footprint                    │ no change⁷,⁸    │ 687.4 MB             │ 691.3 MB             │
│ Data subdirectory size                   │ no change⁹      │ 200.9 MB             │ 200.9 MB             │
│ Index subdirectory size                  │ no change       │ 1.4 MB               │ 1.4 MB               │
│ Total DocC archive size                  │ no change¹⁰     │ 544.8 MB             │ 544.8 MB             │
│ Topic Anchor Checksum                    │ no change       │ 79a05f8af32dd880b420 │ 79a05f8af32dd880b420 │
│ Topic Graph Checksum                     │ no change       │ 02122cc8c951bdb8854c │ 02122cc8c951bdb8854c │
```
I wouldn't expect it to impact perf since it's called so infrequently (only when computing conformances for constrained symbols) that any perf deviation is negligible compared to the overall build.

snprajwal:
@swift-ci test

snprajwal:
@swift-ci test

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
anferbui, line 477: I think it would be good to deterministically added the conformance section only when we needed it, rather than always, so that we're not storing extra information we don't need, WDYT? I don't think there's any case for which we need the conformance section for dependencies, but I could be wrong.

We could check whether the reference is already in `renderReferences` before modifying it, with something like:

```diff
diff --git a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
index dbffde201..f4b800251 100644
--- a/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
@@ -468,6 +468,11 @@ public struct RenderNodeTranslator: SemanticVisitor {


             for dependencyReference in dependencies.topicReferences {
+                // Only add the reference if it doesn't already exist
+                if let _ = renderReferences[dependencyReference.absoluteString] {
+                    continue
+                }
+
                 var dependencyRenderReference: TopicRenderReference
                 if let renderContext, let prerendered = renderContext.store.content(for: dependencyReference)?.renderReference as? TopicRenderReference {
                     dependencyRenderReference = prerendered
@@ -475,10 +480,7 @@ public struct RenderNodeTranslator: SemanticVisitor {
                     var dependencies = RenderReferenceDependencies()
                     dependencyRenderReference = renderer.renderReference(for: dependencyReference, dependencies: &dependencies)
                 }
-                // Add conformance constraints to the dependency reference, if any are present.
-                if let conformanceSection = renderer.conformanceSectionFor(dependencyReference, collectedConstraints: collectedConstraints) {
-                    dependencyRenderReference.conformance = conformanceSection
-                }
+
                 renderReferences[dependencyReference.absoluteString] = dependencyRenderReference
             }
```

anferbui (commented):
Small comment about whether we want to always add conformances or not (adding them unnecessarily will make the Render Node JSON unnecessarily larger)

Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
anferbui: (It would be good to refactor this function a bit to not have very similar logic twice that we need to keep in sync)

snprajwal: Done and done, nice catch!

snprajwal:
@swift-ci test

anferbui (approved):
LGTM :D

PR #1594, opened 2026-07-17
Fix SGF generation in `bin/make-test-bundle`

snprajwal:
`bin/make-test-bundle` generates a symbol graph from the built Swift module as part of creating a test bundle. The include path for the `.swiftmodule` file was set to `.build/debug`. However, Swift 6.0 changed the directory structure to place built modules at `.build/debug/Modules`. This patch updates the path to include the extra component, and also replaces the hardcoded debug build output path with the output of `swift build --show-bin-path`.


bin/benchmark/.gitignore
```diff
@@ -1,9 +0,0 @@
-.DS_Store
-/.build
-/Packages
-/*.xcodeproj
-xcuserdata/
-DerivedData/
-.swiftpm/config/registries.json
-.swiftpm/xcode/package.xcworkspace/contents.xcworkspacedata
-.netrc
```
franklinsch, line 9: I don't know if `.netrc` is covered in the top-level `.gitignore`, but maybe it doesn't matter.

franklinsch (approved):

bin/benchmark/.gitignore
snprajwal: I looked into that, and it seems to be a stale entry. I can't think of a case where DocC needs a .netrc file for credentials currently. We can always add it back if we see the need for it.

snprajwal:
@swift-ci test

PR #1595, opened 2026-07-18
Revert "Revert "Temporarily avoid using FoundationXML to unblock CI""

owenv:
Reverts swiftlang/swift-docc#1092

I believe this may be causing macOS PR CI to fail


owenv:
@swift-ci test

snprajwal:
~https://github.com/swiftlang/swift-docc/pull/1090/changes/b5ead868eb5f0ee5cd9896382a0971cdf7a8b008 has the fix for the specific issue. The XML changes are unrelated, could you please only include the fix from the above commit instead?~

Compiler CI is using an older version of macOS, and may not have the XML fixes that are needed for it to work.

snprajwal (approved):

PR #1597, opened 2026-07-18
Fix non-deterministic symbol declaration selection

snprajwal:
Bug/issue #, if applicable: rdar://182555854&181916399

## Summary

The `min(by:)` comparator in `mainRenderFragments()` compared the LHS to the LHS, making it trivially false and causing the declaration fragments to be non-deterministically chosen. This patch updates this to use the platform ordering in `PlatformName` to pick the declaration of the highest priority platform. It also updates other code paths that choose declaration fragments to use this logic. Additionally, it updates the related overloads highlighting logic to deterministically use the declaration from `mainRenderFragments()` to compute the common fragments.

The only case where this comparator deviates from a strict weak ordering is in the case of macos vs macosX86, where both architectures are stored as the "macOS" platform and have the same priority in the platform hierarchy. However, this is now an unreachable state thanks to the fix in https://github.com/swiftlang/swift-docc-symbolkit/pull/112 which picks arm64 symbols in favour of x86 symbols when declarations from both architectures are present. The swift-docc-symbolkit dependency has been updated here to include the above fix.

## Dependencies

N/A

## Testing

1. Create symbol graphs with a symbol whose declaration varies across multiple platforms.
2. Run `docc convert` multiple times into different output dirs
3. Compare the symbol declaration on the rendered page. Prior to this patch, it might change across runs. With this patch, it remains consistent.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
+++ b/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
@@ -170,8 +170,8 @@ struct DeclarationsSectionTranslator: RenderSectionTranslator {
 
                     let conformance = renderNodeTranslator.contentRenderer.conformanceSectionFor(overloadReference, collectedConstraints: [:])
 
-                    let declarationFragments = overload.declarationVariants[trait]?.values
-                        .first?
+                    let declarationFragments = overload.declarationVariants[trait]?
+                        .mainRenderFragments()?
                         .declarationFragments
                     precondition(
                         declarationFragments != nil,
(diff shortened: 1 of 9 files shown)


Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
anferbui: Are these still needed or should they use `PlatformName.Comparable` or `PlatformName.isInOrder`?

anferbui (approved):
This looks good to me, just left a comment about some remaining logic we might want to update as well.

Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift
snprajwal: Nice catch! Replaced them :)

snprajwal:
@swift-ci test

PR #1598, opened 2026-07-21
Fix benchmark build with swift-argument-parser

snprajwal:
ParsableCommand in swift-argument-parser requires a Sendable conformance, which a `@MainActor` does not satisfy. The CLI entry point runs synchronously on the main thread, so this can be replaced with `MainActor.assumeIsolated` inside the run blocks.


snprajwal:
@swift-ci test

QuietMisdreavus (approved):

PR #1599, opened 2026-07-22
Test `bin/benchmark` and `bin/make-test-bundle` in CI

snprajwal:
We have auxiliary packages in this repository that, in some cases, depend on SwiftDocC. Changes to the main package may impact them, but they are currently not being built or tested in CI. Update `bin/test` to optionally build and test these packages if their source or dependents change. Also update the GitHub Actions workflow to include two optional jobs to build and test auxiliary packages. A prerequisite job checks for changes that impact these packages and sets boolean flags that are used to determine if the individual jobs need to run.

rdar://182804873

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -28,6 +28,7 @@ jobs:
       # - CMake: https://github.com/swiftlang/swift-docc/pull/1115
       enable_windows_checks: false
       enable_macos_checks: true
+
   soundness:
     name: Soundness
     uses: swiftlang/github-workflows/.github/workflows/soundness.yml@0.0.13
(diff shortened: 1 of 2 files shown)


.github/workflows/pull_request.yml
```diff
@@ -49,3 +50,57 @@ jobs:
...
+            --jq '.[].filename')"
+          echo "Changed files:"
+          echo "$changed"
+
+          make_test_bundle=false
+          benchmark=false
+          # `make-test-bundle` has no dependencies, only check its source
+          if echo "$changed" | grep -qE '^bin/make-test-bundle/'; then make_test_bundle=true; fi
+          # `benchmark` depends on SwiftDocC, check its source and the root package manifest
+          if echo "$changed" | grep -qE '^(bin/benchmark/|Package\.(swift|resolved)$)'; then benchmark=true; fi
```
QuietMisdreavus: Shouldn't this also check for changes to `Sources/SwiftDocC/` as well?

bin/test
```diff
@@ -22,6 +22,32 @@ DOCC_ROOT="$(dirname $(dirname $(filepath $0)))"
...
+
+# `make-test-bundle` has no dependencies and no test target, so we just build it.
+if should_build '^bin/make-test-bundle/'; then
+    printf "=> Building make-test-bundle…\n"
+    swift build --package-path "$DOCC_ROOT/bin/make-test-bundle"
+fi
+
+# `benchmark` depends on the root SwiftDocC package (via `../../`)
+# and has a test target, so rebuild and test it when it or the root package changes.
+if should_build '^(bin/benchmark/|Package\.(swift|resolved)$)'; then
```
QuietMisdreavus: Ditto here. If DocC itself changed, we should make sure that the benchmark utility continues to build as expected.

.github/workflows/pull_request.yml
snprajwal: The benchmark utility only uses the schema for `benchmark.json` from SwiftDocC. I don't see the need for us to build and test every time DocC changes, since that would cause it to trigger on almost every PR. However, we could gate it to check for changes to `Sources/SwiftDocC/Benchmark`, that's a nice improvement.

snprajwal:
@swift-ci test

QuietMisdreavus (approved):

snprajwal:
@swift-ci test

PR #1600, opened 2026-07-23
Allow cross language curation in See Also sections

snprajwal:
Bug/issue #, if applicable: rdar://182898058

## Summary

1588c4d2dd8b61f85bde4b9305ca48f87a8a9d0d removed support for curating internal links with no language overlap, and added a warning for cases that are encountered. This is due to the reference being unreachable in the navigator hierarchy. However, See Also sections do not contribute to the navigator hierarchy, and are not bound by this restriction. Update the filter to allow cross-language curation with no overlap in See Also sections.

## Dependencies

N/A

## Testing

1. Create a multi-language framework, with Swift-only and ObjC-only symbols.
2. Curate the ObjC-only symbol in the See Also section of a Swift-only symbol.
3. Previously, this link would be dropped. Now, it is retained.

## Checklist

- [x] Added tests


snprajwal:
@swift-ci test

sofiaromorales (approved):

PR #1601, opened 2026-07-23
Bump timeout for `ConvertServiceTests.processAndAssert`

snprajwal:
The convert service tests has a `processAndAssert` helper, which checks for expectations with a 1.0s timeout. This is too eager on Linux when running with the address sanitizer on CI, where the instrumented code is slowed down enough to cause tests using this helper to fail. Bump the timeout by 0.5s to provide some breathing room.


snprajwal:
@swift-ci test

snprajwal:
@swift-ci test

anferbui (approved):

PR #1602, opened 2026-07-25
feat(cli): Add support for repeating '--additional-symbol-graph-dir' argument to convert / emit-generated-curation commands

amorde:
Bug/issue #, if applicable: 

## Summary

Update `docc convert` and `docc process-catalog emit-generated-curation` to allow repeating the `--additional-symbol-graph-dir` argument to pass multiple directories to be searched for symbol graph files.

Today if you have multiple pre-built collections of symbol graph folders that you want to pass to a single command, you have to copy all of their contents to a single directory first. This can be particularly slow due the number of loose files involved if there are many folders.

Inspired by this bug in Bazel rules_apple https://github.com/bazelbuild/rules_apple/issues/3033 which had incorrectly assumed this is how the CLI worked

## Dependencies

N/A

## Testing

Unit tests are included in this PR.

Manual testing:

Steps:
1. Run `docc convert --additional-symbol-graph-dir {SomeDirectoryA} --additional-symbol-graph-dir {SomeDirectoryB}`
2. Run `docc process-catalog emit-generated-curation --additional-symbol-graph-dir {SomeDirectoryA} --additional-symbol-graph-dir {SomeDirectoryB}`

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
+++ b/Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2024-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2024-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -14,7 +14,7 @@ import SwiftDocC
 /// An action that emits documentation extension files that reflect the auto-generated curation.
 struct EmitGeneratedCurationAction: AsyncAction {
     let catalogURL: URL?
-    let additionalSymbolGraphDirectory: URL?
+    let additionalSymbolGraphDirectories: [URL]
     let outputURL: URL
     let depthLimit: Int?
     let startingPointSymbolLink: String?
@@ -23,7 +23,7 @@ struct EmitGeneratedCurationAction: AsyncAction {
     
     init(
         documentationCatalog: URL?,
-        additionalSymbolGraphDirectory: URL?,
+        additionalSymbolGraphDirectories: [URL],
         outputURL: URL?,
         depthLimit: Int?,
         startingPointSymbolLink: String?,
@@ -37,7 +37,7 @@ struct EmitGeneratedCurationAction: AsyncAction {
         }
         self.depthLimit = depthLimit
         self.startingPointSymbolLink = startingPointSymbolLink
-        self.additionalSymbolGraphDirectory = additionalSymbolGraphDirectory
+        self.additionalSymbolGraphDirectories = additionalSymbolGraphDirectories
         self.fileManager = fileManager
     }
     
@@ -47,7 +47,7 @@ struct EmitGeneratedCurationAction: AsyncAction {
             startingPoint: catalogURL,
             options: BundleDiscoveryOptions(
                 infoPlistFallbacks: [:],
-                additionalSymbolGraphFiles: symbolGraphFiles(in: additionalSymbolGraphDirectory)
+                additionalSymbolGraphFiles: symbolGraphFiles(in: additionalSymbolGraphDirectories)
             )
         )
         let context = try await DocumentationContext(bundle: bundle, dataProvider: dataProvider)
@@ -62,12 +62,4 @@ struct EmitGeneratedCurationAction: AsyncAction {
         
         return ActionResult(didEncounterError: false, outputs: [outputURL])
     }
-}
-
-private func symbolGraphFiles(in directory: URL?) -> [URL] {
-    guard let directory else { return [] }
-    
-    let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
-    return subpaths.map { directory.appendingPathComponent($0) }
-        .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
 }
--- a/Sources/DocCCommandLine/ArgumentParsing/ActionExtensions/ConvertAction+CommandInitialization.swift
+++ b/Sources/DocCCommandLine/ArgumentParsing/ActionExtensions/ConvertAction+CommandInitialization.swift
@@ -96,8 +96,8 @@ extension ConvertAction {
 
 package extension Docc.Convert {
     var bundleDiscoveryOptions: BundleDiscoveryOptions {
-        let additionalSymbolGraphFiles = symbolGraphFiles(in: inputsAndOutputs.additionalSymbolGraphDirectory)
-        
+        let additionalSymbolGraphFiles = symbolGraphFiles(in: inputsAndOutputs.additionalSymbolGraphDirectories)
+
         return BundleDiscoveryOptions(
             fallbackDisplayName: infoPlistFallbacks.fallbackBundleDisplayName,
             fallbackIdentifier: infoPlistFallbacks.fallbackBundleIdentifier,
@@ -106,12 +106,4 @@ package extension Docc.Convert {
             additionalSymbolGraphFiles: additionalSymbolGraphFiles
         )
     }
-}
-
-private func symbolGraphFiles(in directory: URL?) -> [URL] {
-    guard let directory else { return [] }
-    
-    let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
-    return subpaths.map { directory.appendingPathComponent($0) }
-        .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
 }
--- a/Sources/DocCCommandLine/ArgumentParsing/Subcommands/EmitGeneratedCuration.swift
+++ b/Sources/DocCCommandLine/ArgumentParsing/Subcommands/EmitGeneratedCuration.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2023-2024 Apple Inc. and the Swift project authors
+ Copyright (c) 2023-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -50,13 +50,13 @@ extension Docc.ProcessCatalog {
             @Option(
                 name: [.customLong("additional-symbol-graph-dir")],
                 help: ArgumentHelp(
-                    "Path to a directory of additional symbol graph files.",
+                    "Path to a directory of additional symbol graph files. May be specified multiple times.",
                     valueName: "symbol-graph-dir"
                 ),
                 transform: URL.init(fileURLWithPath:)
             )
-            var additionalSymbolGraphDirectory: URL?
-            
+            var additionalSymbolGraphDirectories: [URL] = []
+
             /// A user-provided location where the command will write the updated catalog output.
             @Option(
                 name: [.customLong("output-path")],
@@ -135,7 +135,7 @@ extension Docc.ProcessCatalog {
         func run() async throws {
             let action = try EmitGeneratedCurationAction(
                 documentationCatalog: inputsAndOutputs.documentationCatalog,
-                additionalSymbolGraphDirectory: inputsAndOutputs.additionalSymbolGraphDirectory,
+                additionalSymbolGraphDirectories: inputsAndOutputs.additionalSymbolGraphDirectories,
                 outputURL: inputsAndOutputs.outputURL,
                 depthLimit: generationOptions.depthLimit,
                 startingPointSymbolLink: generationOptions.startingPointSymbolLink
--- a/Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
+++ b/Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
@@ -1,0 +1,25 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2026 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import SwiftDocC
+
+func symbolGraphFiles(in directories: [URL]) -> [URL] {
+    directories
+        // avoid traversing the same exact folders multiple times
+        .uniqueElements(by: \.standardizedFileURL)
+        .flatMap { directory in
+            let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
+            return subpaths.map { directory.appendingPathComponent($0) }
+                .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
+        }
+        // if overlapping paths were provided (e.g. 'foo/' and 'foo/folder') omit duplicates
+        .uniqueElements(by: \.standardizedFileURL)
+}
--- a/Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
+++ b/Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
@@ -11,6 +11,13 @@
 import Foundation
 import SwiftDocC
 
+/// Returns the symbol graph files found by recursively searching the given directories.
+///
+/// Files are deduplicated and returned in the order that their directories were provided in.
+/// Directories that don't exist are ignored.
+///
+/// - Parameter directories: The directories to search for symbol graph files.
+/// - Returns: The symbol graph files in the given directories.
 func symbolGraphFiles(in directories: [URL]) -> [URL] {
     directories
         // avoid traversing the same exact folders multiple times
--- a/Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
+++ b/Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
@@ -18,7 +18,7 @@ import ArgumentParser
 class ConvertSubcommandTests: XCTestCase {
     private let testBundleURL = Bundle.module.url(
         forResource: "LegacyBundle_DoNotUseInNewTests", withExtension: "docc", subdirectory: "Test Bundles")!
-    
+
     private let testTemplateURL = Bundle.module.url(
         forResource: "Test Template", withExtension: nil, subdirectory: "Test Resources")!
 
@@ -244,7 +244,7 @@ class ConvertSubcommandTests: XCTestCase {
                 testBundleURL.path,
             ])
             
-            XCTAssertEqual(convertOptions.inputsAndOutputs.additionalSymbolGraphDirectory, nil)
+            XCTAssertEqual(convertOptions.inputsAndOutputs.additionalSymbolGraphDirectories, [])
         }
         
         // Is set when passed
@@ -256,8 +256,8 @@ class ConvertSubcommandTests: XCTestCase {
             ])
             
             XCTAssertEqual(
-                convertOptions.inputsAndOutputs.additionalSymbolGraphDirectory,
-                URL(fileURLWithPath: "/path/to/folder-of-symbol-graph-files")
+                convertOptions.inputsAndOutputs.additionalSymbolGraphDirectories,
+                [URL(fileURLWithPath: "/path/to/folder-of-symbol-graph-files")]
             )
         }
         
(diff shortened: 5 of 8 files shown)


heckj:
@amorde so the idea here is that you can specify the additional symbol graph directory option multiple times with different directories? 

How does that help or apply to the set of commands `process-archive`? I wasn't aware of a use case where you'd use `process archive` to amend or extend an existing archive with more symbols - I thought that was the space of using `docc merge` to combine archives together.

snprajwal:
> How does that help or apply to the set of commands `process-archive`? I wasn't aware of a use case where you'd use `process archive` to amend or extend an existing archive with more symbols - I thought that was the space of using `docc merge` to combine archives together.

The author is talking about `process-catalog`, whose inputs are the same as `docc convert`:

```
$ xcrun docc process-catalog emit-generated-curation --help
OVERVIEW: Write documentation extension files with markdown representations of DocC's automatic curation.

Pass the same '<catalog-path>' and '--additional-symbol-graph-dir <symbol-graph-dir>' as you would for `docc convert` to emit documentation extension files for your project.
```

Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
```diff
@@ -64,10 +64,12 @@ struct EmitGeneratedCurationAction: AsyncAction {
...
 }
 
-private func symbolGraphFiles(in directory: URL?) -> [URL] {
-    guard let directory else { return [] }
-    
-    let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
-    return subpaths.map { directory.appendingPathComponent($0) }
-        .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
+private func symbolGraphFiles(in directories: [URL]) -> [URL] {
+    guard !directories.isEmpty else { return [] }
```
snprajwal: `flatMap` below returns `[]` for an empty input list, so this guard can be removed.

Sources/DocCCommandLine/ArgumentParsing/ActionExtensions/ConvertAction+CommandInitialization.swift
```diff
@@ -108,10 +108,12 @@ package extension Docc.Convert {
...
 }
 
-private func symbolGraphFiles(in directory: URL?) -> [URL] {
-    guard let directory else { return [] }
-    
-    let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
-    return subpaths.map { directory.appendingPathComponent($0) }
-        .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
+private func symbolGraphFiles(in directories: [URL]) -> [URL] {
+    guard !directories.isEmpty else { return [] }
```
snprajwal: `flatMap` below returns `[]` for an empty input list, so this guard can be removed.

Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
```diff
@@ -64,10 +64,12 @@ struct EmitGeneratedCurationAction: AsyncAction {
     }
 }
 
-private func symbolGraphFiles(in directory: URL?) -> [URL] {
-    guard let directory else { return [] }
-    
-    let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
-    return subpaths.map { directory.appendingPathComponent($0) }
-        .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
+private func symbolGraphFiles(in directories: [URL]) -> [URL] {
```
snprajwal: This method is duplicated across two files, we can move it into `Utility/SymbolGraphFileDiscovery.swift` and re-use it in both places instead

Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
```diff
@@ -64,10 +64,12 @@ struct EmitGeneratedCurationAction: AsyncAction {
...
+    guard !directories.isEmpty else { return [] }
+
+    return directories.flatMap { directory in
+        let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
+        return subpaths.map { directory.appendingPathComponent($0) }
+            .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
```
snprajwal: How would this interact with duplicated paths? E.g. if I pass `foo/bar` and `foo` in two separate flags, the symbol graphs would be discovered and loaded twice.

DocC has a helper to dedupe here:

https://github.com/swiftlang/swift-docc/blob/0a23fbb8bb868a2706f07ab2b82049654a5c05a2/Sources/DocCCommandLine/Utility/FoundationExtensions/Sequence%2BUnique.swift#L12-L16


```suggestion
            .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
            .uniqueElements(by: \.standardizedFileURL)
```

snprajwal (commented):
This is a nice enhancement, thank you for working on it :)

Some non-code bits would need to be updated:

- The help text and usage strings should refer to the plural form of the flag.
- `NotEnoughInformationError` in `DocumentationInputsProvider.swift` should use the plural form in the error message.

snprajwal:
Also, could you please update the copyright headers at the top of the edited files? You'd have to bump the year to 2026.

Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
amorde: Sure - the duplication was existing so I had opted to leave it rather than adding that refactor, but happy to include it

Sources/DocCCommandLine/Action/Actions/EmitGeneratedCurationAction.swift
amorde: Good call. I think it would make sense to apply this to `directories` as well so we don't traverse the same directories multiple times if they were duplicated.

amorde: Fixed and added tests for both duplication scenarios

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
```diff
@@ -0,0 +1,25 @@
...
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import SwiftDocC
+
+func symbolGraphFiles(in directories: [URL]) -> [URL] {
```
snprajwal, line 21: It would be good to add a doc comment explaining what this utility method does:


```suggestion
/// Returns the symbol graph files found by recursively searching the given directories.
///
/// Files are deduplicated and returned in the order that their directories were provided in.
/// Directories that don't exist are ignored.
///
/// - Parameter directories: The directories to search for symbol graph files.
/// - Returns: The symbol graph files in the given directories.
func symbolGraphFiles(in directories: [URL]) -> [URL] {
```

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
```diff
@@ -0,0 +1,25 @@
...
+    directories
+        // avoid traversing the same exact folders multiple times
+        .uniqueElements(by: \.standardizedFileURL)
+        .flatMap { directory in
+        let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
+        return subpaths.map { directory.appendingPathComponent($0) }
+            .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
+            // if overlapping paths were provided (e.g. 'foo/' and 'foo/folder') omit duplicates
+            .uniqueElements(by: \.standardizedFileURL)
+    }
```
snprajwal: This doesn't catch duplicates since the dedupe only happens on a single directory scan, where it is impossible to have duplicates. We need to dedupe _after_ all symbol graphs have been gathered.


```suggestion
    directories
        // avoid traversing the same exact folders multiple times
        .uniqueElements(by: \.standardizedFileURL)
        .flatMap { directory in
            let subpaths = FileManager.default.subpaths(atPath: directory.path) ?? []
            return subpaths.map { directory.appendingPathComponent($0) }
                .filter { DocumentationBundleFileTypes.isSymbolGraphFile($0) }
        }
            // if overlapping paths were provided (e.g. 'foo/' and 'foo/folder') omit duplicates
            .uniqueElements(by: \.standardizedFileURL)
```

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
```diff
@@ -276,6 +279,73 @@ class ConvertSubcommandTests: XCTestCase {
...
+        // Is unique when the directories overlap
+        do {
+            let convertOptions = try Docc.Convert.parse([
+                testBundleURL.path,
+                "--additional-symbol-graph-dir",
+                testBundleURL.path,
+                "--additional-symbol-graph-dir",
+                mixedLanguageFrameworkTestBundleURL.path,
+                "--additional-symbol-graph-dir",
+                mixedLanguageFrameworkTestBundleURL.appendingPathComponent("swift").path
```
snprajwal: This path doesn't exist, it would need to be `symbol-graphs/swift` for the test to actually verify deduping.

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
```diff
@@ -276,6 +279,73 @@ class ConvertSubcommandTests: XCTestCase {
...
+
+            XCTAssertEqual(convertOptions.bundleDiscoveryOptions.additionalSymbolGraphFiles.map { $0.lastPathComponent }.sorted(), [
+                "FillIntroduced.symbols.json",
+                // One in `clang`, one in `swift`
+                "MixedLanguageFramework.symbols.json",
+                "MixedLanguageFramework.symbols.json",
+                "MyKit@SideKit.symbols.json",
+                "mykit-iOS.symbols.json",
+                "sidekit.symbols.json",
+            ])
```
snprajwal, line 325: The two identically named symbol graphs can make this confusing to the reader. Since we check in the assertion above this one for the right symbol graphs being picked up, we can instead assert the count here to ensure the dedupe works as expected:


```suggestion
            XCTAssertEqual(convertOptions.bundleDiscoveryOptions.additionalSymbolGraphFiles.count, 6)
```

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
```diff
@@ -276,6 +279,73 @@ class ConvertSubcommandTests: XCTestCase {
...
```
snprajwal, line 348: Same as the earlier comment:


```suggestion
            XCTAssertEqual(convertOptions.bundleDiscoveryOptions.additionalSymbolGraphFiles.count, 6)
```

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
```diff
@@ -276,6 +279,73 @@ class ConvertSubcommandTests: XCTestCase {
                 "sidekit.symbols.json",
             ])
         }
+
```
snprajwal, line 279: The new assertions are extending an XCTest case that uses the legacy bundle. I looked into whether we can use Swift Testing to write it better, but there's some stuff that needs to be added before we can do that. This is okay for this PR, but could you please include a fixme here so this can be addressed in the future?

```suggestion

  // FIXME: Extract this into a dedicated test for `symbolGraphFiles(in:)` which uses `TestFileSystem`.
  // This is blocked on `FileManagerProtocol` supporting `.subpaths(atPath:)`.
  
```

Sources/DocCCommandLine/ArgumentParsing/Subcommands/EmitGeneratedCuration.swift
```diff
@@ -50,13 +50,13 @@ extension Docc.ProcessCatalog {
             @Option(
                 name: [.customLong("additional-symbol-graph-dir")],
                 help: ArgumentHelp(
-                    "Path to a directory of additional symbol graph files.",
+                    "Path to a directory of additional symbol graph files. May be specified multiple times.",
```
d-ronnqvist, line 53: Is there a common conversation among CLI tools for the help text of list arguments like this? 

We don't currently spell out that other options like `--platform`, `--dependency`, or `--Wwarning`/`--Werror` can be specified multiple times.

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
d-ronnqvist, line 21: If we're extracting this code and refining it, can we also add a `FileManagerProtocol` parameter so that tests can verify this behavior with an in-memory file system?

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
d-ronnqvist: We also have helpers like `FileManagerProtocol.recursiveFiles(startingPoint:options:)` that we could use here. That way we could loop over each directory with a basic for loop and accumulate the found files into a single set.

We could also change the return type to be a set rather than an array to better reflect that the order of the found files isn't guaranteed to be stable across executions.

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
```diff
@@ -18,7 +18,10 @@ import ArgumentParser
 class ConvertSubcommandTests: XCTestCase {
     private let testBundleURL = Bundle.module.url(
         forResource: "LegacyBundle_DoNotUseInNewTests", withExtension: "docc", subdirectory: "Test Bundles")!
-    
+
+    private let mixedLanguageFrameworkTestBundleURL = Bundle.module.url(
```
d-ronnqvist: minor: If this is only used inside one test it can be defined as a local variable inside that test instead.

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
d-ronnqvist: I think it would be better to implement `symbolGraphFiles(in:)` using API that's available on `FileManagerProtocol`, for example `recursiveFiles(startingPoint:options:)` instead.

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
```diff
@@ -0,0 +1,25 @@
...
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Foundation
+import SwiftDocC
+
+func symbolGraphFiles(in directories: [URL]) -> [URL] {
+    directories
+        // avoid traversing the same exact folders multiple times
+        .uniqueElements(by: \.standardizedFileURL)
```
d-ronnqvist, line 24: Question: would it be worth warning about this or filtering these values during the validation of this command line argument?

d-ronnqvist (commented):
I think this is a fair enhancement but I wanted to note that because DocC searches the specified `--additional-symbol-graph-dir` value recursively, tools today can specify a single value that's a common subpath if they want to pass different symbol graph files that exist in different subdirectories. 

I'll also remark that a single `docc convert` call only supports symbol graph files for a single module. DocC will warn if developers try to pass symbol graphs for different module and the general behavior is largely undefined (with several features known to now work correctly).

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
amorde: ah yep. will fix

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
amorde: If we do this then if it were to fail, you'd only see the count is different and you'd have to revert it back or manual debug to see _what_ was actually different

Tests/DocCCommandLineTests/ArgumentParsing/ConvertSubcommandTests.swift
amorde: In general I'm matching the existing code here. I'd hoped I wouldnt have to create a new bundle to add another simple test to these existing tests. 

Shouldn't that FIXME technically be applied to everything in this file using that bundle? why just the test I added?

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
amorde: Could we keep this to a separate change? I was intentionally avoiding adding refactors to this PR to limit the scope of the actual change here

amorde:
> I think this is a fair enhancement but I wanted to note that because DocC searches the specified --additional-symbol-graph-dir value recursively, tools today can specify a single value that's a common subpath if they want to pass different symbol graph files that exist in different subdirectories.

Yes, that works in cases where they are bundled together in a single folder. but if they are in places where you don't want to pay the cost of enumerating large directory hierarchies, you're forced to copy them all to a single folder first, which is the work I'm looking to avoid by this change.

> I'll also remark that a single docc convert call only supports symbol graph files for a single module. DocC will warn if developers try to pass symbol graphs for different module and the general behavior is largely undefined (with several features known to now work correctly).

This does not match my current experience. I've tested this with symbols from many modules and it works, producing an archive that seems to match the new `docc merge` command

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
amorde: I'm not sure. What other behavior would the invoker of the CLI expect other than for these to be deduped in this case? is that warning actionable?

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
amorde: `FileManagerProtocol` does not currently have a `subpaths(atPath:)` API so I would either need to add it or refactor it to use `contentsOfDirectory`

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
amorde: Fixed. Interestingly, even after fixing the incorrect path used in the test, the test passed even without this change

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
d-ronnqvist: I wouldn't say that it's expected for a caller to pass the same symbol graph directory path more than once so I don't know if any behavior is more expected than any other behavior. Regardless, it's almost certainly a mistake for the caller to repeat the same path.

There is a clear action to that warning: remove the duplicate flag and only pass the path once.

(I'll also say that it would be reasonable for this function to expect that the inputs are unique and only deduplicate the outputs)

Sources/DocCCommandLine/Utility/SymbolGraphFileDiscovery.swift
d-ronnqvist: This implementation is already changed from what it was before and the API is made more visible than it was before (no longer file-private). Updating this function to use `FileManagerProtocol` would be a good leave-it-better-than-you-found-it change to the API.

d-ronnqvist: We shouldn't add a `subpaths(atPath:)` API to that protocol. 

Like I mentioned in https://github.com/swiftlang/swift-docc/pull/1602#discussion_r3666989814, we have other helpers for iterating over the files in a directory structure. For example, this implementation could probably be written as a basic for loop using `FileManagerProtocol.recursiveFiles(startingPoint:options:)`

Sources/DocCCommandLine/ArgumentParsing/Subcommands/EmitGeneratedCuration.swift
d-ronnqvist: Whichever approach we go with, it would be good to apply it consistently for all these command line options.

d-ronnqvist:
> > I'll also remark that a single docc convert call only supports symbol graph files for a single module. DocC will warn if developers try to pass symbol graphs for different module and the general behavior is largely undefined (with several features known to now work correctly).
> 
> This does not match my current experience. I've tested this with symbols from many modules and it works, producing an archive that seems to match the new `docc merge` command

You may be lucky and be only minimally impacted by the issues but they are there, otherwise DocC wouldn't [warn about them](https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2658). However, if you're less lucky there are some otherwise reasonable inputs that when combined with this unsupported and problematic setup can have severe impacts like [dropped pages](https://github.com/swiftlang/swift-docc/blob/main/Sources/SwiftDocC/Semantics/ReferenceResolver.swift#L557-L576). 

Even in the lucky case where all of the _pages_ are fine, it remains true that if you pass DocC symbol graphs for different modules then it'll result in a disjointed hierarchy with multiple root pages. One way that this presents itself differently on the rendered page—compared in than an archive created with `docc merge`—is that the navigator sidebar doesn't display the unified hierarchy of _all_ your modules, it only displays one module at a time.

PR #1604, opened 2026-07-28
Preserve contentful HTML when transforming archives

### LINKED ISSUE #1588: docc process-archive strips experimental SEO/HTML content when invoked on an archive
### Description

If you generate an archive using the experimental option`--experimental-transform-for-static-hosting-with-content`
and then use `docc process-archive` on it, the HTML content is regenerated for every symbol within the catalog, and the additional HTML information is stripped, removing the ability to more easily index the content from a crawler for Search SEO or more broadly programmatic access.

When I'm assembling a combined DocC catalog, after I've merged multiple catalogs, I'd like to be able make a final tweak to the hosting base path (for which I currently use `docc process-archive`), and either have it preserve that additional HTML detail from each symbol, or have an option that I can enable that will also generate the additional HTML content into the results.

Related to #1532 

### Checklist

### Expected Behavior

If original DocC contents included the HTML additional content in the stubs, I'd expect that process-archive would preserve that by default

### Actual behavior

It regenerates the HTML for every symbol, stripping any options for additional content there

### Steps To Reproduce

_No response_

6.3.3

```shell
swift-driver version: 1.168.4 Apple Swift version 6.4 (swiftlang-6.4.0.25.4 clang-2100.3.25.1)
Target: arm64-apple-macosx27.0.0
```

Dave861:
## Summary

- Preserve page-specific titles, descriptions, and semantic fallback content when processing a contentful archive.
- Continue applying the new hosting base path and renderer template.
- Add focused virtual-file-system regression coverage.

Fixes #1588

## Testing

- Focused contentful static-hosting regression test
- StaticHostableTransformerTests
- TransformForStaticHostingActionTests
- StaticHostingWithContentTests
- Manual process-archive reproduction
- Source and license checks

The full local test run passed except for two existing tests whose shell commands do not support checkout paths containing spaces.

### DIFF
--- a/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
+++ b/Sources/DocCCommandLine/Transformers/StaticHostableTransformer.swift
@@ -74,8 +74,133 @@ struct StaticHostableTransformer {
                 try fileManager.createDirectory(at: outputDirectoryURL, withIntermediateDirectories: true, attributes: [:])
             }
             
-            try fileManager.createFile(at: outputDirectoryURL.appendingPathComponent("index.html"), contents: indexHTMLData)
+            let outputIndexHTMLURL = outputDirectoryURL.appendingPathComponent("index.html")
+            let outputIndexHTMLData = indexHTMLDataPreservingPageContent(
+                at: outputIndexHTMLURL
+            )
+
+            try fileManager.createFile(
+                at: outputIndexHTMLURL,
+                contents: outputIndexHTMLData
+            )
         }
+    }
+
+    private func indexHTMLDataPreservingPageContent(at outputURL: URL) -> Data {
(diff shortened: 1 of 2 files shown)


d-ronnqvist (changes requested):
Thank you for this contribution. 
Unfortunately, I'm not convinced that this is a solution that we want to pursue. 

I think it would be better to have more discussion about the problem and solution in [the issue](#1588) before reviewing the code.

Dave861:
Thanks for the direction. I’ll pause this implementation and continue the design discussion in #1588.

Dave861:
Thanks for clarifying in #1588. Since the supported workflow is to configure the base path and custom headers/footers on each archive before merging, the preservation approach in this PR isn’t the right direction. I’ll close this out.

PR #1605, opened 2026-07-28
Only crawl each page once during curation

snprajwal:
Bug/issue #, if applicable: rdar://183418909

## Summary

The documentation curator starts crawling at every top-level page. Since a top-level page can be curated under another top-level page, a page can be both a starting point for the crawl and a descendant of another starting point.

We currently only check at the recursive call site if a page has been crawled. Due to this, a top-level page can be processed twice in a non-recursive manner (once as a descendant, and once as a top-level starting point), leading to identical diagnostics being emitted twice. This occurs deterministically with articles as the input is ordered, but non-deterministically with symbols as the input is a set, whose ordering is not guaranteed.

Move the guard to an earlier point where it ensures that each page is only crawled once, either as a top-level page or a descendant.

## Dependencies

N/A

## Testing

1. Create a catalog with two articles, where one curates the other.
2. Add an invalid link to the child article.
3. `xcrun docc convert` on the catalog.
4. Previously, the "invalid link" diagnostic would be emitted twice. Now, it is emitted only once.

## Checklist

- [x] Added tests

### DIFF
--- a/Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
+++ b/Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
@@ -798,4 +798,71 @@ struct DocumentationCuratorTests_new {
         #expect(swiftOnlyRenderNode.seeAlsoSections.flatMap(\.identifiers).contains { $0.hasSuffix("/ObjcOnlyClass") },
                 "Cross-language reference should be present in the parent page's See Also section")
     }
+
+    // This verifies determinism in previously non-deterministic behavior that cannot be reproduced reliably in a test.
+    // If you suspect that your changes might affect this test,
+    // you need to run it repeatedly (and relaunch for each repetition) to verify that the behavior remains deterministic.
+    @Test
+    func validatesEachSymbolTaskGroupLinksOnlyOnce() async throws {
+        let context = try await load(catalog:
+            Folder(name: "unit-test.docc", content: [
+                JSONFile(name: "ModuleName-swift.symbols.json", content: makeSymbolGraph(moduleName: "ModuleName", symbols: [
+                    makeSymbol(id: "some-container", language: .swift, kind: .class, pathComponents: ["SomeContainer"]),
+                    makeSymbol(id: "swift-only-class", language: .swift, kind: .class, pathComponents: ["SwiftOnlyClass"]),
+                ])),
+                JSONFile(name: "ModuleName-objc.symbols.json", content: makeSymbolGraph(moduleName: "ModuleName", symbols: [
+                    makeSymbol(id: "objc-only-class", language: .objectiveC, kind: .class, pathComponents: ["ObjcOnlyClass"]),
+                ])),
+                TextFile(name: "SomeContainer.md", utf8Content: """
+                # ``ModuleName/SomeContainer``
+
+                ## Topics
+
+                ### Curating another top-level symbol
+
+                - ``SwiftOnlyClass``
+                """),
+                TextFile(name: "SwiftOnlyClass.md", utf8Content: """
+                # ``ModuleName/SwiftOnlyClass``
+
+                ## Topics
+
+                ### Cross-language curation
+
+                - ``ObjcOnlyClass``
+                """),
+            ])
+        )
+
+        #expect(context.diagnostics.map(\.identifier) == ["UnreachableCrossLanguageCuration"],
+                "Encountered unexpected problems: \(context.diagnostics.map(\.summary))")
+    }
+
+    @Test
+    func validatesEachArticleTaskGroupLinksOnlyOnce() async throws {
+        let context = try await load(catalog:
+            Folder(name: "unit-test.docc", content: [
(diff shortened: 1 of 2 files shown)


Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -798,4 +798,90 @@ struct DocumentationCuratorTests_new {
         #expect(swiftOnlyRenderNode.seeAlsoSections.flatMap(\.identifiers).contains { $0.hasSuffix("/ObjcOnlyClass") },
                 "Cross-language reference should be present in the parent page's See Also section")
     }
+
+    @Test(arguments: [false, true])
+    func validatesEachSymbolTaskGroupLinksOnlyOnce(crawlContainerFirst: Bool) async throws {
```
d-ronnqvist: Question: was this an issue before or is it adding a test to cover a behavior that was already working correctly?

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -798,4 +798,90 @@ struct DocumentationCuratorTests_new {
...
+                ## Topics
+
+                - <doc:Second>
+                """),
+                TextFile(name: "Second.md", utf8Content: """
+                # Second article
+
+                ## Topics
+
+                - <https://example.com/something>
```
d-ronnqvist, line 857: Question: What behavior is this content verifying in this test?

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -798,4 +798,90 @@ struct DocumentationCuratorTests_new {
...
+            Folder(name: "unit-test.docc", content: [
+                TextFile(name: "Root.md", utf8Content: """
+                # Root
+
+                @Metadata {
+                  @TechnologyRoot
+                }
+
+                Both articles are automatically curated under this page.
+                """),
```
d-ronnqvist: Question: Would the test work the same without this explicit root markup? IIRC DocC will synthesize more or less the same root page for an article-only catalog with more than one articles.

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
```diff
@@ -798,4 +798,90 @@ struct DocumentationCuratorTests_new {
...
+        let moduleReference    = try #require(context.soleRootModuleReference)
+        let containerReference = try #require(context.knownPages.first(where: { $0.lastPathComponent == "SomeContainer"  }))
+        let swiftOnlyReference = try #require(context.knownPages.first(where: { $0.lastPathComponent == "SwiftOnlyClass" }))
+
+        // Top-level symbols are crawled in a non-deterministic order,
+        // so ensure the symbols are processed only once in both orders (container-first and child-first).
+        var crawler = DocumentationCurator(in: context)
+        for reference in [moduleReference] + (crawlContainerFirst ? [containerReference, swiftOnlyReference] : [swiftOnlyReference, containerReference]) {
+            try crawler.crawlChildren(of: reference, relateNodes: { _, _ in })
+        }
```
d-ronnqvist: I find it kind of an anti-pattern to repeat an operation in a test that the context performs _after_ the context has already performed that operation. Is there an observable behavior that we can assert from when the context performs the curation crawling?

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
snprajwal: This was an existing issue that was fixed with the change in this patch, I added a separate test for symbols and articles since the input types are different in these two cases.

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
snprajwal: Previously, the diagnostic would be emitted twice if the page was processed as a descendant first, and as a top-level page again. This verifies that the diagnostic is emitted only once. The diagnostic itself is irrelevant, we could have used any diagnostic that's emitted in DocumentationCurator. I just found this one to be the easiest to simulate.

Tests/SwiftDocCTests/Infrastructure/DocumentationCuratorTests.swift
snprajwal: Yeah it's not pretty, but I was unable to get a good hit rate on the test without doing this container-first/child-first dance. We don't need this with the test for articles since they are always processed in a deterministic order.

That said, we have other tests in DocC where we warn that the behaviour was previously non-deterministic, and requires multiple runs to verify. I'll add that here and reduce the test to a single pass. It should still hit the offending code path 1/2 runs if it ever regresses, so one of the many CI runs we do will catch it.

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

PR #1606, opened 2026-07-28
Serialize DocCHTMLTests on Linux to avoid libxml2 crash

snprajwal:
On Linux, FoundationXML is backed by libxml2, which isn't thread-safe by default. The library uses global state that backs the XMLNode type, and it's the caller's responsibility to synchronize access to it [1]. It also documents that "xmlCleanupParser is not thread-safe", and memory corruption is likely if other threads make libxml2 calls afterwards [2].

Swift Testing runs in parallel by default, with multiple XMLNode trees being built and torn down at the same time. On Linux, this causes the tests to intermittently crash with SIGSEGV (signal 11) inside `XMLNode.deinit`.

Group all DocCHTMLTests suites under a parent suite, and run it serially on Linux to avoid crashes. macOS can continue to run them in parallel as it uses a thread-safe XML implementation.

[1]: https://gnome.pages.gitlab.gnome.org/libxml2/html/parser_8h.html#a642085a16c74b8352e92a89c348cc9c4
[2]: https://gnome.pages.gitlab.gnome.org/libxml2/html/parser_8h.html#a7eb4447bac68a016b428f51cc402369f

A recent occurrence of this: https://github.com/swiftlang/swift-docc/actions/runs/30365670915/job/90296159959?pr=1605

### DIFF
--- a/Tests/DocCHTMLTests/DocCHTMLTestSuites.swift
+++ b/Tests/DocCHTMLTests/DocCHTMLTestSuites.swift
@@ -1,0 +1,34 @@
+/*
+ This source file is part of the Swift.org open source project
+
+ Copyright (c) 2026 Apple Inc. and the Swift project authors
+ Licensed under Apache License v2.0 with Runtime Library Exception
+
+ See https://swift.org/LICENSE.txt for license information
+ See https://swift.org/CONTRIBUTORS.txt for Swift project authors
+*/
+
+import Testing
+
+/// A parent suite that groups together all of the DocCHTMLTests suites.
+///
+/// On Linux, FoundationXML is backed by libxml2, which isn't thread-safe by default.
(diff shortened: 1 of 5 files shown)


d-ronnqvist (commented):
Do you know robot how frequent these test crashes are? I've been working on rewriting that implementation to no longer create `XMLNode` instances and I'm wondering if we should wait for those changes instead, assuming that this is infrequent and has been happening for a full release already.

snprajwal:
I wouldn't say this is very frequent. However, I see no harm in us merging this for now to make CI more robust, and reverting it once your changes land. I'm also happy to wait if you think you'll put your changes up in a PR soon-ish.

d-ronnqvist (approved):

snprajwal:
@swift-ci test

PR #1610, opened 2026-08-03
Include symbol relationships in markdown export

jrturton:
Bug/issue #, if applicable: 183133511

## Summary

Symbol relationships (conforms to, conforming types, inherited by, inherits from) are not currently included in the exported markdown document. While those relationships are available in the manifest, it is still useful content which should be included in the markdown file.

## Dependencies

N/A

## Testing

1. Generate markdown export from any documented project containing protocol or class relationships by running the `convert` command with the `--enable-experimental-markdown-output` flag. 
2. Open the markdown file for a symbol with protocol or class relationships
3. Observe that there is no **Relationships** section (before this change), or that there is now a **Relationships** section containing sub-sections for the relevant relationships categories (conforms to, conforming types, inherits from, inherited by, as appropriate)

## Checklist

- [x] Added tests
- [ ] Updated documentation if necessary: N/A

### DIFF
--- a/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
+++ b/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
@@ -57,23 +57,8 @@ extension MarkdownOutputSemanticVisitor {
         add(targetIdentifier: target.path, type: type, subtype: subtype)
     }
     
-    func pathAndName(from fallbackTarget: String) -> (path: String, name: String) {
-        let path: String
-        let name: String
-        let components = fallbackTarget.components(separatedBy: ".")
-        if components.count > 1 {
-            path = "/documentation/\(components.joined(separator: "/"))"
-            name = components.last ?? fallbackTarget
-        } else {
-            path = fallbackTarget
-            name = fallbackTarget
-        }
-        return (path, name)
-    }
-    
     mutating func add(fallbackTarget: String, type: MarkdownOutputManifest.RelationshipType, subtype: RelationshipsGroup.Kind?) {
-        let (targetIdentifier, _) = pathAndName(from: fallbackTarget)
-        add(targetIdentifier: targetIdentifier, type: type, subtype: subtype)
+        add(targetIdentifier: fallbackTarget, type: type, subtype: subtype)
     }
     
     mutating func add(targetIdentifier: String, type: MarkdownOutputManifest.RelationshipType, subtype: RelationshipsGroup.Kind?) {
@@ -215,9 +200,7 @@ extension MarkdownOutputSemanticVisitor {
                     if let fallback = symbol.relationships.targetFallbacks[destination] {
                         add(fallbackTarget: fallback, type: .relatedSymbol, subtype: relationshipGroup.kind)
                         markdownWalker.startNewParagraphIfRequired()
-                        let (path, name) = pathAndName(from: fallback)
-                        let link = Link(destination: path, title: name, [InlineCode(name)])
-                        markdownWalker.defaultVisit(link)
+                        markdownWalker.defaultVisit(InlineCode(fallback))
                     }
                 }
             }
--- a/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
+++ b/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
@@ -57,14 +57,22 @@ extension MarkdownOutputSemanticVisitor {
         add(targetIdentifier: target.path, type: type, subtype: subtype)
     }
     
-    mutating func add(fallbackTarget: String, type: MarkdownOutputManifest.RelationshipType, subtype: RelationshipsGroup.Kind?) {
-        let targetIdentifier: String
+    func pathAndName(from fallbackTarget: String) -> (path: String, name: String) {
+        let path: String
+        let name: String
         let components = fallbackTarget.components(separatedBy: ".")
         if components.count > 1 {
-            targetIdentifier = "/documentation/\(components.joined(separator: "/"))"
+            path = "/documentation/\(components.joined(separator: "/"))"
+            name = components.last ?? fallbackTarget
         } else {
-            targetIdentifier = fallbackTarget
+            path = fallbackTarget
+            name = fallbackTarget
         }
+        return (path, name)
+    }
+    
+    mutating func add(fallbackTarget: String, type: MarkdownOutputManifest.RelationshipType, subtype: RelationshipsGroup.Kind?) {
+        let (targetIdentifier, _) = pathAndName(from: fallbackTarget)
         add(targetIdentifier: targetIdentifier, type: type, subtype: subtype)
     }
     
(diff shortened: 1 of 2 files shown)


Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
```diff
@@ -57,14 +57,22 @@ extension MarkdownOutputSemanticVisitor {
...
+        let path: String
+        let name: String
         let components = fallbackTarget.components(separatedBy: ".")
         if components.count > 1 {
-            targetIdentifier = "/documentation/\(components.joined(separator: "/"))"
+            path = "/documentation/\(components.joined(separator: "/"))"
+            name = components.last ?? fallbackTarget
         } else {
-            targetIdentifier = fallbackTarget
+            path = fallbackTarget
```
d-ronnqvist: Question: should this also have a "/documentation" prefix?

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
```diff
@@ -187,14 +195,29 @@ extension MarkdownOutputSemanticVisitor {
...
+                    markdownWalker.startNewParagraphIfRequired()
+                    let link = Link(destination: resolved.path, title: resolved.lastPathComponent, [InlineCode(resolved.lastPathComponent)])
+                    markdownWalker.defaultVisit(link)
+                    
                 case .failure:
                     if let fallback = symbol.relationships.targetFallbacks[destination] {
                         add(fallbackTarget: fallback, type: .relatedSymbol, subtype: relationshipGroup.kind)
+                        markdownWalker.startNewParagraphIfRequired()
+                        let (path, name) = pathAndName(from: fallback)
+                        let link = Link(destination: path, title: name, [InlineCode(name)])
```
d-ronnqvist: If the relationship didn't resolve I don't think this markdown link will go anywhere. Would it make sense to display the target fallback text in as inline-code instead of a link?

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,106 @@ struct MarkdownOutputTests {
...
+        
+        let relationships = [
+            SymbolGraph.Relationship(source: "MO_Conformer", target: "MO_Protocol", kind: .conformsTo, targetFallback: nil),
+            SymbolGraph.Relationship(source: "MO_Conformer", target: "s:SH", kind: .conformsTo, targetFallback: "Swift.Hashable")
+        ]
+        
+        let catalog = catalog(files: [
+            JSONFile(name: "MarkdownOutput.symbols.json", content:
+                    makeSymbolGraph(moduleName: "MarkdownOutput", symbols: symbols, relationships: relationships))
+        ])
```
d-ronnqvist: I think you're alone in this style of predefining arrays that are accessed exactly once when passed to the test helper a few lines below. Also, this "MO_" prefix style doesn't seem to be used elsewhere, not even in past tests you've created. Please try to follow the style of the rest of the repository.

```suggestion
        let catalog = catalog(files: [
            JSONFile(symbolGraph: makeSymbolGraph(moduleName: "ModuleName", symbols: [
                makeSymbol(id: "some-struct-id",   kind: .struct,   pathComponents: ["LocalConformer"]),
                makeSymbol(id: "some-protocol-id", kind: .protocol, pathComponents: ["LocalProtocol"]),
            ], relationships: [
                SymbolGraph.Relationship(source: "some-struct-id", target: "some-protocol-id", kind: .conformsTo, targetFallback: nil),
                SymbolGraph.Relationship(source: "some-struct-id", target: "s:SH",             kind: .conformsTo, targetFallback: "Swift.Hashable"),
            ])
        ])
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,106 @@ struct MarkdownOutputTests {
         #expect(node.markdown.contains(expectedList))
     }
     
+    @Test
+    func conformsToIncludedInExport() async throws {
```
d-ronnqvist:
```suggestion
    func includesRelationshipsSectionWhenSymbolConformsToProtocols() async throws {
```

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,106 @@ struct MarkdownOutputTests {
...
+        let markdown = node.markdown
+        #expect(markdown.contains(title))
+        let link = "\n[`LocalProtocol`](/documentation/MarkdownOutput/LocalProtocol)"
+        #expect(markdown.contains(link))
+        let externalLink = "\n[`Hashable`](/documentation/Swift/Hashable)"
+        #expect(markdown.contains(externalLink))
+    }
+    
+    @Test
+    func conformingTypesIncludedInExport() async throws {
```
d-ronnqvist: This has almost the same setup as the test above (except there's no relationship to an external symbol in this test) except that it's varying the protocol page instead of the conforming structure's page. These things could both be verified in the same test.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,106 @@ struct MarkdownOutputTests {
...
+                    makeSymbolGraph(moduleName: "MarkdownOutput", symbols: symbols, relationships: relationships))
+        ])
+        
+        let (node, _) = try await markdownOutput(catalog: catalog, path: "LocalSuper")
+        let title = RelationshipsGroup(kind: .inheritedBy, destinations: []).sectionTitle
+        let markdown = node.markdown
+        #expect(markdown.contains(title))
+        let link = "\n[`LocalSub`](/documentation/MarkdownOutput/LocalSub)"
+        #expect(markdown.contains(link))
+    }
```
d-ronnqvist, line 929: These two tests have the same setup but are verifying different pages (the subclass page or the superclass page). It would help the reader understand that these are the same behaviors being verified if they were the same test.

d-ronnqvist (changes requested):

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
jrturton: Renamed to `protocolRelationshipsIncludedInExport()` and merged with the other test as suggested below

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
jrturton: changed all symbol identifiers to `this-sort-of-thing-id`, inlined all the arrays

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
jrturton: What I'm trying to cover here is the situation where a symbol is declared as, e.g. `Hashable`, but at the point of documentation compilation, only the symbol graph for the local framework is available. In this situation, linking to documentation/swift/hashable is preferable to an inline code mention of `Swift.Hashable` - or do you disagree?

d-ronnqvist: I don't think it's ever good to create a link to a page that doesn't exist. In other output formats, DocC doesn't create links for relationships to external symbols that it couldn't resolve in either a dependency archive or from another external source.

d-ronnqvist: That said; I _wouldn't_ consider this to be blocking

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
d-ronnqvist: @jrturton The only blocking feedback is this question whether it's correct to omit the "/documentation" prefix in this case.

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
jrturton: One use case for the markdown export is to generate output locally to amend or extend an existing body of documentation. In those cases, we would expect the relationships not to resolve, but would like to keep the link formatting consistent. 

Perhaps (and this answers your question above?) if we can infer a path (e.g. `Swift.Hashable`) from the fallback target, we carry on with that, otherwise the relationship is rendered as inline code instead of a link?

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
jrturton: See below, perhaps for a single-component fallback target, we should render in code voice rather than attempt a link. I don't think `/documentation/something` can be a valid symbol link?

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
d-ronnqvist: I don't understand what you mean by this. The Markdown output is supplementary to the JSON output. Both are built at the same time (in the same command). If the rendered page doesn't form a link for the `Swift.Hashable` conformance, isn't that what the Markdown files should be consistent with?

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test

PR #1612, opened 2026-08-04
Don't add section headers for markdown sections that turn out to have no content

jrturton:
Bug/issue #, if applicable: 183308686

## Summary

When generating markdown exports, some documents have section headings with no content. This is because some kinds of content (e.g. HTML, comments) are excluded from output. This PR amends the markdown export to not include these empty section headers by checking that content was actually added during processing of a section.

## Dependencies

N/A

## Testing

Define a function with markdown-exlcluded content forming the entire discussion: 

```
/// Does things that can be described very briefly.
    ///
    /// - Parameters:
    ///   - code: The code you pass in.
    ///
    ///   @Comment { this is a comment }
    public static func myFunction(with code: String) {
    }
```

Generate markdown export for this symbol using `docc convert` with `--enable-experimental-markdown-export`

Before this change, the export will have an empty `## Discussion` heading at the end. After this change, it will not. 

## Checklist

- [x] Added tests
- [ ] Updated documentation if necessary:  N/A

### DIFF
--- a/Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
+++ b/Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
@@ -769,8 +769,13 @@ struct MarkdownOutputTests {
     @Test
     func sectionsThatAreEmptyAfterFilteringDoNotHaveHeadingsAdded() async throws {
         let catalog = catalog(files: [
-            JSONFile(name: "MarkdownOutput.symbols.json", content: makeSymbolGraph(moduleName: "MarkdownOutput", symbols: [
-                makeSymbol(id: "MarkdownSymbol.myFunction", kind: .method, pathComponents: ["MarkdownSymbol", "myFunction(_:)"], docComment: """
+            JSONFile(
+                name: "MarkdownOutput.symbols.json",
+                content: makeSymbolGraph(
+                    moduleName: "MarkdownOutput",
+                    symbols: [
+                        makeSymbol(id: "markdown-symbol-id", kind: .struct, pathComponents: ["MarkdownSymbol"]),
+                        makeSymbol(id: "markdown-symbol-my-function-id", kind: .method, pathComponents: ["MarkdownSymbol", "myFunction(_:)"], docComment: """
                     Everything is described in the abstract.
                     
                     - Parameters:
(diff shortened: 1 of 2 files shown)


Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,28 @@ struct MarkdownOutputTests {
         #expect(node.markdown.contains(expectedList))
     }
     
+    @Test
+    func sectionsThatAreEmptyAfterFilteringDoNotHaveHeadingsAdded() async throws {
+        let catalog = catalog(files: [
+            JSONFile(name: "MarkdownOutput.symbols.json", content: makeSymbolGraph(moduleName: "MarkdownOutput", symbols: [
+                makeSymbol(id: "MarkdownSymbol.myFunction", kind: .method, pathComponents: ["MarkdownSymbol", "myFunction(_:)"], docComment: """
```
d-ronnqvist: I don't think this is a valid symbol because the container doesn't exist in the symbol graph file. This sort of "sparse" symbol graph should only be passed from the ConvertService (where DocC accounts for various  odd behaviors) but this isn't coming from the convert service.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -766,6 +766,28 @@ struct MarkdownOutputTests {
         #expect(node.markdown.contains(expectedList))
     }
     
+    @Test
+    func sectionsThatAreEmptyAfterFilteringDoNotHaveHeadingsAdded() async throws {
```
d-ronnqvist, line 820: Does this behavior apply to all sections or only to the automatic Discussion / Overview heading?

If it applies to all sections it would be good to update the test to cover the behavior of another empty section. If not, then it would be good to rename the test to be more specific so that it's clearer what behavior is being verified.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
jrturton: Updated

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
jrturton: It applies to all `Section`s. Updated the content of the test to reflect this

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

d-ronnqvist:
@swift-ci please test macOS

d-ronnqvist:
@swift-ci please test

PR #1617, opened 2026-08-06
Consider all availability variants before raising warning about DeprecationSummary

d-ronnqvist:
Bug/issue #, if applicable: rdar://184145251

## Summary

This fixes a false positive where the warning about using `@DeprecationSummary` for an available symbol didn't consider all availability variants if DocC was passed a collection of platform-specific symbol graph files (for the same module) in the same build where each symbol graph file only contains the in-source availability annotations for that platform.

Before this change, the node would only look at one variant of the availability which could lead to a false positive warning about the symbol being available if it was available for _that_ platform. This warning was non-deterministic because it wasn't deterministic which of the availability variants the node looked at.

## Dependencies

None

## Testing

I'm not aware of any quick and easy way to create this setup but this is a tedious and fairly complicated way:

- Create a project with one symbol that's available for one platform.
    - Add a `@DeprecationSummary` for 
- Generate a symbol graph file for that platform (for example by building documentation and discarding the documentation and setting aside the intermediate symbol graph file for that platform).
- For a number of other platforms:  
  - Update the in-source availability annotation to be for _that_ platform
  - For _only some_ of the platforms, 
  - Generate, and set aside, a new symbol graph for that platform.
- Call `docc convert` and pass it all the gathered symbol graph files for all those platforms (each with only availability information for that platform)
  - There shouldn't be a warning about the `@DeprecationSummary` because the symbol is deprecated for one (or some of those platforms)
- Repeat the `docc convert` a few times (you don't need to regather the symbol graph files). 
  - There shouldn't be a `@DeprecationSummary` warning when rebuilding with the same inputs.

## Checklist

- [x] Added tests

### DIFF
--- a/Tests/SwiftDocCTests/Model/DeprecationSummaryTests.swift
+++ b/Tests/SwiftDocCTests/Model/DeprecationSummaryTests.swift
@@ -424,6 +424,49 @@ struct DeprecationSummaryTests {
         let topicRef = try #require(deprecatedRef, "Expected to find a topic render reference for SomeTypeAlias")
         #expect(topicRef.isDeprecated == true, "Symbol should be marked as deprecated with a message")
     }
+    
+    @Test(arguments: Self.allMainPlatforms)
+    func doesNotWarnAboutDeprecationSummaryIfSymbolIsDeprecatedForOnePlatform(_ deprecatedPlatform: SymbolGraph.Platform) async throws {
+        let name = "requestImageDataForAsset:options:resultHandler:"
+        
+        let catalog = Folder(name: "unit-test.docc") {
+            for (number, platform) in zip(1..., Self.allMainPlatforms) {
+                JSONFile(name: "SomeModule-\(number).symbols.json", content: makeSymbolGraph(moduleName: "SomeModule", platform: platform, symbols: [
+                    makeSymbol(id: "some-symbol-id", kind: .protocol, pathComponents: ["SomeClass", name], docComment: "Nothing", availability: [
+                        .init(domainName: platform.name, introduced: .init(major: 1, minor: 2, patch: 3), deprecated: platform == deprecatedPlatform ? .init(major: 2, minor: 3, patch: 4) : nil)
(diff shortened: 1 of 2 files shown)


d-ronnqvist:
@swift-ci please test

d-ronnqvist:
It looks like the various GitHub action CI checks are failing with `Error: Service Unavailable`

snprajwal:
There is an ongoing incident with GitHub Actions: https://www.githubstatus.com/incidents/qcvjkzcs7j74

Tests/SwiftDocCTests/Model/DeprecationSummaryTests.swift
```diff
@@ -424,6 +424,49 @@ struct DeprecationSummaryTests {
...
         #expect(topicRef.isDeprecated == true, "Symbol should be marked as deprecated with a message")
     }
+    
+    @Test(arguments: Self.allMainPlatforms)
+    func doesNotWarnAboutDeprecationSummaryIfSymbolIsDeprecatedForOnePlatform(_ deprecatedPlatform: SymbolGraph.Platform) async throws {
+        let name = "requestImageDataForAsset:options:resultHandler:"
+        
+        let catalog = Folder(name: "unit-test.docc") {
+            for (number, platform) in zip(1..., Self.allMainPlatforms) {
+                JSONFile(name: "SomeModule-\(number).symbols.json", content: makeSymbolGraph(moduleName: "SomeModule", platform: platform, symbols: [
```
snprajwal, line 434: Question: why do we use a number here for the suffix instead of `platform.name`?

snprajwal (approved):

Tests/SwiftDocCTests/Model/DeprecationSummaryTests.swift
d-ronnqvist: Mostly just to dispel any notion that the platform name in the file name would have any significance to how it's processed.

PR #1619, opened 2026-08-06
Set "Tool" eyebrow and rename DisplayName to "Documentation (DocC)"

heckj:
## Summary
- Adds `@TitleHeading("Tool")` to the `docc` root page, so the page shows a "Tool" eyebrow above its title.
- Updates `@DisplayName` from "DocC" to "Documentation (DocC)" to match the label already used for this catalog in the combined Swift.org documentation's "Tools" navigation group.

### DIFF
--- a/Sources/docc/DocCDocumentation.docc/DocC Documentation.md
+++ b/Sources/docc/DocCDocumentation.docc/DocC Documentation.md
@@ -1,7 +1,8 @@
 # ``docc``
 
 @Metadata {
-    @DisplayName("DocC")
+    @DisplayName("Documentation (DocC)")
+    @TitleHeading("Tool")
 }
 
 Produce rich API reference documentation and interactive tutorials for your Swift framework or package.


Sources/docc/DocCDocumentation.docc/DocC Documentation.md
```diff
@@ -1,7 +1,8 @@
 # ``docc``
 
 @Metadata {
-    @DisplayName("DocC")
+    @DisplayName("Documentation (DocC)")
+    @TitleHeading("Tool")
```
d-ronnqvist, line 5: The [live page](https://www.swift.org/documentation/docc/) already displays "Tool" as the title heading. Is there a particular reason for customizing the header like this instead of the way it's customized today?

Sources/docc/DocCDocumentation.docc/DocC Documentation.md
```diff
@@ -1,7 +1,8 @@
 # ``docc``
 
 @Metadata {
-    @DisplayName("DocC")
+    @DisplayName("Documentation (DocC)")
```
d-ronnqvist, line 4: What's the reason for this change? If it's to indicate that DocC is a documentation tool then I feel that the abstract is a better place to convey that information, which I feel that it's sufficiently clear about today:

> Produce rich API reference documentation and interactive tutorials for your Swift framework or package.

Sources/docc/DocCDocumentation.docc/DocC Documentation.md
heckj: Primarily for consistency - this catalog was setting it via the Info.plist if I recall correctly, I'm making a swath of PRs to tweak content for an upcoming combined Swift.org docs collection, and these updates are to copyedit it into consistency with that presentation. 

I'm happy to drop the `@TitleHeading("Tool")` here if you'd prefer, since it's already accommodated elsewhere.

Sources/docc/DocCDocumentation.docc/DocC Documentation.md
heckj: It's to provide consistency in copyediting for a combined presentation of the swift.org. You can see the prototype at https://heckj.github.io/docs/main/documentation/. I'd previously tweaked up index.json in that combined presentation so the left nav was more scannable for a combined set of content, and we wanted to push those changes so they were consistent down into the content itself as well.

Outside of this PR, I've sent an email around to a number of folks who are (or already have) seen similar PRs to explain the broader set, since it may be easier to see in the broader, holistic context of that combined set of docs and it's presentation than as stand-alone PRs to repositories.

PR #1621, opened 2026-08-07
Fix the low hanging fruit of strict-concurrency warnings

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This fixes the very easiest strict-concurrency warnings that are raised when `.enableExperimentalFeature("StrictConcurrency")` is added to the Swift settings for `SwiftDocC` and `DocCCommandLine`. 

Note: I'm sure that there are more types that can safely be marked `Sendable` but I didn't see any warnings about those types so I didn't go through any additional types to audit them for sendability.

## Dependencies

None

## Testing

- Include `.enableExperimentalFeature("StrictConcurrency")` in the Swift settings for the 
- Build docc and count the number of compiler warnings
- Checkout the "main" branch and repeat the same two steps from above
- Compare the number of warnings between this branch and main. 
  - The number of warnings should have gone down by a lot (over a hundred)

## Checklist

- ~[ ] Added tests~ There's nothing 
- ~[ ] Updated documentation if necessary`

### DIFF
--- a/Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
+++ b/Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -11,7 +11,7 @@
 /// Resolves and validates the arguments needed to enable the documentation coverage feature.
 ///
 /// These options are used by the `Convert` subcommand.
-public struct DocumentationCoverageOptions {
+public struct DocumentationCoverageOptions: Sendable {
     public init(
         level: DocumentationCoverageLevel,
         kindFilterOptions: KindFilterOptions = []) {
@@ -54,7 +54,7 @@ extension DocumentationCoverageOptions {
 }
 
 /// Specifies whether the documentation coverage feature is enabled and, if it is, what amount of specificity is selected.
-public enum DocumentationCoverageLevel: String, Codable, CaseIterable {
+public enum DocumentationCoverageLevel: String, Codable, CaseIterable, Sendable {
     /// No documentation coverage data should be emitted and no documentation coverage information should be displayed in console
     case none
     /// Documentation coverage data should be emitted and a high-level summary should be displayed in console
@@ -67,7 +67,7 @@ extension DocumentationCoverageOptions {
 
     /// Represents kinds to select and display documentation coverage statistics for.
     /// Note: This enum is not meant to be persisted between runs
-    public struct KindFilterOptions: OptionSet, Hashable, CustomDebugStringConvertible {
+    public struct KindFilterOptions: OptionSet, Hashable, CustomDebugStringConvertible, Sendable {
 
         public typealias RawValue = Int
         public let rawValue: Int
--- a/Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
+++ b/Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
@@ -25,9 +25,11 @@ public struct DocumentationCoverageOptions: Sendable {
     }
 
     /// An instance configured to represent the choice not to produce any documentation coverage artifacts or output.
-    public static var noCoverage: DocumentationCoverageOptions = DocumentationCoverageOptions(
-        level: .none,
-        kindFilterOptions: [])
+    public static var noCoverage: DocumentationCoverageOptions {
+        get { DocumentationCoverageOptions(level: .none, kindFilterOptions: []) }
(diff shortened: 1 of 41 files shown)


Sources/SwiftDocC/Infrastructure/Workspace/DocumentationWorkspaceDataProvider.swift
```diff
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
```
QuietMisdreavus: Not sure if updating the year here was an accident or came from running `update-license-comments`, but there were no other edits in this file.

QuietMisdreavus (approved):
Thanks for moving this forward!

QuietMisdreavus:
@swift-ci Please test

Sources/SwiftDocC/Infrastructure/Bundle Assets/DataAssetManager.swift
```diff
@@ -297,10 +297,12 @@ public struct DataTraitCollection: Hashable, Codable {
...
-    public static var allCases: [DataTraitCollection] = {
-        return UserInterfaceStyle.allCases.flatMap { style in DisplayScale.allCases.map { .init(userInterfaceStyle: style, displayScale: $0)}}
-    }()
-    
+    public static var allCases: [DataTraitCollection] {
+        get { _allCases }
+        @available(*, deprecated, message: "Create a new value instead of modifying this one. This deprecated setter will be removed after 6.5 is released.")
+        set { _allCases = newValue }
+    }
+    private static var _allCases: [DataTraitCollection] = UserInterfaceStyle.allCases.flatMap { style in DisplayScale.allCases.map { .init(userInterfaceStyle: style, displayScale: $0)}}
```
snprajwal, line 305: Does this remove the diagnostic here? AFAIU, it just moves it into a private var instead of a public one, but it's still a mutable static var. I didn't compile this branch, but an equivalent example for this case still emits a concurrency warning.

Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
```diff
@@ -25,9 +25,11 @@ public struct DocumentationCoverageOptions {
     }
 
     /// An instance configured to represent the choice not to produce any documentation coverage artifacts or output.
-    public static var noCoverage: DocumentationCoverageOptions = DocumentationCoverageOptions(
-        level: .none,
-        kindFilterOptions: [])
+    public static var noCoverage: DocumentationCoverageOptions {
+        get { DocumentationCoverageOptions(level: .none, kindFilterOptions: []) }
+        @available(*, deprecated, message: "Create a new value instead of modifying this one. This deprecated setter will be removed after 6.5 is released.")
```
snprajwal: (non-blocking) It might be valuable to indicate that setting this variable is now a no-op:


```suggestion
        @available(*, deprecated, message: "Setting this value has no effect; create a new value instead. This value will become read-only after 6.5 is released.")
```

(same comment applies to other deprecations added below in this PR)

Sources/SwiftDocC/Infrastructure/Bundle Assets/DataAssetManager.swift
d-ronnqvist: This _doesn't_ resolve the warning but we can't solve this specific one without a source breaking change. By deprecating the setter we enable ourselves to solve this warning in the future after we've changed this to a read only property.

Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
d-ronnqvist: I don't think there is any real use case for modifying this property (or the other similar one). 

I would almost consider it a conceptual bug fix of the API that it would no longer be possible to modify the `noCoverage` value to represent a configuration that specifies "some" coverage.

snprajwal: Yeah, I was just thinking if the wording could be more explicit for downstream consumers, even if they are incorrectly using the API. Since it's public, it _might_ be possible, even if the chances are low. In any case, I don't have a strong opinion, this is not blocking.

snprajwal (approved):

d-ronnqvist:
@swift-ci please test

Sources/SwiftDocC/Coverage/DocumentationCoverageOptions.swift
d-ronnqvist: Rephrased in https://github.com/swiftlang/swift-docc/pull/1621/commits/949a53901cfcf44b7a170192ce52010e8d03ea56

Sources/SwiftDocC/Infrastructure/Workspace/DocumentationWorkspaceDataProvider.swift
d-ronnqvist: Fixed in https://github.com/swiftlang/swift-docc/pull/1621/commits/198504b9357c864a31f0314ed310d4e7c792b89d

d-ronnqvist:
@swift-ci please test

Issue #1282, opened 2025-08-26
Support DocC on Windows


compnerd:
- [ ] NIO dependency for Preview Server does not support Windows
- [ ] Need to support "invalid" characters in file names (i.e. `:`)

compnerd:
@snprajwal btw, it would be good to get the CMake build in CI so that we can start shipping DocC again (even if it doesn't work atm).

PR #1623, opened 2026-08-11
Add CMake build on Windows in GitHub Actions

snprajwal:
The Swift toolchain on Windows uses CMake to build the toolchain components. This requires us to maintain working CMake build manifests in DocC, and ensure that CMakeLists.txt is updated when files are added or removed. Add a CI workflow that performs a CMake build on Windows. This should unblock shipping DocC on Windows [1], and also allow consumers of the DocC library on Windows (like SourceKit-LSP) to link against it [2].

Some of DocC's dependencies have had changes that are required to support a CMake build:

- swift-lmdb does not support CMake builds in the pinned version, and added support in Dec 2024 [3].
- swift-crypto previously vendored swift-asn1 for downstream consumers as part of its CMake build, and now requires consumers to provide the package themselves [4].

Swift CI for this repository still uses Swift 6.0, preventing us from jumping to the latest minor version for these dependencies. Conservatively update them to the latest version that supports Swift 6.0 and also enables the CMake build.

[1]: https://github.com/swiftlang/swift-docc/issues/1282#issuecomment-5184067246
[2]: https://github.com/swiftlang/swift-docc/pull/818#issuecomment-2486603033
[3]: https://github.com/swiftlang/swift-lmdb/commit/b77c1ac8c4c8e52cbddd890236a36736f445715e
[4]: https://github.com/apple/swift-crypto/commit/af6be725f5c1e7356878d3cc6a8fdb9c764e14c5


compnerd (approved):
The extracted script feels a bit over abstract to me. It seems that we should simply be able to have that within the GHA definition itself. 🤷‍♂️

snprajwal:
> The extracted script feels a bit over abstract to me. It seems that we should simply be able to have that within the GHA definition itself. 🤷‍♂️

That's fair, I used an external script to avoid any gotchas with having multi-line scripts within the YAML, but we probably won't modify it enough to warrant that. I've moved it into the YAML, and retained just the one build helper to make it _slightly_ less verbose.

snprajwal:
Swift CI is still on Swift 6.0, making it troublesome to bump dependencies. I've conservatively bumped the few deps we absolutely need while maintaining support for 6.0.

snprajwal:
@swift-ci test

PR #1624, opened 2026-08-14
Deterministically fold platform availabilities

snprajwal:
Bug/issue #, if applicable: rdar://184277552

## Summary

Availability items are accumulated per platform into a language trait that collapses availabilities for each interface language. The availability is processed by iterating a dictionary, whose ordering is not stable, leading to platform availability varying based on the first processed selector for that platform. Sort availabilities with the same comparator as used to identify the default comparator, to ensure that the accumulated availabilities are consistent.

## Dependencies

N/A

## Testing

1. Create a catalog with symbol graphs for two different platforms.
2. Add a symbol to both symbol graphs, where it is marked unconditionally available for platform A on one and unconditionally unavailable for platform A on the other (can be done with `#if os(<os>)` in source)
3. Build the catalog multiple times. Previously, a platform would appear and disappear. It is now consistently included.

## Checklist

- [x] Added tests

### DIFF
--- a/Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
+++ b/Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
@@ -1,7 +1,7 @@
 /*
  This source file is part of the Swift.org open source project
 
- Copyright (c) 2021-2025 Apple Inc. and the Swift project authors
+ Copyright (c) 2021-2026 Apple Inc. and the Swift project authors
  Licensed under Apache License v2.0 with Runtime Library Exception
 
  See https://swift.org/LICENSE.txt for license information
@@ -558,7 +558,7 @@ extension Symbol {
     }
 
     func mergeDeclarations(unifiedSymbol: UnifiedSymbolGraph.Symbol) throws {
-        for (selector, mixins) in unifiedSymbol.mixins {
+        for (selector, mixins) in unifiedSymbol.mixins.sortedBySelector() {
             if let mergingDeclaration = mixins[SymbolGraph.Symbol.DeclarationFragments.mixinKey] as? SymbolGraph.Symbol.DeclarationFragments {
                 let availability = mixins[SymbolGraph.Symbol.Availability.mixinKey] as? SymbolGraph.Symbol.Availability
                 let alternateSymbols = mixins[SymbolGraph.Symbol.AlternateSymbols.mixinKey] as? SymbolGraph.Symbol.AlternateSymbols
@@ -571,7 +571,7 @@ extension Symbol {
     /// Merge the different availability variants defined in the unified symbol,
     /// and update the availability of the canonical symbol to consider all the different availability mixins instead of only the first one.
     func mergeAvailabilities(unifiedSymbol: UnifiedSymbolGraph.Symbol) {
-        for (selector, mixins) in unifiedSymbol.mixins {
+        for (selector, mixins) in unifiedSymbol.mixins.sortedBySelector() {
             let trait = DocumentationDataVariantsTrait(for: selector)
             if let unifiedSymbolAvailability = mixins[SymbolGraph.Symbol.Availability.mixinKey] as? SymbolGraph.Symbol.Availability {
                 for availabilityItem in unifiedSymbolAvailability.availability {
--- a/Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
+++ b/Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
@@ -20,32 +20,7 @@ extension UnifiedSymbolGraph.Symbol {
     private func defaultSelector(
         in selectors: some Sequence<UnifiedSymbolGraph.Selector>
     ) -> UnifiedSymbolGraph.Selector? {
-        // Return the default selector based on the ordering defined below.
-        return selectors.sorted { lhsSelector, rhsSelector in
-            switch (lhsSelector.interfaceLanguage, rhsSelector.interfaceLanguage) {
-                
-            // If both selectors are Swift, pick the one that has the most matching platforms with the other selectors.
-            case ("swift", "swift"):
-                let nonSwiftSelectors = selectors.filter { $0.interfaceLanguage != "swift" }
-                
-                let lhsMatchingPlatformsCount = nonSwiftSelectors.filter { $0.platform == lhsSelector.platform }.count
-                let rhsMatchingPlatformsCount = nonSwiftSelectors.filter { $0.platform == rhsSelector.platform }.count
-                
-                if lhsMatchingPlatformsCount == rhsMatchingPlatformsCount {
-                    // If they have the same number of matching platforms, use the hierarchical platform order.
-                    return PlatformName.areInIncreasingOrder(lhsSelector.platform, rhsSelector.platform)
-                }
-                
-                return lhsMatchingPlatformsCount > rhsMatchingPlatformsCount
-            case ("swift", _):
-                return true
-            case (_, "swift"):
-                return false
-            default:
-                // Use the hierarchical platform order
-                return PlatformName.areInIncreasingOrder(lhsSelector.platform, rhsSelector.platform)
-            }
-        }.first
+        return selectors.min(by: UnifiedSymbolGraph.Selector.Ordering(among: selectors).areInIncreasingOrder)
     }
 
     func symbol(forSelector selector: UnifiedSymbolGraph.Selector?) -> SymbolGraph.Symbol? {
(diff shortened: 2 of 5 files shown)


Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
```diff
@@ -153,7 +153,7 @@ struct PathHierarchyTests_new {
         #expect(links["some-symbol-id"] == "/ModuleName/\(symbolName)", "Links allow any special characters")
     }
 
-    // This verifies determinism in a previously non-deterministic behavior that cannot be reproduced reliably in a test.
+    // This verifies determinism in previously non-deterministic behavior that cannot be reproduced reliably in a test.
```
anferbui, line 156: The previous version was also grammatically correct, depending on whether we're verifying one behaviour ("in a non-deterministic behavior") or behaviour in general ("in non-deterministic behavior").

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -125,6 +100,53 @@ extension UnifiedSymbolGraph.Symbol {
...
 
+extension UnifiedSymbolGraph.Selector {
+    /// Returns a comparator that performs a cascading ordering of selectors.
+    /// A Swift selector is preferred over a non-Swift one.
+    /// If both selectors are for Swift, the value with the most non-Swift selectors for its platform is chosen.
+    /// If it is still a tie, the hierarchical platform order is used.
+    ///
+    /// NOTE: This method is not a comparator in itself. It is designed to return a comparator,
+    /// so that the number of non-Swift comparators per platform is computed once and reused.
+    static func areInIncreasingOrder(
```
anferbui, line 111: I'm unsure about this function name; `PlatformName.areInIncreasingOrder` makes more sense as a name because it returns a Boolean which answers the question of whether they are in increasing order, but this returns a function which is the actual comparator itself. How about calling it something like `increasingOrderComparator(among:)`?

In general I'm finding this function quite confusing and I'm wondering if the sorting logic could be simplified at all. It feels like we're using the wrong abstractions if we're having to look at the other elements of the array like this to be able to sort the array. However I understand that any refactoring here would make this fix unnecessarily large so happy for this to be discussed separately.

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -125,6 +100,53 @@ extension UnifiedSymbolGraph.Symbol {
...
+        among selectors: some Sequence<UnifiedSymbolGraph.Selector>
+    ) -> (UnifiedSymbolGraph.Selector, UnifiedSymbolGraph.Selector) -> Bool {
+        // Precompute the number of non-Swift selectors per platform
+        var nonSwiftSelectorsPerPlatform: [String?: Int] = [:]
+        for selector in selectors where selector.interfaceLanguage != "swift" {
+            nonSwiftSelectorsPerPlatform[selector.platform, default: 0] += 1
+        }
```
anferbui, line 118: Not for changing as part of this PR, but I'm curious: Why are we sorting for most non-Swift platform availabilities rather than the most Swift platform availabilities?

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -571,7 +571,7 @@ extension Symbol {
     /// Merge the different availability variants defined in the unified symbol,
     /// and update the availability of the canonical symbol to consider all the different availability mixins instead of only the first one.
     func mergeAvailabilities(unifiedSymbol: UnifiedSymbolGraph.Symbol) {
-        for (selector, mixins) in unifiedSymbol.mixins {
+        for (selector, mixins) in unifiedSymbol.mixins.sortedBySelector() {
```
anferbui, line 574: I understand how merging the availabilities in a deterministic order is an improvement here, but it also sounds like we might be merging availabilities incorrectly; what is the correct result for this case you mentioned in the PR description?
> Add a symbol to both symbol graphs, where it is marked unconditionally available for platform A on one and unconditionally unavailable for platform A on the other

Should the symbol be conditionally available or unconditionally unavailable for platform A? Now that we're processing availabilities in a consistent order, which one will be chosen? The one with the most non-Swift platform availabilities? (I am also confused as to why this is the criteria for sorting but I've asked that in a separate comment). 
(i.e. do we need a separate fix to how we merge availabilities)

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -125,6 +100,53 @@ extension UnifiedSymbolGraph.Symbol {
...
 }
 
+extension UnifiedSymbolGraph.Selector {
+    /// Returns a comparator that performs a cascading ordering of selectors.
+    /// A Swift selector is preferred over a non-Swift one.
+    /// If both selectors are for Swift, the value with the most non-Swift selectors for its platform is chosen.
+    /// If it is still a tie, the hierarchical platform order is used.
+    ///
+    /// NOTE: This method is not a comparator in itself. It is designed to return a comparator,
+    /// so that the number of non-Swift comparators per platform is computed once and reused.
```
anferbui, line 110:
```suggestion
    /// so that the number of non-Swift selectors per platform is computed once and reused.
```

Tests/SwiftDocCTests/Model/AvailabilityTests.swift
```diff
@@ -325,6 +325,43 @@ struct AvailabilityTests {
...
+                            pathComponents: ["SomeClass"],
+                            availability: [
+                                makeAvailabilityItem(domainName: "iOS", introduced: iOSIntroducedVersion)
+                            ]
+                        )
+                    ]
+                ))
+            }
+        }
+        let context = try await load(catalog: catalog)
```
anferbui, line 354: Can we assert `context.diagnostics.isEmpty` like in other tests in this file?

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
```diff
@@ -558,7 +558,7 @@ extension Symbol {
     }
 
     func mergeDeclarations(unifiedSymbol: UnifiedSymbolGraph.Symbol) throws {
-        for (selector, mixins) in unifiedSymbol.mixins {
+        for (selector, mixins) in unifiedSymbol.mixins.sortedBySelector() {
```
anferbui, line 561: Follow-up work: There's some other places which also iterate over `symbol.mixins`, should we update them to use `.sortedBySelector()` too?
- [`traitSpecificSignatures(_:)`](https://github.com/swiftlang/swift-docc/blob/f04b597c37998c72c54a22b888a932f3762283cf/Sources/SwiftDocC/Model/ParametersAndReturnValidator.swift#L363)
- [`addMissingAvailability(_:)`](https://github.com/swiftlang/swift-docc/blob/f04b597c37998c72c54a22b888a932f3762283cf/Sources/SwiftDocC/Infrastructure/Symbol%20Graph/SymbolGraphLoader.swift#L214)

I don't know if there's an easy way we can override the default iteration logic for `symbol.mixins` to just always use `.sortedBySelector()` so that there is no way for us to forget to do it manually.

anferbui, line 561: Should we add a comment here and below that we're sorting them to ensure deterministic merging behaviour?

Tests/SwiftDocCTests/Infrastructure/PathHierarchyTests.swift
snprajwal: True, I just felt this is a better representation of what the comment intended to convey (which is that it tests previously non-deterministic behaviour that may not be attributed to one specific symptom)

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
snprajwal: A selector is a pair of language + platform. Since we group data on a per-platform basis, we want to count the number of non-Swift selectors *for this platform*. The number of Swift selectors for a given platform is always 1, and counting the number of platforms with Swift selectors sounds incorrect to me, since that provides info coming from across multiple platforms. For additional context, please see https://github.com/swiftlang/swift-docc/pull/93 where this logic was introduced.

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
snprajwal: Since the sort depends on the aggregate of non-Swift selectors for a given platform, we can't make it self-contained in a way where we directly provide a comparator. However, I understand why this appears confusing (and the naming), I can rework this into a local type that then provides a regular comparator.

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
snprajwal: I would think it is implicit that the sort is used to require stable ordering for the process happening within the loop. IMO, there is no value to be gained by stating that it is to ensure determinism, the sort itself sufficiently conveys that.

snprajwal: The latter method does not do any order-sensitive operations, so a sort is not necessary. The former does seem to have latent non-determinism, updating it. Nice one!

Sources/SwiftDocC/Semantics/Symbol/Symbol.swift
snprajwal: In all valid cases, the availability coming from the symbol graph for the relevant platform (i.e. the iOS symbol graph data for iOS) is the authoritative source, and is selected via [`filterItems`](https://github.com/snprajwal/swift-docc/blob/43efcb8a6981d13a008f296a6bf9be2964fcfae0/Sources/SwiftDocC/Utility/SymbolGraphAvailability%2BFilter.swift#L14-L23) within [`mergeDeclaration`](https://github.com/snprajwal/swift-docc/blob/43efcb8a6981d13a008f296a6bf9be2964fcfae0/Sources/SwiftDocC/Semantics/Symbol/Symbol.swift#L548).

The example I've provided can occur only with platforms that do not have their own symbol graphs, e.g. Mac Catalyst or iPadOS. In those cases, I don't think there is a "correct" choice to make, because conflicting info for the same platform is semantically invalid. This PR makes the choice deterministic, but I believe the data itself must be fixed at source; DocC shouldn't do anything.

anferbui: That makes sense to me, thank you for the clarification

anferbui (approved):
Beautiful, thanks!

snprajwal:
@swift-ci test

PR #1625, opened 2026-08-18
Autocurated topics missing from markdown export

jrturton:
Bug/issue #, if applicable: 184720446

## Summary

Autocurated topic sections are missing from the markdown export

## Dependencies

N/A

## Testing

Create markdown (using convert / --enable-experimental-markdown-export) documenting an item that should have auto curated sections, for example, a class with an in-source description,  some methods, instance variables, and initialisers, but no separately documented curation. 

Before the changes in this PR, the markdown export of the symbol will contain the symbol and description, but no reference to the child symbols. The rendered documentation will have those child symbols in groups under a "topics" section. 

After these changes, the rendered documentation and markdown should both contain the child symbols.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
```diff
@@ -67,6 +67,67 @@ extension MarkdownOutputSemanticVisitor {
...
+    /// - Parameters:
+    ///   - groups: The automatically generated task groups to render, in the order they should appear.
+    ///   - hasTopicsHeading: Whether the page already has a "Topics" heading. If it doesn't, and this call renders any
+    ///     groups, a heading is added and this value is set to `true` so that later calls don't add a second one.
+    mutating func visit(automaticTaskGroups groups: [AutomaticCuration.TaskGroup], hasTopicsHeading: inout Bool) {
+        let groups = groups.filter { !$0.references.isEmpty }
+        guard !groups.isEmpty else { return }
+
+        if !hasTopicsHeading {
+            markdownWalker.visit(Heading(level: 2, Text(TopicsSection.title ?? "Topics")))
```
snprajwal: `TopicsSection.title` will always be "Topics" in this case, it's a compile-time constant:

https://github.com/swiftlang/swift-docc/blob/21b4a5321e5bf8bee7bc7d94619da6b5671006b8/Sources/SwiftDocC/Model/Section/Sections/Topics.swift#L17-L18

I also see that we've just used "Topics" exclusively [below](https://github.com/jrturton/swift-docc/blob/c6e06c80c3122519e9a21a549c9deaa0d645fb0b/Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift#L159), so we can possibly do the same here just for consistency:


```suggestion
            markdownWalker.visit(Heading(level: 2, Text("Topics")))
```

Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputSemanticVisitor.swift
```diff
@@ -67,6 +67,67 @@ extension MarkdownOutputSemanticVisitor {
...
+            hasTopicsHeading = true
+        }
+
+        // Copied to a local because the walker requires exclusive access to `self` for the duration of the closure below.
+        let context = self.context
+        markdownWalker.withRenderingLinkList {
+            for group in groups {
+                // Automatically generated task groups always have a title.
+                $0.visit(Heading(level: 3, Text(group.title ?? "Group")))
```
snprajwal: The group title *shouldn't* be empty, but if it is, the current fallback is "Symbols":

https://github.com/swiftlang/swift-docc/blob/21b4a5321e5bf8bee7bc7d94619da6b5671006b8/Sources/SwiftDocC/Infrastructure/Topic%20Graph/AutomaticCuration.swift#L266

It might be better to use that as the fallback here

snprajwal (approved):
LGTM, just a couple of non-blocking comments

PR #1627, opened 2026-08-19
Include declaration variants when exporting markdown

jrturton:
Bug/issue #, if applicable: 166606746

## Summary

In the declaration section of a symbol with platform variants (e.g. a property represented as UIColor on iOS and NSColor on macOS), only a single, non-deterministic variant is included in the markdown export. Variants should be listed in a predictable order like the rendered documentation.

Alternate declarations should also be included, e.g the completion handler and async variants shown in https://developer.apple.com/documentation/avfoundation/avassetwriter/finishwriting(completionhandler:)

## Dependencies

N/A

### Platform Variants

From a framework that has alternate declarations per platform, for example: 

```
/// A public symbol
public struct MyPublicSymbol {
        
    #if os(macOS)
    public func doSomething(_ with: Int = 3) { 
    }
    #else
    public func doSomething(_ with: Int = 2) {
    }
    #endif

}
```

Create the symbol graphs for each platform, e.g.:

```
xcodebuild -scheme MyFramework -destination 'generic/platform=iOS' OTHER_SWIFT_FLAGS="-emit-symbol-graph -emit-symbol-graph-dir $(pwd)/output/ios -active-platform-availability-only" build
```

Rename the symbol graphs to include the platform name and place them in the same directory. Then, build the documentation for the framework, passing in the symbol graph directory and the `--enable-experimental-markdown-export` flag. 

Before this change, the markdown page for the function with alternate declarations will list only one declaration. After the change, the markdown will contain both declarations, in platform priority order, with headings per platform.

### Declaration Variants

In a framework that has objective-c types visible to swift, create an objective-C symbol that would be converted into an asynchronous and completion-based symbol:

```
// Widget.h
  @interface Widget : NSObject
...  
- (void)doSomethingWithCompletionHandler:(void (^)(NSString *name))completionHandler;

  @end

```

Create the documentation for this symbol with the markdown export enabled. Before this change, only the non-asynchronous version of the method will be shown in the markdown declaration. After, both will be. 

## Checklist

- [x] Added tests


Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -53,6 +53,68 @@ struct MarkdownOutputTests {
...
+            makeSymbolGraph(
+                moduleName: "MarkdownOutput",
+                platform: .init(operatingSystem: .init(name: platform)),
+                symbols: [
+                    makeSymbol(id: "markdown-symbol-id", kind: .struct, pathComponents: ["MarkdownSymbol"]),
+                    makeSymbol(id: "markdown-symbol-variants-id", kind: .func, pathComponents: ["MarkdownSymbol", "varies(_:)"], otherMixins: [declaration(name: "varies", type: type)]),
+                    makeSymbol(id: "markdown-symbol-no-variants-id", kind: .func, pathComponents: ["MarkdownSymbol", "notVaries(_:)"], otherMixins: [declaration(name: "notVaries", type: "Int")]),
+                ],
+                relationships: [
+                    SymbolGraph.Relationship(source: "markdown-symbol-func-id", target: "markdown-symbol-id", kind: .memberOf, targetFallback: nil),
```
snprajwal: The `markdown-symbol-func-id` symbol doesn't exist, did you mean:

```suggestion
                    SymbolGraph.Relationship(source: "markdown-symbol-variants-id", target: "markdown-symbol-id", kind: .memberOf, targetFallback: nil),
```

Sources/SwiftDocC/Semantics/Symbol/PlatformDeclarations.swift
```diff
@@ -24,10 +24,28 @@ extension [[PlatformName?]: SymbolGraph.Symbol.DeclarationFragments] {
     func renderDeclarationTokens() -> [DeclarationRenderSection.Token]? {
         mainRenderFragments()?.declarationFragments.renderDeclarationTokens()
     }
+    
+    /// The declarations sorted by platform priority, with the list of platforms per declaration also sorted by priority.
+    func sortedByPlatformPriority() -> [(platforms: Self.Key, declaration: Self.Value)] {
```
snprajwal, line 41: This looks to be mirroring what happens in `DeclarationsSectionTranslator`, where it adds fallback platforms before performing this sort:

https://github.com/swiftlang/swift-docc/blob/b33facd091fa8d444441a6dced75b75a9252ebd9/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift#L222-L223

https://github.com/swiftlang/swift-docc/blob/b33facd091fa8d444441a6dced75b75a9252ebd9/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/DeclarationsSectionTranslator.swift#L287-L294

The markdown would not contain the fallback platforms (iPadOS and Mac Catalyst), we should re-use that logic here. It might be good to move `expandPlatformsWithFallbacks` into a static func in `PlatformName.swift` so you can reuse it here like `PlatformName.expandingFallbacks(platforms)`.

Sources/SwiftDocC/Semantics/Symbol/PlatformName.swift
```diff
@@ -165,3 +165,26 @@ public struct PlatformName: Codable, Hashable, Comparable, Sendable {
...
+    ///
+    /// This function uses the centralized `DefaultAvailability.fallbackPlatforms` mapping to ensure
+    /// consistency with platform expansion logic used throughout the codebase.
+    ///
+    /// For example, when iOS is present in the platforms array, this function adds iPadOS and Mac Catalyst
+    /// if they are not already included.
+    ///
+    /// - Parameter platforms: The original platforms array.
+    /// - Returns: The platforms array with fallback platforms added where applicable.
+    static func expandPlatformsWithFallbacks(_ platforms: [PlatformName?]) -> [PlatformName?] {
```
snprajwal, line 181: Minor, non-blocking: this reads a bit awkwardly given that it's already nested within `PlatformName`. I feel `expandingFallbacks()` or `addingFallbacks()` is less verbose.

We can also have this in the type itself instead of an extension, I don't see a reason to keep this helper separate.

Sources/SwiftDocC/Semantics/Symbol/PlatformDeclarations.swift
```diff
@@ -24,10 +24,40 @@ extension [[PlatformName?]: SymbolGraph.Symbol.DeclarationFragments] {
...
+    func expandingPlatforms() -> Self {
+        let expandedKeyValues = map { platforms, declaration in
+            let expanded = PlatformName.expandPlatformsWithFallbacks(platforms)
+                .sorted {
+                    PlatformName.areInIncreasingOrder($0?.rawValue, $1?.rawValue)
+                }
+            return (expanded, declaration)
+        }
+        return Self(uniqueKeysWithValues: expandedKeyValues)
+    }
```
snprajwal: This converts the dictionary to an array, which is again converted back to a dictionary in the return statement. It also duplicates the sort that happens within `sortedByPlatformPriority`. You can instead use this method to just map to the expanded platforms, and perform all sorting in `sortedByPlatformPriority`. You can move the sorting method into an extension on `[([PlatformName?], SymbolGraph.Symbol.DeclarationFragments)]` to retain the method chaining.

jrturton: As far as I know, you can't use map to update the keys of a dictionary, so you have to go via an array / uniqueKeys. And if we don't sort the platforms here, there is the possibility that we have two sets of platform names that are effectively the same but in a different order, which would lead to duplicates in the final output.

snprajwal: That's correct, my suggestion is to change this method to return an array instead, and operate on that, so we skip the array -> dict -> array conversion that currently happens. Something like:

```
extension [[PlatformName?]: SymbolGraph.Symbol.DeclarationFragments] {
    func addingFallbackPlatforms() -> [(platforms: Self.Key, declaration: Self.Value)] {
        map { (platforms: PlatformName.addingFallbacks($0.key), declaration: $0.value) }
    }
}

extension [(platforms: [PlatformName?], declaration: SymbolGraph.Symbol.DeclarationFragments)] {
    func sortedByPlatformOrder() -> Self {
        map { (platforms: $0.platforms.sorted { PlatformName.areInIncreasingOrder($0?.rawValue, $1?.rawValue) }, declaration: $0.declaration) }
            .sorted { PlatformName.areInIncreasingOrder($0.platforms.first??.rawValue, $1.platforms.first??.rawValue) }
    }
}
```

That way, you can keep the `declaration.addingFallbackPlatforms().sortedByPlatformOrder()` method chain while only going from dict to array once. If you're less picky about retaining the chain, we can just inline this method instead, since it just proxies to the static method in `PlatformName`.

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -53,6 +53,120 @@ struct MarkdownOutputTests {
...
+        # notVaries(_:)
+        
+        ```
+        func notVaries(value: Int)
+        ```
+        """
+        #expect(nonVariantNode.markdown.contains(expectedNonVariant))
+        
+        let (alternateNode, _) = try await markdownOutput(catalog: catalog, path: "MarkdownSymbol/doSomething(completionHandler:)")
+        print(alternateNode.markdown)
```
snprajwal:
```suggestion
```

Sources/SwiftDocC/Semantics/Symbol/PlatformDeclarations.swift
```diff
@@ -26,8 +26,33 @@ extension [[PlatformName?]: SymbolGraph.Symbol.DeclarationFragments] {
...
+        let expandedArray: [(Key, Value)] = compactMap { platforms, value in
+            let expanded = PlatformName.addingFallbacks(platforms)
+                .sorted { PlatformName.areInIncreasingOrder($0?.rawValue, $1?.rawValue) }
+            if keys.contains(expanded) { return nil }
+            keys.insert(expanded)
+            return (expanded, value)
+        }
+        return expandedArray.sorted {
+            PlatformName.areInIncreasingOrder($0.0.first??.rawValue, $1.0.first??.rawValue)
+        }
```
snprajwal: Each platform is condensed into one group, so the groups are all disjoint even after expansion, no need to dedupe here:


```suggestion
        map { platforms, value in
            (PlatformName.addingFallbacks(platforms)
                .sorted { PlatformName.areInIncreasingOrder($0?.rawValue, $1?.rawValue) },
            value)
        }
        .sorted { PlatformName.areInIncreasingOrder($0.0.first??.rawValue, $1.0.first??.rawValue) }
```

snprajwal (approved):
LGTM, just some dead code to remove. Thanks for iterating on this!

PR #1628, opened 2026-08-20
Use subheading when creating symbol links in markdown topics lists

jrturton:
Bug/issue #, if applicable: 185215463

## Summary

When outputting a topics list, only the title of the symbol is included in the link name. It is more useful, and closer matches the rendered documentation, if the subheading declaration is used instead, e.g.: 

```
[`var property: Int](/documentation/Framework/Type/property)

The property's abstract
```

Rather than 

```
[`property`](/documentation/Framework/Type/property)

The property's abstract
```

## Dependencies

N/A

## Testing

Export markdown using `convert` with the `--enable-experimental-markdown-export` flag enabled. Examine the markdown for any symbol with members or methods. Before this change, the markdown will only contain the name of the member or method in the link. After, it will have a declaration-style name that matches what you see in the rendered documentation.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Model/MarkdownOutput/Translation/MarkdownOutputMarkdownWalker.swift
```diff
@@ -240,10 +240,11 @@ extension MarkdownOutputMarkupWalker {
                 
         if isRenderingLinkList {
             linkListAbstract = (doc.semantic as? Symbol)?.abstract
-            if let fragments = symbol.navigator {
+            if let fragments = symbol.subHeadingVariants[.swift] ?? symbol.declaration.mainRenderFragments()?.declarationFragments {
```
snprajwal, line 243: Nice one, FYI the navigator uses a reduced subset of the subheading to prevent the entries from becoming too long, so it omits some information.

snprajwal (approved):

Tests/SwiftDocCTests/Rendering/Markdown/MarkdownOutputTests.swift
```diff
@@ -289,6 +289,56 @@ struct MarkdownOutputTests {
...
+                                    .init(kind: .text, spelling: " }", preciseIdentifier: nil),
+                                ],
+                                subHeading: [
+                                    .init(kind: .keyword, spelling: "var", preciseIdentifier: nil),
+                                    .init(kind: .text, spelling: " ", preciseIdentifier: nil),
+                                    .init(kind: .identifier, spelling: "property", preciseIdentifier: nil),
+                                    .init(kind: .text, spelling: ": ", preciseIdentifier: nil),
+                                    .init(kind: .typeIdentifier, spelling: "Int", preciseIdentifier: "s:Si"),
+                                ],
+                            ),
```
snprajwal: Trailing comma support for certain cases was added [in Swift 6.1](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0439-trailing-comma-lists.md), while Swift CI still uses Swift 6.0 on macOS 😬 removing this should let CI pass


```suggestion
                            )
```

edit: I just noticed this comma is inside a list, which is supported syntax and shouldn't have broken CI (Linux doesn't complain on Swift 5.9), not sure what's going on ¯\_(ツ)_/¯

jrturton: That's the thing it's complaining about in the logs, so let's see what happens...

PR #1629, opened 2026-08-21
Deterministically render conformance lists

snprajwal:
Bug/issue #, if applicable: rdar://184277688

## Summary

Relationships are processed by iterating over an unordered sets, and the destinations are sorted by title, which does not guarantee ordering if titles are identical.

Use a strict weak ordering to sort relationship destinations by sorting destinations with identical titles using their reference URLs. Also sort selectors before iterating over relationships for variants.

## Dependencies

N/A

## Testing

1. Create a catalog with a symbol graph containing a type that conforms to multiple protocols, where two protocols share titles (identical names in different modules)
3. Build the catalog multiple times. Previously, the listed conformances would flip order between builds. With this patch, they are ordered consistently.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift
```diff
@@ -1261,6 +1261,21 @@ public struct RenderNodeTranslator: SemanticVisitor {
...
+        _ lhs: TopicReference,
+        _ rhs: TopicReference,
+        titles: [TopicReference: String]
+    ) -> Bool {
+        let lhsTitle = titles[lhs] ?? ""
+        let rhsTitle = titles[rhs] ?? ""
+        if lhsTitle != rhsTitle {
+            return lhsTitle < rhsTitle
+        }
+        return (lhs.url?.absoluteString ?? "") < (rhs.url?.absoluteString ?? "")
```
anferbui: In other places in DocC, we raise an `assertionFailure`  if our assumptions are incorrect, I don't know if we want to follow this pattern here as well. It would be really unexpected if these topic references did not have a reference URL. 
```suggestion
        guard let lhsReferenceUrl = lhs.url?.absoluteString, let rhsReferenceUrl = rhs.url?.absoluteString else {
            assertionFailure("<explanatory message>")
        }
        return lhsReferenceUrl < rhsReferenceUrl
```

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -153,6 +153,18 @@ extension Dictionary where Key == UnifiedSymbolGraph.Selector {
...
 
+extension SymbolGraph.Relationship {
+    static func areInIncreasingOrder(
+        _ lhs: SymbolGraph.Relationship,
+        _ rhs: SymbolGraph.Relationship
+    ) -> Bool {
+        if lhs.source != rhs.source { return lhs.source < rhs.source }
+        if lhs.target != rhs.target { return lhs.target < rhs.target }
+        if lhs.kind.rawValue != rhs.kind.rawValue { return lhs.kind.rawValue < rhs.kind.rawValue }
+        return (lhs.targetFallback ?? "") < (rhs.targetFallback ?? "")
```
anferbui: Similar point here about raising an assertion failure

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -153,6 +153,18 @@ extension Dictionary where Key == UnifiedSymbolGraph.Selector {
     }
 }
 
+extension SymbolGraph.Relationship {
+    static func areInIncreasingOrder(
```
anferbui, line 157: Nit: This is undocumented, unlike our other `areInIncreasingOrder` functions which have some docs.

Tests/SwiftDocCTests/Rendering/RenderNodeTranslatorTests.swift
```diff
@@ -1626,4 +1626,57 @@ struct RenderNodeTranslatorTests_new {
         let renderedReference = try #require(renderNode.references[memberReference.absoluteString] as? TopicRenderReference)
         #expect(renderedReference.conformance?.constraints.plainText == "Self is Bar.")
     }
+
+
```
anferbui, line 1630: Nit: Two blank lines
```suggestion

```

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
anferbui, line 164: We're ignoring `mixins` here, is that intentional? Could we add a comment if so?

As a side note, as long as we have any field which we don't take into account, that will be a potential source of non-determinism when all other fields are equal. It would be great if we could take `mixins` into account here, because it's part of the `Equatable` conformance for this type:
https://github.com/swiftlang/swift-docc-symbolkit/blob/c1f9484553d29afb981b95458b37c1c3d3e1bcd7/Sources/SymbolKit/SymbolGraph/Relationship/Relationship.swift#L262-L284

Sources/SwiftDocC/Semantics/Symbol/UnifiedSymbol+Extensions.swift
```diff
@@ -153,6 +153,18 @@ extension Dictionary where Key == UnifiedSymbolGraph.Selector {
     }
 }
 
+extension SymbolGraph.Relationship {
```
anferbui, line 156: Nit: Is this the right place for this extension? All the other functions in this file are for unified symbols, but this is a symbol graph extension. There's some other `SymbolGraph.Relationship` extensions which live in `Infrastructure/Symbol Graph/`.

anferbui (commented):
Left some comments, the most important one is about a potential remaining source of non-determinism, the rest are nits.

PR #1630, opened 2026-08-21
Update GitHub Actions workflows to v0.0.14

snprajwal:
One of the macOS GitHub Actions workflows currently fails due to the runner not having the Xcode version it requires. This was updated in https://github.com/swiftlang/github-workflows/pull/309, and shipped with the v0.0.14 release. Update GHA to use the newer version.


QuietMisdreavus (commented):
🚀

QuietMisdreavus (approved):

PR #1631, opened 2026-08-24
Deprecate almost all cases of the "bundle" terminology

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This renames a handful of types and API to avoid using the ambiguous and confusing "bundle" terminology. 

To understand what's confusing about the "bundle" terminology it helps to start with clarifying that a "bundle" is not the '.docc' directory of files; that is a "catalog". Instead, if we go by the use of the type previously named `DocumentationBundle` we can infer that a bundle is the collection of files that make up the documentation input. This _could_ be the contents of a catalog ('.docc' directory) but it could also be synthesized from a symbol graph directory passed via `--additional-symbol-graph-dir` or from the data properties of a `ConvertRequest`. In both cases there isn't a '.docc' catalog involved in creating a "bundle". 

In a few cases, the term "bundle" was also used to refer to the build documentation output, which we typically call a documentation "archive" and to refer to a unit of documentation from an external source (which isn't necessarily a documentation "archive"). As such, when encountering the term "bundle" in code, it could refer to at least these 4 things:
- A collection of discovered input files (which doesn't need to come from a catalog)
- An incorrect synonym for a documentation catalog
- An incorrect synonym for a documentation archive
- An external source of documentation which may or may not be a documentation archive (in which case the usage is an incorrect synonym like above).

This PR aims to address most of this confusion by renaming _almost_ all cases of the "bundle" terminology into either:
- `inputs` (referring to the collection of discovered input files)
- a `catalog` (referring to a '.docc' directory but not its contents)
- an `archive` (referring to the documentation output)
- a `source` (referring to an external source of documentation)

I hope that these more specific terms will be less confusing for contributors.

## Dependencies

None

## Testing

## Checklist

- [x] ~Added~ Updated tests


Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
```diff
@@ -340,14 +340,14 @@ extension Docc {
...
             ///
-            /// If the documentation catalogs's Info.plist file contains a bundle identifier, the documentation catalog ignores this fallback identifier.
+            /// If the documentation catalogs's Info.plist file contains a "bundle identifier", the documentation catalog ignores this fallback identifier.
             @Option(
                 name: [.customLong("fallback-bundle-identifier"), .customLong("bundle-identifier")], // Remove spelling without "fallback" prefix when other tools no longer use it. (rdar://72449411)
                 help: ArgumentHelp("A fallback bundle identifier if no value is provided in the documentation catalogs's Info.plist file.", discussion: """
-                If no bundle identifier is provided in the catalogs's Info.plist file or via the '--fallback-bundle-identifier' option, \
-                DocC will infer a bundle identifier from the display name.
+                If no identifier is provided in the catalogs's Info.plist file or via the '--fallback-bundle-identifier' option, \
+                DocC will infer a identifier from the display name.
```
snprajwal:
```suggestion
                DocC will infer an identifier from the display name.
```

Sources/SwiftDocC/Infrastructure/External Data/OutOfProcessReferenceResolver.swift
```diff
@@ -22,7 +22,7 @@ private import SymbolKit
 /// When creating an out-of-process resolver using ``init(processLocation:errorOutputHandler:)`` to communicate with another executable;
 /// DocC launches your link resolver executable and declares _its_ own ``Capabilities`` as a raw value passed via the `--capabilities` option.
 /// Your link resolver executable is expected to respond with a ``ResponseV2/identifierAndCapabilities(_:_:)`` message that declares:
-/// - The documentation bundle identifier that the executable can to resolve links for.
+/// - The documentation identifier that the executable can to resolve links for.
```
snprajwal:
```suggestion
/// - The documentation identifier that the executable can resolve links for.
```

Sources/SwiftDocC/Infrastructure/DocumentationContext+Inputs.swift
```diff
@@ -11,6 +11,7 @@
 public import Foundation
 import DocCCommon
 
+extension DocumentationContext {
```
snprajwal, line 14: Minor: since this adds a level of nesting, we should ideally indent everything inside it by one more level.

Sources/SwiftDocC/Infrastructure/ConvertOutputConsumer.swift
```diff
@@ -70,3 +70,18 @@ package protocol ExternalNodeConsumer {
...
+    }
+}
+
+// Default implementation so that conforming types don't need to implement deprecated API.
+public extension ConvertOutputConsumer {
+    @available(*, deprecated)
+    func consume(assetsInInputs inputs: DocumentationContext.Inputs) throws {
+        try consume(assetsInBundle: inputs)
+    }
+}
```
snprajwal, line 89: This looks like it would cause infinite recursion for existing consumers who only implement the old method. We've removed `consume(assetsInBundle:)` from the protocol requirements, so the extension method will now be statically dispatched. And AFAICT, new consumers who use both these default impls will also infinitely recurse.

d-ronnqvist: I replaced this call with a code comment that explains why it's safe to do nothing in this default implementation.

Sources/SwiftDocC/Infrastructure/DocumentationContext+Inputs.swift
d-ronnqvist: I wanted to avoid reindenting the full file but if you prefer to indent it I'm happy to do so.

snprajwal (approved):
Thanks David, this is great! 🚀

Sources/SwiftDocC/Infrastructure/DocumentationContext+Inputs.swift
snprajwal: I prefer the indentation being representative of how many levels deep the code is, so it's easy to know at a glance if something is top-level or attached to a type. Thanks for updating!

PR #1633, opened 2026-08-24
don't translate render sections if the input data is empty

QuietMisdreavus:
Bug/issue #, if applicable: Part of rdar://184352914

## Summary

In the RenderNodeTranslator, whenever we convert a symbol to a RenderNode, we call [a number of RenderSectionTranslators](https://github.com/swiftlang/swift-docc/blob/98a01609d5ef8d4d592c71824b39e5e01d32e2e8/Sources/SwiftDocC/Model/Rendering/RenderNodeTranslator.swift#L1532-L1556) with the symbol to return an optional CodableContentSection (read: a type-erased wrapper around an arbitrary render section). In some situations when given a symbol without the relevant data in question (e.g. a declaration or parameters), the translator in question will call [`translateSectionToVariantCollection`](https://github.com/swiftlang/swift-docc/blob/98a01609d5ef8d4d592c71824b39e5e01d32e2e8/Sources/SwiftDocC/Model/Rendering/RenderSectionTranslator/RenderSectionTranslator.swift#L25-L35), which, ultimately, returns a VariantCollection with no default value and an empty variants array.

In practice, this is usually harmless; these empty variant collections are dropped when encoding the RenderNode to JSON, so DocC-Render and any other tool that consumes Render JSON doesn't see malformed data. However, this does cause a problem in the JSON Patch emitter, specifically in the part that emits array indices for patch pointers. The pointer indices are generated based on the in-memory RenderNode, which includes the empty content sections. However, if a pointer needs to be generated to a section *after* that empty section, the index is no longer pointing to the correct place. In some internal testing, we found that this specific situation created a JSON Patch that could not be applied with DocC's own JSONPatchApplier.

Rather than attempt to write "adjusted indices" into the JSON Patch machinery, this PR instead opts to prevent creating these empty variant collections in the first place. That way, the in-memory representation fully matches the eventual on-disk representation, and the indices are correct without any adjustment.

## Dependencies

None

## Testing

This bug was found in a relatively uncommon scenario using symbol graphs representing plist symbols, which were given subtly different Swift and Objective-C variants (for irrelevant internal reasons). For an integration test, use the symbol data from the test in this PR to create two symbol graphs, and ensure that the previewed page does not duplicate the "Attributes" section with the list of possible types. The included unit test is intended as an end-to-end test both to ensure that the array index is in-bounds and that the generated patches apply correctly without errors using DocC's JSONPatchApplier.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -36,7 +36,25 @@ public struct DocumentationDataVariants<Variant> {
...
         values.isEmpty
     }
-    
+
+    /// Whether this variant collection has a default value or any variant set.
+    ///
+    /// If the variant type is a collection type, this property will also check that the default
+    /// value is not empty, even if it's set.
+    public var hasAnyValue: Bool {
+        guard self.isEmpty else { return true }
```
patshaughnessy: This always returns true ("there is at least one value") if the collection is not empty. Is that correct? Because the function checks if default value is an empty collection maybe it should also check each of the variants is an empty collection.

Also, maybe we should rename the function to describe the empty check... something like `hasAnyNonEmptyValue` ?

patshaughnessy (changes requested):
The `VariantOverridesEndToEndTests` test looks great, but could you write some unit tests for the new `hasAnyValue` function? That might help spell out exactly what behavior you're proposing.

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: I did think about doing a deep check in the variant values, but decided against it because the default value is the only place where not doing the check caused problems. Specifically, the declarations variants in the `Symbol` type are initialized with an empty dictionary as the default value, but no variant values. The same holds for a few other fields on the semantic `Symbol` type. I can definitely add the deep check, though; otherwise it feels like a rake ripe for stepping on. 😅

On the other hand, i'm less keen on changing the name of the property. The collection-probing behavior only happens *if the variant type is a collection*; there are plenty of examples throughout the codebase where that's not the case. Calling it "has any value" makes the most sense to me in both cases.

QuietMisdreavus: Added the variant collection probe in 070db7f, added the `hasAnyValue` tests in a0571e8.

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -36,7 +36,28 @@ public struct DocumentationDataVariants<Variant> {
...
+
+    /// Whether this variant collection has a default value or any variant set.
+    ///
+    /// If the variant type is a collection type, this property will also make sure that
+    /// any collection value, whether in the default value or any variant value, is non-empty.
+    /// In other words, if a `DocumentationDataVariants<[String]>` has a
+    /// default value of `[]` and no variants, this property will still return false.
+    public var hasAnyValue: Bool {
+        func hasValue(_ value: Variant?) -> Bool {
+            if let collectionValue = value as? (any Collection) {
```
patshaughnessy, line 54: Note that String will satisfy `as? (any Collection)` - probably that's fine but note hasAnyValue will return false for an empty string. Something to think about.

patshaughnessy (approved):
Thanks for the deeper check!

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -36,7 +36,28 @@ public struct DocumentationDataVariants<Variant> {
...
         values.isEmpty
     }
-    
+
+    /// Whether this variant collection has a default value or any variant set.
+    ///
+    /// If the variant type is a collection type, this property will also make sure that
+    /// any collection value, whether in the default value or any variant value, is non-empty.
+    /// In other words, if a `DocumentationDataVariants<[String]>` has a
+    /// default value of `[]` and no variants, this property will still return false.
```
snprajwal, line 51: (non-blocking) Given the fact that a set-but-empty string returns `false` as Pat pointed out below (which IMO is correct behaviour), this could instead be written as:


```suggestion
    /// Whether this variant collection has a default value or a non-empty variant.
    ///
    /// If the variant type is a collection type, this property will also make sure that
    /// the collection itself, whether in the default value or any variant value, is non-empty.
    /// In other words, if a `DocumentationDataVariants<[String]>` has a
    /// default value of `[]` and no variants, this property will still return false.
```

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -36,7 +36,28 @@ public struct DocumentationDataVariants<Variant> {
     public var isEmpty: Bool {
```
snprajwal, line 37: I think this name is a misrepresentation of what this property provides. With the introduction of `.hasAnyValue` below, it only makes it even more confusing to the reader at call-sites as to what the difference between the two is.

It would be good to rename this to `hasAnyVariants`, making it clearer that it only checks if the variants are empty. What do you think?

Tests/SwiftDocCTests/Semantics/DocumentationDataVariantsTests.swift
```diff
@@ -97,4 +100,18 @@ class DocumentationDataVariantsTests: XCTestCase {
             "MyLanguage"
         )
     }
+
+    func testHasAnyValue() {
```
snprajwal, line 108: Would it be possible to create a `struct DocumentationDataVariantsTests_new` below, and use Swift Testing for this? That's what we've been doing with other test files, and we're trying to not use XCTest for new tests.

Tests/SwiftDocCTests/Semantics/DocumentationDataVariantsTests.swift
```diff
@@ -97,4 +100,18 @@ class DocumentationDataVariantsTests: XCTestCase {
...
+        // hasAnyValue returns true only if a value is set in either the default or in a variant
```
snprajwal, line 109: (non-blocking) Ref the above comment on rewording:


```suggestion
        // hasAnyValue returns true only if a value is non-empty in either the default or in a variant
```

snprajwal (changes requested):

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: This is public API, so we can't rename it without a deprecation cycle 😓

snprajwal: Yeah :/ still worth doing IMO

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: I would hesitate to apply the "non-empty" language without qualification, since there are several instances of `DocumentationDataVariants` with variant types that are not collections where that doesn't apply. I can tweak the docs slightly if you'd like, but i don't want to take these suggestions as-is.

snprajwal: Go ahead, this was just a suggestion 😄

Tests/SwiftDocCTests/Semantics/DocumentationDataVariantsTests.swift
QuietMisdreavus: Rewording this comment is not necessary IMO since the collection/non-empty behavior is explained in the block of tests underneath it.

Tests/SwiftDocCTests/Semantics/DocumentationDataVariantsTests.swift
```diff
@@ -42,9 +43,12 @@ class DocumentationDataVariantsTests: XCTestCase {
...
     }
     
-    func testIsEmpty() throws {
-        XCTAssert(DocumentationDataVariants<String>().isEmpty)
-        XCTAssertFalse(DocumentationDataVariants<String>(values: [.swift : "Swift"]).isEmpty)
+    func testHasAnyVariants() throws {
+        XCTAssert(DocumentationDataVariants<String>().hasAnyVariants)
+        XCTAssertFalse(DocumentationDataVariants<String>(values: [.swift : "Swift"]).hasAnyVariants)
+
+        // isEmpty only checks the variant values, not the default value
```
snprajwal:
```suggestion

        // `hasAnyVariants` does not check the default value
```

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -33,10 +33,37 @@ public struct DocumentationDataVariants<Variant> {
     }
     
     /// Whether there are any variants for this piece of information about the documentation node
+    @available(*, deprecated, renamed: "hasAnyVariants", message: "Use 'hasAnyVariants' instead. This deprecated API will be removed after Swift 6.6 is released.")
```
snprajwal, line 36: We're still deprecating symbols for 6.5 (#1643), we should be able to do the same here:


```suggestion
    @available(*, deprecated, renamed: "hasAnyVariants", message: "Use 'hasAnyVariants' instead. This deprecated API will be removed after 6.5 is released.")
```

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -33,10 +33,37 @@ public struct DocumentationDataVariants<Variant> {
...
     public var isEmpty: Bool {
+        self.hasAnyVariants
+    }
+
+    /// Whether there are any variants for this piece of information about the documentation node
+    public var hasAnyVariants: Bool {
```
snprajwal, line 42: Renaming the existing condition makes this property inverted -- `hasAnyVariants` returns true when there are no variants. We should flip the condition below to `!values.isEmpty`, but if you prefer retaining the same semantics as asking `isEmpty`, we can also rename it to `hasNoVariants`.

snprajwal (changes requested):

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
d-ronnqvist: There's a good chance that by the time this gets merged we will be deprecating symbols for 6.6.

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: Whoops, i've fixed this now.

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: With Xcode 27 in RC today (and Swift 6.4 along with it) i wrote this to target 6.6 on the assumption that it was "close enough" to the release of 6.4 that it felt like too short of a deprecation window to remove it after 6.5. I can move it if you think it's necessary, but that was my rationale.

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
```diff
@@ -33,10 +33,37 @@ public struct DocumentationDataVariants<Variant> {
     }
     
     /// Whether there are any variants for this piece of information about the documentation node
+    @available(*, deprecated, renamed: "hasAnyVariants", message: "Use 'hasAnyVariants' instead. This deprecated API will be removed after Swift 6.6 is released.")
     public var isEmpty: Bool {
-        values.isEmpty
+        self.hasAnyVariants
```
snprajwal: This would also need to be flipped 😬


```suggestion
        !self.hasAnyVariants
```

snprajwal (approved):
LGTM pending one correctness fix below, thanks for iterating on this!

Sources/SwiftDocC/Semantics/Symbol/DocumentationDataVariants.swift
QuietMisdreavus: 🫠 Thanks for the catch, fixed now.

PR #1635, opened 2026-08-26
Create HTML nodes instead of XML nodes, except when parsing unknown raw HTML

d-ronnqvist:

## Summary

This continues the work of rdar://165755530 to phase out `Foundation.XMLNode` for HTML output by directly creating `HTMLNode` values where the code was previously creating `XMLElement` values. There are two exceptions to this where `XMLElements` are still created:
- When parsing unknown raw HTML text in the markup.
- When parsing the raw unknown custom header/footer files.

## Dependencies

None

## Testing

## Checklist

- ~[ ] Added tests~


Sources/DocCHTML/HTMLNode+Attributes.swift
```diff
@@ -397,7 +397,7 @@ extension HTMLNode {
         package static func srcLang(_ value: String) -> Attribute { .init(name: "srclang", value: value) } 
 
         /// A list of one or more resource URLs and their descriptors for the `<img>` or `<source>` element.
-        package static func srcSet(_ urlStrings: [String]) -> Attribute { .init(name: "srcset", value: urlStrings.joined(separator: " ")) }
+        package static func srcSet(_ urlStrings: [String]) -> Attribute { .init(name: "srcset", value: urlStrings.joined(separator: ", ")) }
```
d-ronnqvist, line 400: FYI: The existing tests caught that the original PR had gotten the separator wrong.

Sources/DocCHTML/MarkdownRenderer+Availability.swift
```diff
@@ -53,23 +46,22 @@ package extension MarkdownRenderer {
...
-            var attributes = [
-                "aria-label": "\(text), \(description)",
-                "title": description
+            var attributes: [HTMLNode.Attribute] = [
+                .init(name: "aria-label", value: "\(text), \(description)"),
+                .title(description),
             ]
             if $0.isBeta {
-                attributes["class"] = "beta"
+                attributes.insert(.class("beta"),       at: 1)
```
d-ronnqvist, line 54: FYI: The `HTMLFormatter` formats the attributes in the order they're added to ensure a stable output. By inserting the "class" attribute at this location, they attributes remain sorted in alphabetical order so the formatted HTML output remains stable and sorted without doing the work of sorting the attributes.

Sources/DocCHTML/MarkdownRenderer+Declaration.swift
```diff
@@ -72,13 +65,13 @@ package extension MarkdownRenderer {
                 fallthrough
             }
             // If the token refers to a symbol that the `linkProvider` is aware of, make that fragment a link to that symbol.
-            return .element(named: "a", children: [consume text], attributes: [
-                "href": path(to: reference),
-                "class": fragment.kind.htmlClassName
-            ])
+            return a(attributes: [
+                .class(fragment.kind.htmlClassName),
+                .href(path(to: reference)),
```
d-ronnqvist, line 70: FYI: Like mentioned [above](https://github.com/swiftlang/swift-docc/pull/1635/changes#r3861282877), this adds the attributes in alphabetical order so that the formatted HTML remains the same. Regardless of the order that they're defined, the formatted HTML is stable.

Sources/DocCHTML/MarkdownRenderer+Declaration.swift
```diff
@@ -239,7 +232,7 @@ private enum DeclarationFormatter {
     ///   - fragments: The SymbolKit declaration fragments to create a pretty printed HTML output for.
     ///   - renderer: The renderer that resolves USRs and determines the relative path from the current page to the linked page.
     /// - Returns: The list of XML nodes that represent the syntax-highlightable declaration fragments.
```
snprajwal:
```suggestion
    /// - Returns: The list of HTML nodes that represent the syntax-highlightable declaration fragments.
```

Sources/SwiftDocC/Model/Rendering/HTML/HTMLRenderer+FullPage.swift
```diff
@@ -10,11 +10,12 @@
 
 #if canImport(FoundationXML)
 // TODO: Consider other HTML rendering options as a future improvement (rdar://165755530)
-package import FoundationXML
-package import FoundationEssentials
+// XML nodes is still used to parse the raw HTML for the header and footer files.
```
snprajwal:
```suggestion
// XML nodes are still used to parse the raw HTML for the header and footer files.
```

Sources/DocCHTML/MarkdownRenderer.swift
```diff
@@ -10,7 +10,8 @@
 
 #if canImport(FoundationXML)
 // TODO: Consider other HTML rendering options as a future improvement (rdar://165755530)
-package import FoundationXML
+// XML nodes is still used to parse raw HTML in the markup.
```
snprajwal:
```suggestion
// XML nodes are still used to parse raw HTML in the markup.
```

Sources/DocCHTML/MarkdownRenderer.swift
```diff
@@ -888,3 +849,11 @@ func urlReadableFragment(_ fragment: some StringProtocol) -> String {
     
     return fragment
 }
+
+
+extension MarkdownRenderer {
+    /// Created a `<a>` element with a relative link to the given element.
```
snprajwal:
```suggestion
    /// Created an `<a>` element with a relative link to the given element.
```

Sources/DocCHTML/MarkdownRenderer.swift
```diff
@@ -888,3 +849,11 @@ func urlReadableFragment(_ fragment: some StringProtocol) -> String {
     
     return fragment
 }
+
```
snprajwal, line 852: Superfluous newline
```suggestion
```

snprajwal (approved):
I am concerned that removing the serialising for the test suites will cause CI to become flaky again, since we still construct XMLNode trees in some tests. I would prefer if we retained them until we've removed all XMLNode tree construction, but if you think it's safe to remove it now, I'm not against it.

LGTM otherwise, thank you!

Sources/DocCHTML/MarkdownRenderer.swift
d-ronnqvist:
```suggestion
    /// Creates an `<a>` element with a relative link to the given element.
```

PR #1636, opened 2026-08-27
Use package trait to gate SwiftNIO dependency

snprajwal:
## Summary

DocC depends on SwiftNIO to provide the preview server. SwiftNIO is currently not available on Windows, causing the build to fail when compiling this feature. For this reason, the preview server is gated behind a conditional import check in order to compile it only when NIOHTTP1 is available.

While this strategy allows DocC to build successfully on platforms without NIOHTTP1, it does not remove SwiftNIO from the dependency graph. This leads to the dependency being included for downstream consumers of DocC, even in cases where they depend on a different version of SwiftNIO, or do not depend on it at all. Swift 6.1 introduced package traits [1], which can be used to gate the dependency. Disabling the trait will remove SwiftNIO from the dependency graph.

Introduce the `PreviewServer` trait, which is enabled by default, and gate the preview server behind a build setting that combines this trait with the platforms NIOHTTP1 is available on.

[1]: https://docs.swift.org/swiftpm/documentation/packagemanagerdocs/packagetraits/

## Dependencies

N/A

## Testing

Without this patch, `swift package show-dependencies` includes SwiftNIO in the list.

With this patch:

- `swift package show-dependencies` includes it as the `PackageServer` trait is enabled by default
- `swift package show-dependencies --disable-default-traits` does not show SwiftNIO in the list

## Checklist

- ~[ ] Added tests~ NFC


heckj (approved):
Love it - nice setup that includes some platform-specific choices that flow way more easily into the conditional compilations, and let us flex to other things down the road if needed.

d-ronnqvist (approved):

PR #1639, opened 2026-08-31
Enable member-import-visibility feature

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This enables the member import visibility feature [SE-0444](https://github.com/swiftlang/swift-evolution/blob/main/proposals/0444-member-import-visibility.md) added in Swift 6.1. 

This means that files need to import the modules of the members that they access, resulting in the same name lookup behavior for both top-level declarations and members. This PR also adds all necessary imports for the issues raised by this feature. 

## Dependencies

None. This feature was added in Swift 6.1, which is already specified in as the tools version in the package manifest file.

## Testing

## Checklist

- ~[ ] Added tests~


Tests/SwiftDocCTests/Rendering/AutomaticSeeAlsoTests.swift
```diff
@@ -13,6 +13,8 @@ import XCTest
 @testable import SwiftDocC
 import DocCTestUtilities
 import DocCCommon
+import SymbolKit
+import SymbolKit
```
QuietMisdreavus: Looks like a duplicate import here.

snprajwal (approved):

PR #1641, opened 2026-08-31
Emit index.html files with the original `<noscript>` tag for Tutorial pages

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This fixes a bug where DocC wouldn't emit an index.html file for tutorial pages when passed the `--experimental-transform-for-static-hosting-with-content` flag.

With these changes, DocC emits a index.html file with the original `<noscript>` content, but with the tutorial page's specific title and description in the as metadata in the `<head>`.

## Dependencies

None.

## Testing

- Build documentation for a collection of tutorials and pass the `--experimental-transform-for-static-hosting-with-content` flag.
- Inspect the built documentation:
  - There should be an index.html file for each tutorial (and table of contents page, and tutorial article page)
  - Each index.html file should:
    - have the original `<noscript>` content from the template HTML file. 
    - be customized with the page's title and abstract/description in `<head>` tag.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Model/Rendering/HTML/HTMLContentConsumer.swift
```diff
@@ -12,8 +12,7 @@ package import DocCHTML
 
 /// A consumer for HTML content produced during documentation conversion.
 package protocol HTMLContentConsumer {
-    // One reason that this is its own protocol, rather than an extension of ConvertOutputConsumer, is so that we can avoid exposing `XMLNode` in any public API.
-    // That way, we are completely free to replace the entire internal HTML rendering implementation with something else in the future, without breaking API.
+    // This is its own protocol, rather than an extension of ConvertOutputConsumer, so that we can freely make changes without breaking public API.
```
snprajwal, line 15: Now that we've deprecated the public API, do we want to make this an extension in the future?

Sources/DocCCommandLine/Action/Actions/Convert/FileWritingHTMLContentConsumer.swift
```diff
@@ -74,14 +74,18 @@ struct FileWritingHTMLContentConsumer: HTMLContentConsumer {
...
             title: String,
             plainDescription: String?,
             prettyPrint: Bool
         ) -> String {
             var copy = original
             // Replace the content in reverse order so that the earlier ranges remain valid.
-            copy.replaceSubrange(contentReplacementRange, with: content.rendered(prettyPrinted: prettyPrint))
+            
+            if let content {
+                // Keep the "JavaScript required" information in the <noscript> tag unless we have a static HTML representation of the page's content.
```
snprajwal: This comment explains why we gate the copy on having non-nil content, but appears to explain the copy statement it's attached to instead. Might make more sense to have it on the if-block.


```suggestion
            // Keep the "JavaScript required" information in the <noscript> tag unless we have a static HTML representation of the page's content.
            if let content {
```

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
```diff
@@ -595,6 +595,150 @@ struct FileWritingHTMLContentConsumerTests {
         </html>
         """)
     }
+    
+    @Test
+    func preservesNoScriptTagForTutorials() async throws {
+        // Note: we need to change this test once we add support for emitting a static HTML representation of tutorials.
```
snprajwal: Nit: this might be better added as a to-do instead.


```suggestion
        // TODO: change this test once we add support for emitting a static HTML representation of tutorials.
```

snprajwal (approved):

Tests/DocCCommandLineTests/FileWritingHTMLContentConsumerTests.swift
d-ronnqvist: I personally don't think of this as a to-do because it can't be completed until the change in behavior that it references happens. 

My intention with this comment was more to communicate that when this test fails in the future because we add support for emitting static HTML for tutorials then the behavior that the test covers is no longer necessary and we should instead update the test. I'll update the comment to better reflect that intention. 
```suggestion
        // Once we add support for emitting a static HTML representation of tutorials, this test will fail.
        // That is expected, and the `<noscript>`-preserving behavior for tutorials is no longer necessary at that point. 
        // When this happens we should update/rewrite this test to reflect the expected static HTML tutorial content instead. 
```

Sources/SwiftDocC/Model/Rendering/HTML/HTMLContentConsumer.swift
d-ronnqvist: As long as the protocol is public I'd rather avoid modifying it. We can revisit the extension after the protocol is no longer public.

PR #1642, opened 2026-08-31
Promote static-hosting-with-content feature to non-experimental and enable by default

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This promotes the static-hosting-with-content feature to no longer be considered "experimental" and enables it by default.

Before these changes, the 3 spellings for the different static-hosting behaviors were as follows:
 * `--experimental-transform-for-static-hosting-with-content` Create a per page index.html file based on the provided template and modify each page to contain a minimal static representation of the page's content. This allows for the documentation to be minimally browsed without JavaScript enabled and allows tools that don't evaluate JavaScript to access the information on the requested page.
 * `--transform-for-static-hosting` **(default)** Make an identical copy of the same template index.html file for each without any per-page content. The documentation required JavaScript to display any pages and tools that don't evaluate JavaScript cannot access any information on the pages.
 * `--no-transform-for-static-hosting` Don't create or copy any per-page index.html files. The documentation doesn't support static hosting environments and requires custom routing rules on the web server.
 
With these changes, the spellings of these behaviors change accordingly (the spelling of the last behavior is unchanged)

| From | To |
|---|---|
| `--experimental-transform-for-static-hosting-with-content` | `--transform-for-static-hosting` (**default**) | 
|`--transform-for-static-hosting` (**default**) |  `--transform-for-static-hosting-without-content` 
| `--no-transform-for-static-hosting` | `--no-transform-for-static-hosting` (unchanged) |

## Dependencies

None.

## Testing

1. Build some symbol/article documentation and inspect the output.
   - There should be an index.html file for each page in the output.
     - Each index.html file should have a minimal static semantic representation of its content in the `<noscript>` tag, as well as title and description metadata in the page's `<head>` tag.
    
2. Build the same symbol/article documentation, pass `--experimental-transform-for-static-hosting-with-content`, and inspect the output.
   - There should be an index.html file for each page in the output. (same as above)
     - Each index.html file should have a minimal static semantic representation of its content in the `<noscript>` tag, as well as title and description metadata in the page's `<head>` tag. (same as above)
    
3. Build the same symbol/article documentation, pass `--transform-for-static-hosting-without-content`, and inspect the output.
   - There should be an index.html file for each page in the output. (same as above)
     - All index.html files should have identical content and should have the same "This page requires JavaScript" information in the `<noscript>` tag as the template file.

## Checklist

- [x] Updated ~Added~ tests

### DIFF
--- a/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
+++ b/Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
@@ -36,8 +36,17 @@ public struct ConvertAction: AsyncAction {
     private let documentationCoverageOptions: DocumentationCoverageOptions
     let diagnosticEngine: DiagnosticEngine
 
-    private let transformForStaticHosting: Bool
-    private let includeContentInEachHTMLFile: Bool
+    /// A configuration for how the action should transform the output for static hosting environments
+    package enum TransformForStaticHostingOptions: Equatable {
+        /// Transform the archive to support static hosting environments and minimal browsing without JavaScript enabled.
+        case withContent
+        /// Transform the archive to support static hosting environments but requiring JavaScript to render pages.
+        case withoutContent
+        /// Don't transform the archive to support static hosting environments.
+        case noTransform
+    }
+    
+    private let transformForStaticHostingOptions: TransformForStaticHostingOptions
     private let hostingBasePath: String?
     
     let sourceRepository: SourceRepository?
@@ -75,7 +84,7 @@ public struct ConvertAction: AsyncAction {
     ///   - experimentalEnableCustomTemplates: `true` if the convert action should enable support for custom "header.html" and "footer.html" template files, otherwise `false`.
     ///   - experimentalModifyCatalogWithGeneratedCuration: `true` if the convert action should write documentation extension files containing markdown representations of DocC's automatic curation into the `documentationBundleURL`, otherwise `false`.
     ///   - featureFlags: A collection of feature flags that the convert action uses to enable or disable certain optional behaviors.
-    ///   - transformForStaticHosting: `true` if the convert action should process the build documentation archive so that it supports a static hosting environment, otherwise `false`.
+    ///   - transformForStaticHostingOptions: Options to configure how the convert action transforms the built documentation archive so that it supports a static hosting environment.
     ///   - includeContentInEachHTMLFile: `true` if the convert action should process each static hosting HTML file so that it includes documentation content for environments without JavaScript enabled, otherwise `false`.
     ///   - allowArbitraryCatalogDirectories: `true` if the convert action should consider the root location as a documentation catalog if it doesn't discover another catalog, otherwise `false`.
     ///   - hostingBasePath: The base path where the built documentation archive will be hosted at.
@@ -107,8 +116,7 @@ public struct ConvertAction: AsyncAction {
         experimentalEnableCustomTemplates: Bool = false,
         experimentalModifyCatalogWithGeneratedCuration: Bool = false,
         featureFlags: FeatureFlags = .init(),
-        transformForStaticHosting: Bool = false,
-        includeContentInEachHTMLFile: Bool = false,
+        transformForStaticHostingOptions: TransformForStaticHostingOptions = .noTransform,
         allowArbitraryCatalogDirectories: Bool = false,
         hostingBasePath: String? = nil,
         sourceRepository: SourceRepository? = nil,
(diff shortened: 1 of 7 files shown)


Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
```diff
@@ -75,7 +84,7 @@ public struct ConvertAction: AsyncAction {
     ///   - experimentalEnableCustomTemplates: `true` if the convert action should enable support for custom "header.html" and "footer.html" template files, otherwise `false`.
     ///   - experimentalModifyCatalogWithGeneratedCuration: `true` if the convert action should write documentation extension files containing markdown representations of DocC's automatic curation into the `documentationBundleURL`, otherwise `false`.
     ///   - featureFlags: A collection of feature flags that the convert action uses to enable or disable certain optional behaviors.
-    ///   - transformForStaticHosting: `true` if the convert action should process the build documentation archive so that it supports a static hosting environment, otherwise `false`.
+    ///   - transformForStaticHostingOptions: Options to configure how the convert action transforms the built documentation archive so that it supports a static hosting environment.
     ///   - includeContentInEachHTMLFile: `true` if the convert action should process each static hosting HTML file so that it includes documentation content for environments without JavaScript enabled, otherwise `false`.
```
patshaughnessy: `includeContentInEachHTMLFile` is no longer a parameter; let's remove it from the doc comment also.

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
```diff
@@ -163,30 +163,29 @@ extension Docc {
...
 
-            /// A Boolean value that is true if the DocC archive produced by this conversion will support static hosting environments.
+            /// A Boolean value that is true if the DocC archive produced by this conversion will support static hosting environments and will support minimal browsing without JavaScript enabled.
             ///
             /// This value defaults to true but can be explicitly disabled with the `--no-transform-for-static-hosting` flag.
             @Flag(
                 inversion: .prefixedNo,
                 exclusivity: .exclusive,
-                help: "Produce a DocC archive that supports static hosting environments."
+                help: "Produce a DocC archive that include documentation content in each HTML file for static hosting environments."
```
patshaughnessy:
```suggestion
                help: "Produce a DocC archive that includes documentation content in each HTML file for static hosting environments."
```

Also: the bit about "includes documentation content" is actually controlled by the next flag: `--transform-for-static-hosting-without-content`, so maybe we should omit that from this help text entirely?

patshaughnessy (approved):
Looks good 👍 I just left a couple of documentation nits.

However, I find the option names a bit confusing. But I love the new enum and the case names that you use in the code. I know using boolean flags is simpler for a CLI, but maybe we could expose this choice to DocC users as a 3 case enum also? i.e. just use a single CLI option with three possible values? What do you think?

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
```diff
@@ -163,30 +163,29 @@ extension Docc {
...
+            ///
+            /// This value is ignored if ``transformForStaticHosting`` is `false`.
+            @Flag(
+                name: [.customLong("transform-for-static-hosting-without-content")],
+                help: "Omit documentation content from each HTML file for static hosting environments."
+            )
+            var omitContentFromStaticHostingOutput = false
             
-            mutating func validate() throws {
-                if experimentalTransformForStaticHostingWithContent, !transformForStaticHosting {
```
snprajwal, line 181: I don't think we should remove validation, we want to warn if someone uses a combination of flags that don't make sense. IIUC, with these changes, passing both `--no-transform-for-static-hosting` and `--transform-for-static-hosting-without-content` will silently resolve to `.noTransform`.

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
d-ronnqvist: With these changes content is included by default which means that if a developer passes `--transform-for-static-hosting` they get an archive that includes documentation content I each HTML file. Updated this help text to reflect that.

Sources/DocCCommandLine/ArgumentParsing/Subcommands/Convert.swift
d-ronnqvist: Added back in https://github.com/swiftlang/swift-docc/pull/1642/commits/3a89ff645c131c664e4f0603a24db203c1ebe899

Sources/DocCCommandLine/Action/Actions/Convert/ConvertAction.swift
d-ronnqvist: Removed in https://github.com/swiftlang/swift-docc/pull/1642/commits/5d845829d52d1e282e54a0f0f703f468fd429cb7

PR #1649, opened 2026-09-01
Omit Swift attributes from the suggested type-based disambiguation

d-ronnqvist:
Bug/issue #, if applicable: rdar://185325880

## Summary

This fixes a bug in DocC's suggested disambiguation when a symbol's name is ambiguous and one of the overloads has a result builder parameter. 

Before these changes, DocC would include portions of the result builder's type before that parameter's closure type. If applied, a link with this suggested disambiguation may not always parse correctly—depending on which portions of the result builder's type were included—resulting in what was intended as only a parameter type being identifies as part parameter type and part return value. This ultimately resulted in the link with the suggested disambiguation failing to resolve. 

These changes fix the root of this issue—the incorrect suggestion—by ignoring the Swift parameter's fragment's that appear before the parameter's name.

## Dependencies

None.

## Testing

- Create a Swift project with a generic container type that defines an inner result builder and two overloads of a function or method where one overload has a parameter using that result builder. For example:

   ```swift
   /// An ambiguous link to ``doSomething(with:)``
   public struct SomeValue<Something> {
       @resultBuilder
       public enum Builder {
           public static func buildBlock() -> SomeValue<Something> {
               fatalError()
           }
       }
   }

   public func doSomething(with value: SomeValue<Int>) {
       fatalError()
   }

   public func doSomething(@SomeValue<Int>.Builder with body: () -> SomeValue<Int>) {
       fatalError()
   }
   ```

- Build documentation for this code and inspect the warning about the ambiguous symbol link 
   - The suggested disambiguation for the overload with the result builder parameter should only suggest the closure type (`()->SomeValue<Int>` in the example above)
   
- Apply the suggested disambiguation for the overload with the result builder parameter and rebuild the documentation
   - The updated link with the suggested disambiguation should resolve to that overload of the symbol.

## Checklist

- [x] Added tests


snprajwal (approved):
Nice one, thanks!

PR #1651, opened 2026-09-03
Fully build topic graph before automatic curation

snprajwal:
Bug/issue #, if applicable: rdar://184277709

## Summary

When the documentation context crawls symbols and builds the topic graph, it checks if a descendant is transitively curated under its canonical container. If so, then it skips automatic curation of the descendant to prevent duplicate curation. However, this operation was being performed as the topic graph edges were being constructed. Furthermore, the check for the nearest container symbol used the first symbol it encountered at the shortest distance, which may not be correct in cases where there are multiple container symbols at the same distance from the container being processed. This led to automatic curation lists changing across builds while the input remained consistent.

Split the crawl into two phases, where the first phase fully populates the topic graph, and the second phase then checks whether descendants should be automatically curated under their canonical containers. Update the nearest symbol check to compare against all nearest symbol containers instead of the just the first one.

## Dependencies

N/A

## Testing

1. Create an archive with two structs, and a method inside one of them.
2. Add an API collection that curates the method, and curate this collection under both structs.
3. Build documentation multiple times. Prior to this patch, the method is sometimes auto-curated under the struct that contains it. With this patch, it is consistently excluded.

You can also refer [this example](https://github.com/snprajwal/swift-docc/blob/dda4387b21f76c75003c4e0c59da54092c8a6d7c/Sources/SwiftDocC/Infrastructure/DocumentationContext.swift#L2528-L2540) in the comments to test this.

## Checklist

- [x] Added tests


Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
```diff
@@ -18,6 +18,25 @@ extension DirectedGraph {
...
         IteratorSequence(GraphNodeIterator(from: startingPoint, traversal: .depthFirst, in: self))
     }
+
+    /// Returns every node that matches the given predicate at the shortest distance from a given element.
+    ///
+    /// - Parameters:
+    ///   - startingPoint: The element to measure distances from.
+    ///   - predicate: A closure that indicates whether a node is a match.
+    /// - Returns: The set of matching nodes at the shortest distance from `startingPoint`.
+    func nearestNodes(from startingPoint: Node, matching predicate: (Node) -> Bool) -> Set<Node> {
```
d-ronnqvist: Because this isn't really a general graph traversal algorithm I think I would prefer that it'd be a private extension in the same file as the calling code that needs this specific behavior.

Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
```diff
@@ -18,6 +18,25 @@ extension DirectedGraph {
...
+        var nodes = [startingPoint]
+        var seen: Set<Node> = [startingPoint]
+        while !nodes.isEmpty {
+            let matches = nodes.filter(predicate)
+            if !matches.isEmpty {
+                return Set(matches)
+            }
+            nodes = nodes.flatMap { neighbors(of: $0) }.filter { seen.insert($0).inserted }
+        }
+        return []
```
d-ronnqvist: I have a vague hunch that this could be a bit more allocation-heavy than the logic that it replaces and if this may need to run for every page then that could add up, causing both a slowdown and increased memory usage. Can you benchmarked these changes on some large project that is predominantly automatic curations and some large project that is predominant authored curation?

snprajwal: Large project with mostly manual curation:

```
┌──────────────────────────────────────────┬─────────────────┬──────────────────────┬──────────────────────┐
│ Metric                                   │ Change          │ 871966648            │ current              │
├──────────────────────────────────────────┼─────────────────┼──────────────────────┼──────────────────────┤
│ Duration for 'convert-total-time'        │ no change¹,²    │ 0.653 secs           │ 0.658 secs           │
│ Duration for 'documentation-processing'  │ no change³      │ 0.277 secs           │ 0.271 secs           │
│ Duration for 'finalize-navigation-index' │ no change⁴,⁵    │ 0.02 secs            │ 0.021 secs           │
│ Peak memory footprint                    │ no change⁶,⁷    │ 123.6 MB             │ 123.2 MB             │
│ Data subdirectory size                   │ no change       │ 29.3 MB              │ 29.3 MB              │
│ Index subdirectory size                  │ no change       │ 608 KB               │ 608 KB               │
│ Total DocC archive size                  │ no change       │ 131.2 MB             │ 131.2 MB             │
│ Topic Anchor Checksum                    │ no change       │ a62b963da27041bb45fe │ a62b963da27041bb45fe │
│ Topic Graph Checksum                     │ no change       │ 5da2e9a762adc7a8e0b2 │ 5da2e9a762adc7a8e0b2 │
└──────────────────────────────────────────┴─────────────────┴──────────────────────┴──────────────────────┘
```

Large project with mostly automatic curation:

```
┌──────────────────────────────────────────┬─────────────────┬──────────────────────┬──────────────────────┐
│ Metric                                   │ Change          │ 871966648            │ current              │
├──────────────────────────────────────────┼─────────────────┼──────────────────────┼──────────────────────┤
│ Duration for 'convert-total-time'        │ no change¹,²    │ 0.8 secs             │ 0.822 secs           │
│ Duration for 'documentation-processing'  │ no change³,⁴    │ 0.283 secs           │ 0.281 secs           │
│ Duration for 'finalize-navigation-index' │ no change⁵,⁶    │ 0.037 secs           │ 0.038 secs           │
│ Peak memory footprint                    │ no change⁷      │ 276.4 MB             │ 275.5 MB             │
│ Data subdirectory size                   │ no change       │ 41.9 MB              │ 41.9 MB              │
│ Index subdirectory size                  │ no change       │ 848 KB               │ 848 KB               │
│ Total DocC archive size                  │ no change       │ 383.9 MB             │ 383.9 MB             │
│ Topic Anchor Checksum                    │ no change       │ 3e1066fc481a4b717a94 │ 3e1066fc481a4b717a94 │
│ Topic Graph Checksum                     │ no change       │ a10543752aa21fd9a0fd │ a10543752aa21fd9a0fd │
└──────────────────────────────────────────┴─────────────────┴──────────────────────┴──────────────────────┘
```

From what I understand, we don't allocate any more than we used to when doing BFS with `.first(where:)`, just that it now happens after curation rather than during.

Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
snprajwal: I would say this is just a variant of BFS, and similar to the extensions we already have in `DirectedGraph+Paths.swift`. Having this here also means that we're able to unit test it independent of the curation logic. Is there a strong reason to not consider it appropriate as a traversal function?

Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
d-ronnqvist: I see that most of those measurements have a second footnote on them which is an indication from the script that the data was too either too noisy or showed possible signs of linear bias. Because of this, the raw output is inconclusive and needs to be checked by a person.

d-ronnqvist: I think I missed that this was processing the _reverse_ edges which meant that in practice the number of neighbors is almost always one. I'm still slightly concerned that a future caller could call this on the hierarchies primary direction where eagerly fetching entire depths could easily result in unintentionally fetching and processing all members of all top-level types.

Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
d-ronnqvist: I think that if the curation code didn't need this specific logic, we wouldn't keep this code. Rewriting the construction of the documentation hierarchy is on my list and I'm not sure that this code would remain useful then. I would find it more useful to have more coverage of the impact this logic has on the curation and the construction of the documentation hierarchy because those would be valuable regression tests for any changes to curation. Ultimately, it doesn't matter much if this isolated logic/algorithm is correct if its only use in curation doesn't produce the desired outcome.

Sources/SwiftDocC/Utility/Graphs/DirectedGraph+Traversal.swift
snprajwal: I did another benchmarking run, this time with no other processes running on my machine.

Large project with mostly manual curation:

```
│ Metric                                   │ Change          │ 1e2527d0a            │ current              │
│ Duration for 'convert-total-time'        │ no change¹      │ 0.631 secs           │ 0.633 secs           │
│ Duration for 'documentation-processing'  │ no change²      │ 0.268 secs           │ 0.264 secs           │
│ Duration for 'finalize-navigation-index' │ +2.951%³        │ 0.02 secs            │ 0.02 secs            │
│ Peak memory footprint                    │ no change⁴      │ 124 MB               │ 123.7 MB             │
│ Data subdirectory size                   │ no change       │ 29.3 MB              │ 29.3 MB              │
│ Index subdirectory size                  │ no change       │ 608 KB               │ 608 KB               │
│ Total DocC archive size                  │ no change       │ 131.2 MB             │ 131.2 MB             │
│ Topic Anchor Checksum                    │ no change       │ a62b963da27041bb45fe │ a62b963da27041bb45fe │
│ Topic Graph Checksum                     │ no change       │ 5da2e9a762adc7a8e0b2 │ 5da2e9a762adc7a8e0b2 │
```

Large project with mostly automatic curation:

```
│ Metric                                   │ Change          │ 1e2527d0a            │ current              │
│ Duration for 'convert-total-time'        │ -1.111%¹        │ 0.777 secs           │ 0.769 secs           │
│ Duration for 'documentation-processing'  │ -1.259%²,³      │ 0.267 secs           │ 0.264 secs           │
│ Duration for 'finalize-navigation-index' │ no change⁴      │ 0.035 secs           │ 0.035 secs           │
│ Peak memory footprint                    │ no change⁵      │ 275.4 MB             │ 276.1 MB             │
│ Data subdirectory size                   │ no change       │ 41.9 MB              │ 41.9 MB              │
│ Index subdirectory size                  │ no change       │ 848 KB               │ 848 KB               │
│ Total DocC archive size                  │ no change       │ 383.9 MB             │ 383.9 MB             │
│ Topic Anchor Checksum                    │ no change       │ 3e1066fc481a4b717a94 │ 3e1066fc481a4b717a94 │
│ Topic Graph Checksum                     │ no change       │ a10543752aa21fd9a0fd │ a10543752aa21fd9a0fd │
```

In both cases, the footnote for memory footprint mentions that there is no statistically significant difference. I did not see any footnotes about bias this time for this metric.

snprajwal: > I'm still slightly concerned that a future caller could call this on the hierarchies primary direction where eagerly fetching entire depths could easily result in unintentionally fetching and processing all members of all top-level types.

I've moved it into a private extension in the file where we call this, with a note explaining this risk.

d-ronnqvist (approved):

PR #1652, opened 2026-09-03
Disable preview server trait by default on unsupported platforms

rpaulo:
## Summary

- Introduce `previewServerEnabledByDefault` based on host platform
- Only enable the preview server trait on macOS, iOS, Linux, and Android
- Prevents SwiftNIO dependency issues on platforms like FreeBSD

Note that this can be reverted once Swift NIO supports FreeBSD: https://github.com/apple/swift-nio/pull/3398

## Dependencies

None

## Testing

Built Swift on FreeBSD.

## Checklist

- [ ] Added tests

They are N/A.


Package.swift
```diff
@@ -51,6 +51,12 @@ let previewServerSettings: [SwiftSetting] = [
     .define("PREVIEW_SERVER", .when(platforms: NIOPlatforms, traits: [previewServerTrait]))
 ]
 
+#if os(macOS) || os(iOS) || os(Linux) || os(Android)
```
snprajwal, line 59: This change means that now, if someone cross-compiles on a host platform that does not support SwiftNIO but for a target platform that _does_ support it, the preview server would be disabled by default. It would be good to document this:


```suggestion
// Disable the preview server trait by default on hosts where SwiftNIO does not build.
// This allows dependency resolution to exclude SwiftNIO.
//
// Note: This will also disable the trait when cross-compiling for targets where SwiftNIO *does* build.
// In such cases, manually enable the trait with `--traits PreviewServer`.
#if os(macOS) || os(iOS) || os(Linux) || os(Android)
```

snprajwal (approved):
Nice one, the trait itself gates for the _target_ platform, but this change lets us disable the trait on the _host_.

Note: I don't think we should revert this after SwiftNIO gets FreeBSD support, since this gate is complementary to the existing one. We should instead add FreeBSD to the list of supported platforms.

PR #1653, opened 2026-09-07
Emit slightly more compact HTML for symbol names in topic sections

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This outputs slightly more compact HTML for symbol names in topic sections. 

Before this, a symbol's name in its topic section item used `span` tags with classes for both kinds of fragments in the topic section item's `code` tag. However, because there's only two kinds of tags, only one of them needs to be wrapped in a `span` tag to distinguish it from the other:

```diff
  <code>
-   <span class="decorator">func </span>
+   func 
    <span class="identifier">distance</span>
-   <span class="decorator">(</span>
+   (
    <span class="identifier">from</span>
-   <span class="decorator">
-     :<wbr> Self.Index, 
-   </span>
+   :<wbr> Self.Index, 
    <span class="identifier">to</span>
-   <span class="decorator">
-     :<wbr> Self.Index)<wbr> -><wbr> Int
-   </span>
+   :<wbr> Self.Index)<wbr> -><wbr> Int
  </code>
```

Because topics sections (and see also sections) are quite commonly occurring, this makes the static HTML output about 0.5% smaller without rendering any different. 

## Dependencies

None.

## Testing

- Build documentation and pass `--output-format experimental-html-for-development`
- Inspect some rendered page that includes some symbols in topic sections (either automatic or authored).
  - The symbol names in the topic section should visually display the same as before (fragments colored in either gray or light blue depending on the specific symbol)

## Checklist

- [x] ~Added~ Updated tests


patshaughnessy (approved):
Looks good to me!

PR #1659, opened 2026-09-14
Reduce some allocations across the build

snprajwal:
## Summary

Reduce allocations with the below changes:

- Reuse `DocumentationContentRenderer` instead of creating a new instance
- Reuse available variant traits for a page
- Collapse allocations in `DirectedGraph.nearestNodes` (and make it more readable at a glance)
- Make a few of the iterator chains lazy
- Extract the exclusion set in the navigator into a static set

This does not reduce peak memory usage, but helps with transient allocations.

It shows a modest ~1.5% improvement in build time for large frameworks with overloads.
It also shows a ~5% improvement in the navigator build phase for large frameworks.

## Dependencies

N/A

## Testing

N/A (NFC)

## Checklist

~- [ ] Added tests~


Sources/SwiftDocC/Indexing/RenderNode+Indexable.swift
```diff
@@ -17,7 +17,7 @@ extension RenderNode {
     }
 
     var rawIndexableTextContent: String {
-        return contentSections
+        return contentSections.lazy
```
d-ronnqvist: I don't think `lazy` helps here because the code iterates through all the sections either way.

If we want to make this computation faster I _think_ it would be better to combine the `filter` and `map` logic into a single `compactMap` call.

Sources/SwiftDocC/Indexing/RenderBlockContent+TextIndexing.swift
```diff
@@ -71,7 +71,7 @@ extension RenderBlockContent: TextIndexing {
         case .links(let links):
             // Matches the behavior in `RenderInlineContent+TextIndexing` for a
             // `RenderInlineContent.reference`
-            return links.items
+            return links.items.lazy
```
d-ronnqvist: I don't think `lazy` helps here because the code iterates through all the items either way.

If we wanted to make this compilation faster I _think_ it'd be better to move the `\.title` access into the `compactMap` call instead.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
```diff
@@ -348,7 +348,7 @@ struct PathHierarchy {
                 acc[uniqueID] = Node(symbol: symbol, name: symbol.pathComponents.last!)
             }
 
-            for relationship in unifiedGraph.relationshipsByLanguage.flatMap(\.value) where relationship.kind == .overloadOf {
+            for relationship in unifiedGraph.relationshipsByLanguage.lazy.flatMap(\.value) where relationship.kind == .overloadOf {
```
d-ronnqvist, line 351: I don't think `lazy` helps here because the code iterates through all the relationships either way.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
```diff
@@ -1147,8 +1147,8 @@ public class DocumentationContext {
                     }
 
                     let overloadGroups: [String: Set<String>] =
-                    unifiedSymbolGraph.relationshipsByLanguage.values.flatMap({
-                        $0.filter { $0.kind == .overloadOf }
+                    unifiedSymbolGraph.relationshipsByLanguage.values.lazy.flatMap({
+                        $0.lazy.filter { $0.kind == .overloadOf }
```
d-ronnqvist, line 1151: I don't think `lazy` helps on either of these lines because the code iterates through all the relationships either way.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
snprajwal: We do iterate through everything, but if my understanding of `.lazy` is correct, we skip an intermediate allocation in the flat map here.

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
snprajwal: We do iterate through everything, but if my understanding of `.lazy` is correct, we skip an intermediate allocation in the flat map and filter here, and directly allocate in the `.reduce` below.

d-ronnqvist: In my experience, it can be difficult to judge the impact of `.lazy` without measuring. Sometimes the added overhead and additional abstraction can even make execution slower. 

Because this computation was so well isolated I crudely microbenchmarked it and found that adding `.lazy` does improve this computation by ~20–25%. 

That said, I also tried rewriting the expression to move the filtering loop into the reduce call (see below) and that improved this computation by 50%.

```swift
let overloadGroups: [String: Set<String>] = graph.relationshipsByLanguage.values.reduce(into: [:]) { acc, relationships in
    for relationship in relationships where relationship.kind == . overloadOf {
        acc[relationship.target, default: []].insert(relationship.source)
    }
}
```

That matches my past experiences; if `lazy` helps, it usually helps even more to switch to a regular for-loop instead.

Sources/SwiftDocC/Infrastructure/Link Resolution/PathHierarchy.swift
d-ronnqvist: Similar to [this comment](https://github.com/swiftlang/swift-docc/pull/1659/changes/BASE..a01613e4451f44525e7e9435eedec88c932dcd11#r4006873241), some crude microbenchmarking indicate that adding `.lazy` helps _some_ but that rewriting this as two loops indicate an bigger improvement. 

```swift
for relationships in unifiedGraph.relationshipsByLanguage.values {
    for relationship in relationships where relationship.kind == .overloadOf {
...
```

Further, that entire loop won't do any work if `overloadGroupNodes` is empty so checking that and doing an early continue results in an improvement that's so large that it's measurable on the level of the entire path hierarchy initialization:

```swift
guard !overloadGroupNodes.isEmpty else {
    continue
}
```

snprajwal: Oo nice find, updated!

Sources/SwiftDocC/Infrastructure/DocumentationContext.swift
snprajwal: Noted, updated!

d-ronnqvist (approved):

d-ronnqvist:
@swift-ci please test

PR #1662, opened 2026-09-14
Faster decoding of link dependencies

d-ronnqvist:
Bug/issue #, if applicable: 

## Summary

This speeds up decoding of link dependencies by conforming `LinkDestinationSummary` to `FastJSONDecodable` (https://github.com/swiftlang/swift-docc/commit/511e30743d63548942da9144a088f8b8889108d7) and adopting it when decoding dependencies (https://github.com/swiftlang/swift-docc/commit/870cf51fb5fad598cfc359b86df3f5cf799b2886). 

It then updates the out-of-process resolver used to resolve links to other documentation systems so that all `LinkDestinationSummary` code paths use the same code (https://github.com/swiftlang/swift-docc/commit/2523662feb0df42f23c1c63f86b2a63486bacb63).

## Dependencies

None.

## Testing

This should be a non-functional change but if there are bugs they would be either with:
- passing a `--dependency` to `docc convert`
- configuring DocC to call a custom script to resolve links to another documentation system (see for example, `bin/test-data-external-resolver`)

In addition to the test coverage (verifying that both decoding implementations produce the same values) I've tried to manually verify both approaches. 

## Checklist

- [x] ~Added~ Updated tests


franklinsch:
Thanks David, are you able to share a performance analysis of these changes so that we can assess the code complexity vs. speedup tradeoffs?

d-ronnqvist:
> Thanks David, are you able to share a performance analysis of these changes so that we can assess the code complexity vs. speedup tradeoffs?

The speedup depends greatly on the size and number of dependencies. For example, the Swift standard library as a dependency takes around 900ms to decode today but with these changes it takes closer to 160ms. Across a range of different dependencies that I tested with—both single language and multi language ranging from ~1k pages to over 10k pages—the task of decoding dependencies is about 4-6 _times_ faster while also using about 40% less memory.

This won't have any impact on a build that doesn't specify any dependencies or that doesn't resolve any content from other documentation systems (because that wouldn't exercise the code that decodes dependencies). On the flip side; for a small project that have a large dependency (like Swift standard library or Foundation) the full build takes in the ballpark of 3-5 seconds, in which case saving ¾ of a second is rather significant.

d-ronnqvist:
As for the "complexity" of this decoding code; there's a lot of boilerplate in a `FastJSONDecodable` conformance but for the most part the logic is quite straight forward. In the few cases where the decoding is more complicated (`any RenderReference`, `SourceLanguage`, and `DocumentationNode.Kind`) most of that complexity stems from the custom logic/behavior that we already have for decoding that type of information and that we need to adapt to a pattern where fields/keys are decoded in the order that they are encountered in the data.

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+                let decoded = try self.decode(String.self)
+                guard let encodedData = Data(base64Encoded: decoded, options: []) else {
+                    throw self.makeDataCorruptedError(message: "Decoded string is not Base-64 encoded")
+                }
+                iconBase64 = encodedData
+            }
+            else if self.matchKey("\"syntax\"", byteOffset: -1) {
+                syntax = try self.decode(String.self)
+            }
+            else if self.matchKey("title") {
```
snprajwal, line 779: If I understand this correctly: https://github.com/swiftlang/swift-docc/blob/7dff578e24a81754a4136276b24cc88f64507d1d/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift#L198-L199

then this will match `titleInlineContent` since the prefix doesn't capture the `"` for this key.

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+            else if decoder.matchKey("identifier") {
+                identifier = try decoder.decode(RenderReferenceIdentifier.self)
+            }
+            else if decoder.matchKey("metadata") {
+                metadata = try decoder.decode(RenderContentMetadata.self)
+            }
+            else if decoder.matchKey("isActive") {
+                isActive = try decoder.decode(Bool.self)
+            }
+            else if decoder.matchKey("overridingTitle") {
```
snprajwal, line 395: If I understand this correctly: https://github.com/swiftlang/swift-docc/blob/7dff578e24a81754a4136276b24cc88f64507d1d/Sources/DocCCommon/JSONDecoding/FastSymbolGraphJSONDecoder.swift#L198-L199

then this will match `overridingtitleInlineContent` since the prefix doesn't capture the `"` for this key.

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+extension RenderReferenceIdentifier: FastJSONDecodable {
+    package init(using decoder: inout FastSymbolGraphJSONDecoder) throws(DecodingError) {
+        let rawValue = try decoder.decode(String.self)
+        self = .init(consume rawValue)
+    }
+}
+
+extension RenderContentMetadata: FastJSONDecodable {
+    package init(using decoder: inout FastSymbolGraphJSONDecoder) throws(DecodingError) {
+        while try decoder.advanceToNextKey() {
```
snprajwal, line 496: Do we need to descend into this object before advancing to the next key?

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+}
+
+extension ConformanceSection: FastJSONDecodable {
+    package init(using decoder: inout FastSymbolGraphJSONDecoder) throws(DecodingError) {
+        typealias _MaybeDecodedValue = Optional
+
+        // 1 required property
+        var constraints: _MaybeDecodedValue<[RenderInlineContent]> = nil
+
+        while try decoder.advanceToNextKey() {
```
snprajwal, line 1027: Do we need to descend into this object before advancing to the next key?

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+    package init(using decoder: inout FastSymbolGraphJSONDecoder) throws(DecodingError) {
+        typealias _MaybeDecodedValue = Optional
+
+        // 1 required property
+        var line: _MaybeDecodedValue<Int> = nil
+        // 2 properties with default values
+        var start:  Int? = nil
+        var length: Int? = nil
+
+        while try decoder.advanceToNextKey() {
```
snprajwal, line 1056: Do we need to descend into this object before advancing to the next key?

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+
+extension RenderNode.Tag: FastJSONDecodable {
+    package init(using decoder: inout FastSymbolGraphJSONDecoder) throws(DecodingError) {
+        typealias _MaybeDecodedValue = Optional
+
+        // 2 required property
+        var type: _MaybeDecodedValue<String> = nil
+        var text: _MaybeDecodedValue<String> = nil
+
+        while try decoder.advanceToNextKey() {
```
snprajwal, line 1122: Do we need to descend into this object before advancing to the next key?

Sources/SwiftDocC/LinkTargets/LinkDestinationSummary+FastJSONDecodable.swift
```diff
@@ -0,0 +1,1219 @@
...
+        typealias _MaybeDecodedValue = Optional
+
+        // 1 required property
+        var constraints: _MaybeDecodedValue<[RenderInlineContent]> = nil
+
+        while try decoder.advanceToNextKey() {
+            if decoder.matchKey("constraints") {
+                constraints = try decoder.decode([RenderInlineContent].self)
+            }
+            // Do nothing for all unknown keys
```
snprajwal, line 1031: I see that `ConformanceSection` has two more keys apart from `constraints`: https://github.com/swiftlang/swift-docc/blob/7dff578e24a81754a4136276b24cc88f64507d1d/Sources/SwiftDocC/Model/Rendering/Symbol/ConformanceSection.swift#L33-L40

I think we only use the default values for those keys anyway, so ignoring them is fine, but maybe we want to decode them here for completeness' sake?

snprajwal:
We might also want to add some tests for the types where we need to descend into them but aren't

PR #1665, opened 2026-09-15
Decode RenderNode variant overrides as variant collection values

d-ronnqvist:
Bug/issue #, if applicable: rdar://128050800

## Summary

## Dependencies

## Testing

applicable._

Steps:

## Checklist

- [x] Added tests


Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -71,7 +74,296 @@ extension RenderNode: Codable {
...
+                            throw makeUnsupportedJSONPatchError()
+                        }
+                        // Instead of replacing the language we add it to the reference
+                        let languageID = try patchContainer.decode(String.self, forKey: .value)
+                        identifier = identifier.addingSourceLanguages([SourceLanguage(id: languageID)])
+                        
+                    // Raise a decoding error for any of the non-varying RenderNode properties
+                    case .schemaVersion,
+                         .kind,
+                         .hierarchy,
```
d-ronnqvist: This property _was_ non-varying when I made this commit but #1081 made it varying. It's patches would need to be handled like the other varying properties above.

d-ronnqvist: Resolved in https://github.com/swiftlang/swift-docc/pull/1665/commits/4572698f8e2de5d5b3b33b90467411f91fc939ef

Sources/SwiftDocC/Converter/RenderNode+Coding.swift
```diff
@@ -223,7 +223,7 @@ public extension RenderNode {
             try TopicRenderReferenceEncoder.addRenderReferences(
                 to: &renderNodeData,
                 references: references,
-                encodeAccumulatedVariantOverrides: variantOverrides == nil,
+                encodeAccumulatedVariantOverrides: false,
```
patshaughnessy: Should this be true instead? If we pass a false here, the encoded render node will never have a value for `variantOverrides`.

Tests/SwiftDocCTests/Rendering/Variants/VariantOverridesEndToEndTests.swift
```diff
@@ -17,35 +17,34 @@ import Testing
 import DocCCommon
 
 struct VariantOverridesEndToEndTests {
-    @Test("Renders correct JSON Patches when building variant overrides for plist symbols")
-    func testPlistVariantOverridesFromSymbolGraphs() async throws {
+    func decodedRendedNodeHasSamePropertyListDetailsVariantsAsOriginal() async throws {
```
patshaughnessy, line 20: We're missing `@Test` here, preventing the test from running.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -71,7 +75,307 @@ extension RenderNode: Codable {
...
+                                throw DecodingError.dataCorruptedError(forKey: .pointer, in: patchContainer, debugDescription: "JSON patch index \(index) is out of bounds for collection with \(primaryContentSectionsVariants.count) elements at \(pointer).")
+                            }
+                            switch operation {
+                            case .replace:
+                                let value = try patchContainer.decode(CodableContentSection?.self, forKey: .value)
+                                primaryContentSectionsVariants[index].addPatch(.replace(value: value), toVariantWithTraits: traits)
+                            case .add:
+                                let value = try patchContainer.decode(CodableContentSection?.self, forKey: .value)
+                                if primaryContentSectionsVariants.indices.contains(index) {
+                                    primaryContentSectionsVariants[index].addPatch(.add(value: value), toVariantWithTraits: traits)
```
patshaughnessy, line 250: After testing this with some real archives and render nodes, it looks like the index will be incorrect here when one of the primary content sections has a default value of nil.

The problem is that during encoding, the `encodeVariantCollectionArrayIfNotEmpty` function for the array case, `variantCollectionValues: [VariantCollection<Value?>]`, uses two different styles of indexes:

```
try encodeIfNotEmpty(variantCollectionValues.compactMap(\.defaultValue), forKey: key)
```

... this removes any nil values while encoding, so the JSON array ends up with only non-nil values. But:

```
        for (index, variantCollection) in variantCollectionValues.enumerated() {
            variantCollection.addVariantsToEncoder(
                encoder,
                
                // Add the index to the encoder's coding path, since the coding path refers to the array.
                pointer: JSONPointer(from: encoder.codingPath + [key, JSON.IntegerKey(index)]),
                // Since `nil` default values are filtered out above, we need to know when to `add` instead of `replace` an item.
                isDefaultValueEncoded: variantCollection.defaultValue != nil
            )
        }
```

...this uses the original index values for the JSONPointer strings.

The encoder is fine, but when we try to decode using your new algorithm there will be a mismatch between the number of elements in the default value array, and the index numbers that appear in the JSON patch.

I think we need to rework how we apply the patches here during decoding. Let me know if you'd like me to work on that.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -71,7 +75,307 @@ extension RenderNode: Codable {
...
+                                    primaryContentSectionsVariants = []
+                                } else {
+                                    let values = try patchContainer.decode([CodableContentSection?].self, forKey: .value)
+                                    
+                                    // Replace all the values
+                                    for (index, value) in values.enumerated() {
+                                        if primaryContentSectionsVariants.indices.contains(index) {
+                                            primaryContentSectionsVariants[index].addPatch(.replace(value: value), toVariantWithTraits: traits)
+                                        } else {
+                                            primaryContentSectionsVariants[index] = VariantCollection(
```
patshaughnessy, line 186: Should this be an append, not an assignment? We know the index is out of range here in the else clause.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -71,7 +75,307 @@ extension RenderNode: Codable {
...
+                                            )
+                                        }
+                                    }
+                                    let countDifference = primaryContentSectionsVariants.count - values.count
+                                    if countDifference == 0 {
+                                        // All values are already replaced
+                                        continue
+                                    } else if countDifference < 0 {
+                                        // There are more variant values than default values. Those need to be added.
+                                        for value in values.dropFirst(primaryContentSections.count) {
```
patshaughnessy, line 198: The count here, `primaryContentSections.count`, will be incorrect in the same scenario as the earlier comment. When there's a nil default value for some primary content section, the indexes and counts will not line up because of the compactMap in the encoder.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -71,7 +75,307 @@ extension RenderNode: Codable {
...
+                                            primaryContentSectionsVariants.append(VariantCollection(
+                                                defaultValue: nil,
+                                                variants: [.init(traits: traits, patch: [.replace(value: value)])]
+                                            ))
+                                        }
+                                    } else {
+                                        // There are fewer variant values than default values. Those extra default values need to be removed.
+                                        for index in countDifference ..< primaryContentSectionsVariants.count {
```
patshaughnessy, line 206:
```suggestion
                                        for index in values.count ..< primaryContentSectionsVariants.count {
```

I think what you have will produce the wrong patch indexes.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -139,3 +443,192 @@ extension KeyedEncodingContainer {
...
+            throw DecodingError.dataCorruptedError(forKey: .pointer, in: self, debugDescription: "Unexpected non-index JSON patch for element of list: \(operation.rawValue) at \(try _pointer).")
+        }
+        
+        guard remainingPathComponents.isEmpty else {
+            // None of the variant collections hold values that are variant collections. Thus, in practice we don't support patches to individual properties of elements of any list.
+            throw DecodingError.dataCorruptedError(forKey: .pointer, in: self, debugDescription: "Unsupported JSON patch for modification of individual property of an element in a list: \(operation.rawValue) at \(try _pointer).")
+        }
+        
+        var modifiedValue = mainValue()
+        guard modifiedValue.indices.contains(index) else {
```
patshaughnessy, line 577:
```suggestion
        guard modifiedValue.indices.contains(index) || (operation == .add && index == modifiedValue.endIndex) else {
```

I think for the insert/add case only, the end index is valid. You have code like this above on line 239 but we need it here also.

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -139,3 +443,192 @@ extension KeyedEncodingContainer {
...
+    
+    func _decodeOptionalArrayPatch<Value>(
+        mainValue: @autoclosure () -> [Value]?,
+        operation: PatchOperation,
+        pathComponents: [String],
+        nilValue: (() -> [Value])? = nil
+    ) throws -> VariantPatchOperation<[Value]?> {
+        guard !pathComponents.isEmpty else {
+            // The patch applies to the array of values, replace it as a scalar
+            return try _decodeScalarPatch(operation: operation, pathComponents: pathComponents, nilValue: { [] })
```
patshaughnessy, line 602:
```suggestion
            return try _decodeScalarPatch(operation: operation, pathComponents: pathComponents, nilValue: { nil })
```

For optional arrays we should use `nil` as the nil value, not an empty array, right?

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
```diff
@@ -139,3 +443,192 @@ extension KeyedEncodingContainer {
...
+                return .replace(value: value)
+            }
+            let value = try decode(Value.self, forKey: .value)
+            return .replace(value: value)
+        case .add:
+            if try decodeNil(forKey: .value) {
+                guard let value = nilValue?() else {
+                    throw DecodingError.dataCorruptedError(forKey: .value, in: self, debugDescription: "Unexpected 'null' value for \(operation.rawValue) JSON patch at \(try _pointer).")
+                }
+                return .add(value: value)
```
patshaughnessy, line 544: Should this be `.replace` to reflect how the encoder changes replace to add? In other words, should the decoder revert back from add to replace here? (And two lines below?)

patshaughnessy (changes requested):
Amazing work 👏 👏 

I'm very impressed by how you implemented this, in particular how you were able to find a type-safe way to transform from the JSON Patch/Operation encoded values back to the DocC Render Node variant collections. Amazing! You didn't have to resort to encoding/decoding JSON loops to convert between these two structures. I love how you were able to use `nestedUnkeyedContainer` and then iterate over the keys, switching from one type to another.

I ran some extensive tests today, and found some bugs in this which I've commented on here.

I'm happy to take this PR from you and finish it up - we're 95% of the way there just a few issues to work through!

Sources/SwiftDocC/Model/Rendering/RenderNode/RenderNode+Codable.swift
d-ronnqvist: Yes, I think that's what I meant intended here.

Sources/SwiftDocC/Converter/RenderNode+Coding.swift
patshaughnessy: I've started working on this PR - so far I've addressed this specific issue only in: 506ec40065d81268814622c5d3e9fe8784593932.

I'll continue to address the other issues next week.

Tests/SwiftDocCTests/Rendering/Variants/VariantOverridesEndToEndTests.swift
patshaughnessy: Added back in e0ca308a2b0d76834700b53883dd583351f2e316.

PR #1667, opened 2026-09-17
updates the code block annotation to on by default

heckj:
Bug/issue #, if applicable: 

## Summary

Makes the code annotation parsing and copy-to-clipboard active by default, and removes the `experimental` from the flags.
Also updates the tests, documentation, and touches the feature flags for DocC to reflect things correctly for the new state.

Related thread on Swift Forums: https://forums.swift.org/t/rfc-fully-enabling-enable-experimental-code-block-annotations/89348

## Dependencies

n/a

## Testing

Inline unit tests updated, and the simplest check is generating docs with a code block in it from this PR, and viewing it in DocC, where you'll see a copy-to-code button over that block when you hober.

## Checklist

- [x] Added tests
- [x./bi] Updated documentation if necessary


patshaughnessy (changes requested):
Thanks a lot for this, @heckj ! I like the idea of enabling this by default as you explained in the forum post.

Reading your unit test changes, I was wondering if DocC should display a warning if code annotations are explicitly disabled, but the compiler finds some annotated code. What do you think? Reading the code it looks like it does not now. Conversely, it looks like DocC does warn about invalid annotation syntax, even if the feature is disabled. Maybe this sort of thing is out of scope for this PR?

A couple of nits from outside the diff in this PR:

There's still a reference to the experimental flag in FeatureFlags+Info.swift:

```
        /// This feature flag corresponds to ``FeatureFlags/isExperimentalCodeBlockAnnotationsEnabled``.
        public var experimentalCodeBlockAnnotations: Bool?
```

Should we rename this also?

And in RenderBlockContent.swift:

```
       // empty initializer with default values
        @available(*, deprecated, renamed: "init(copyToClipboard:showLineNumbers:wrap:lineAnnotations:)", message: "Use 'CodeBlockOptions.init(copyToClipboard:showLineNumbers:wrap:lineAnnotations:)' instead. This deprecated API will be removed after 6.5 is released.")
        public init() {
            self.language = ""
            self.copyToClipboard = FeatureFlags.current.isExperimentalCodeBlockAnnotationsEnabled
            self.showLineNumbers = false
            self.wrap = 0
            self.lineAnnotations = []
        }
```

Probably we should change that to `self.copyToClipboard = isCodeBlockAnnotationsEnabled` even though this is deprecated.

heckj:
We intentionally didn't add any warnings if you put in code blocks but had that disabled, as it meant to be explicitly ignored in that scenario and we didn't want to burden folks with an incoming experimental flag. We did check to report is any parsed flags *were* read as invalid, as we expected that was likely an authoring error that an author would want to explicitly know about.

If we'd like to update the warning scenarios with those going full-on live and enabled, I'm totally game - but I think that may be better as a follow up PR - and I'm happy to help assemble and wrangle that.

heckj:
I wasn't 100% confident on the feature flag initialization conventions, so I followed the pattern established by `experimentalOverloadedSymbolPresentation` - holler if that's isn't appropriate. I think the only thing exercising this new public initializer is the tests included. And I did the additional work to keep allowing the older Info.plist key, which I presumed was the right way to go forward here, but if that's not needed, let me know and we can simplify some of this and strip that test.

@patshaughnessy I think this sorts the nits you called out - let me know if I tackled any of the feature flag or migration pieces in an awkward way.

patshaughnessy (approved):
Looks good - thanks Joe!

patshaughnessy:
@swift-ci please test

